PR #285, opened 2014-12-12
Fixes for dependencies and subscribers properties in sensu::check.

jonathanio:
Fixes two issues with the Puppet configuration:

- The dependencies property is not handled correctly by the provider causing it to be added or removed with alternate runs; and
- Set subscribers property in sensu::check as undef to prevent empty array output in configuration.


jonathanio:
This failure seams to be an issue with the puppet module itself. It fails on my branch and on master on my machine at the moment?

PR #286, opened 2014-12-14
Revert "Flapjack support for puppet"

jlambert121:
Reverts sensu/sensu-puppet#284

I'm going to revert this for a few reasons:

This by default will install the flapjack extension for all users
The flapjack plugin is pulled directly from GH meaning installs are not idempotent (that file changes over time)
- which leads to we need a generic method for installing extensions as well as configuring them


jessereynolds:
Awww, exciting...

jlambert121:
This was added as generic extension support in #297

jessereynolds:
@jlambert121 ah brilliant, thanks.

poolski:
Beautiful!

PR #302, opened 2015-01-25
Add parameter to allow purging plugins, handlers, extensions and mutators

nhinds:
The default behaviour is not to purge, matching the existing behaviour.

Currently if I remove a `sensu::plugin`, `sensu::handler` or `sensu::extension` definition, the plugin/handler/extension remains on disk. This parameter allows these to be removed automatically when they are removed from puppet.

NB: there is currently a `purge` setting on `sensu::plugin` which is supposed to do this, but it only works if the `sensu::manage_plugins_dir` parameter is set to `false` - if `sensu::manage_plugins_dir` is `true`, `sensu::package` manages `/etc/sensu/plugins` and doesn't set the `purge` parameter.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -175,11 +175,6 @@
 #   Default: false
 #   Valid values: true, false
 #
-# [*purge_scripts*]
-#   Boolean.  If unused plugins, handlers, extensions and mutators should be removed from the system
-#   Default: false
-#   Valid values: true, false
-#
 # [*use_embedded_ruby*]
 #   Boolean.  If the embedded ruby should be used
 #   Default: false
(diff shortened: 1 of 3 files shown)


jessereynolds:
+1 ... I was expecting this behaviour when i was getting started with this module.

jamtur01:
@jlambert121 ?

jlambert121:
Is there a reason to not reuse the existing purge flag?  Having multiple purge flags seems to make the code more confusing.

nhinds:
Fair enough - I can reuse `sensu::purge_config` for this.

The reason I avoided using it initially is that I thought it was intended to control only the JSON configuration files in `conf.d`, and because the files in `plugins`/`handlers`/`extensions`/`mutators` feel more like code than configuration. But I'm happy to use the same flag, since I'd expect most people who set `purge_config` would also want the rest of the files to be managed by puppet.

jlambert121:
I could be wrong, but my thought is people are either going to want puppet to delete unused stuff for them or not.  If this turns out to be untrue we will have to get more granular.

jlambert121:
Thanks for the PR!

PR #303, opened 2015-01-26
Sort array properties before comparison

dpeters:
The order of properties can be different when a catalog is compiled by a
different puppet master (i.e. clustered puppet masters).

Sorting in the type should prevent resources from being updated just to
change the sort order of an array.


jlambert121:
Thanks for the PR!

PR #312, opened 2015-02-09
Fix dependency chain when deploy plugins directory

bashtoni:
When deploying a whole directory of plugins, the module wasn't ensuring that the sensu package was already installed.


jlambert121:
Thanks!

PR #316, opened 2015-02-12
Fix redis noauth

bashtoni:
Commit 9ece56f8 adds support for Redis authentication, but breaks previously working setups, because Sensu now tries to authenticate with an empty password if none was specified.

This commit means that existing setups will continue to function, but allows Redis authentication to be used if desired.


jlambert121:
Good point - thanks!

PR #319, opened 2015-02-15
Also allow rabbitmq_ssl_private_key to accept the key itself (or a varialbe contianing the key) or a value in hiera; in this case, the will be written to a file.

brianwhigham:
This should retain the current functionality and add the option to use heira, a variable containing the key, or a string with the key itself.  Any of these three options would cause the key.pem file to be written with the value given.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -101,9 +101,10 @@
 #   Defaul: false
 #
 # [*rabbitmq_ssl_private_key*]
-#   String.  Private key to be used by sensu to connect to rabbitmq
-#     If the value starts with 'puppet://' the file will be copied and used.  The key itself can be
-#     defined inline or using hiera.  Absolute paths will just be used
+#   String.  Private key to be used by sensu to connect to rabbitmq. If the value starts with
+#     'puppet://' the file will be copied and used.  Also the key itself can be given as the
+#     parameter value, or a variable, or using hiera.  Absolute paths will just be used as
+#     a file reference, as you'd normally configure sensu.
 #   Default: undef
 #
 # [*rabbitmq_ssl_cert_chain*]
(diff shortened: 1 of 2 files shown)


jlambert121:
Would you squish these commits into one, add some tests, and I'll merge it?  Thanks!

jlambert121:
@brianwhigham have you had a chance to look at this?

dkiser:
I was just looking for exactly this functionality.  Should the same logic also apply to the CA chain file?

ghoneycutt:
I'm not sure that we should allow puppet:/// sources for these files as it promotes forking modules and keeping sensitive information tightly coupled with a module. Instead we should prefer content while also allowing local sources to be used, in case the files are managed elsewhere such as a package.

ghoneycutt:
This will change the behavior of `rabbitmq_ssl_private_key` and `rabbitmq_ssl_cert_chain`

Phil-Friderici:
Isn't that functionality (at least the main part) already included in the module ?
https://github.com/sensu/sensu-puppet/blob/v2.24.0/manifests/rabbitmq/config.pp#L78-L87
https://github.com/sensu/sensu-puppet/blob/v2.24.0/manifests/init.pp#L151-L156

ghoneycutt:
Which functionality? I'm saying that I don't think allowing `puppet:///` sources are a good idea, even though it currently exists.

ghoneycutt:
@Phil-Friderici ping

Phil-Friderici:
Sorry, I got confused because this PR tries to add the feature to allow ```puppet://``.  Actually this feature is already in the code [1]. Wanted to point out that there is no need to merge this anymore.

[1] https://github.com/sensu/sensu-puppet/blob/master/manifests/rabbitmq/config.pp#L64-L78

ghoneycutt:
It appears this functionality already exists. Please reopen if we missed something. Thank you.

PR #321, opened 2015-02-25
add option to purge plugins directory

yyejun:
/etc/sensu/plugins is not purged by puppet even when manage manage_plugins_dir is set to true. This seems very counterintuitive to me.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -239,7 +239,7 @@ class sensu (
   $plugins                  = [],
   $plugins_dir              = undef,
   $purge_config             = false,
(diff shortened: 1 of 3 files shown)


jlambert121:
@yyejun thanks - this fixes the issue with the previous implementation.  I think we need to default purge to false though (avoid breaking current installations and those that use packages to install plugins).  Also could you throw some tests around it and it'll be good to go!

yyejun:
@jlambert121 thanks. Updated as you suggested.

jlambert121:
Awesome - thanks!

PR #323, opened 2015-02-27
to_type convert :undef into string

keymone:
i'm not digging too much into why puppet is inconsistent with the `@should` variable between client/server and standalone modes, but that's how it is. `@should` contains :undef as a symbol and that makes bunch of types not in sync because values read from json never contain ruby symbols.


jlambert121:
Thanks!

PR #326, opened 2015-03-01
Fix type typo

bbanzai:
Change api_password type description to String from Integer


bbanzai:
Not sure why this is breaking the build.

jlambert121:
This isn't what caused the build to fail - making another PR to fix that.  Thanks for fixing this!

PR #329, opened 2015-03-03
Restrict access to the client config file to protect client tokens

jinnko:
When using command token substitution we put passwords in the client
config file.  In order to properly protect the passwords this file
must not be world readable.


jlambert121:
Great - thanks!

PR #333, opened 2015-03-18
Add support for :reconnect_on_error.

jonathanio:
Added properties for both Redis and RabbitMQ types to enable support for the new :reconnect_on_error option in Sensu 1.7.0.


jlambert121:
Thanks!

PR #341, opened 2015-04-10
Make client port configurable, issue #335

### LINKED ISSUE #335: Unable to configure client port
By default the client port is set to 3030. This is configurable https://github.com/sensu/sensu/blob/master/lib/sensu/client/process.rb#L315-L322 using the 'port' variable in the client config.

superseb:
Only debate we could have is wether to add a default, I added a default because the bind address property has a default.

Fixes #335


jamtur01:
Thanks! Support a default.

PR #343, opened 2015-04-13
Add port to check_args so it doesn't gets cleared by custom property

### LINKED ISSUE #342: Client_custom overrides client_port
If custom_client param is left as default, null, then it nullifies client_port.

```
Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[sensu-client.test.com]/port: port changed '19030' to '19030'
Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[sensu-client.test.com]/custom: custom changed 'port => 19030' to ''
```

superseb:
This should fix #342


s33dunda:
Thanks @superseb!

Issue #339, opened 2015-04-09
sensu_gem provider proxy support


bjwschaap:
I'd like to use embedded Ruby, and the sensu_gem provider to install gems to be used by my plugins. I am behind a corporate firewall.
However, the current method of using the default package type and a custom provider doesn't allow for parameterized proxy configuration.

jlambert121:
Closed with #344

PR #344, opened 2015-04-14
Add install_options for (sensu-)gem provider(s)

bjwschaap:
This is a proposed solution for issue #339, which allows to pass install_options to the parent gem provider.


jamtur01:
Sure - seems reasonable. Thanks!

PR #348, opened 2015-05-04
Updating APT source to use new apt module version

nicolas-fernandez-olx:
New version of Puppet APT module was released the PR will fix this issue:
``` 
Puppet (err): Invalid parameter include_src on Apt::Source[sensu] at /etc/puppet/modules/sensu/manifests/repo/apt.pp:35
```

PR #351, opened 2015-05-12
make apt class compliant with puppetlabs/apt

bleuchtang:
I've got this error with the last version of puppetlabs/apt:
Error: Invalid parameter include_src on Apt::Source[sensu] at /srv/puppet/test/puppet/modules-ext/sensu/manifests/repo/apt.pp:35 on node lab07.xilopix.net
Wrapped exception:
Invalid parameter include_src

This commit fix it.


bleuchtang:
duplicated with #348

PR #352, opened 2015-05-12
Hiera Lookups

bleuchtang:
Add Hiera Lookups (for extensions, handlers, and checks) and create resources with them.


jamtur01:
Thanks!

PR #355, opened 2015-05-24
Update of apt::source to support new version of puppetlabs-apt

zdenekjanda:
This is needed for supporting new puppetlabs-apt module versions.

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -28,9 +28,10 @@ class sensu::repo::apt {
       location    => $url,
       release     => 'sensu',
       repos       => $sensu::repo,
-      include_src => false,
-      key         => $sensu::repo_key_id,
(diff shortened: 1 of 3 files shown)


jamtur01:
Looks like some tests failed.

Is it backwards compatible?

zdenekjanda:
Backward compatibility needs to be improved, but do we need it ? Here similar situation - https://github.com/jenkinsci/puppet-jenkins/pull/301. New apt will break most of modules around, similar discussions all around.

jlambert121:
I think this is a dupe of #348.  We've merged some stuff that ups our puppet dependencies, maybe we merge one of these and figure out what else we need/want for a major version bump.

zdenekjanda:
I am in favor of #348 then, closing this PR.

PR #356, opened 2015-05-26
Ensure Package[sensu] is not already define

bleuchtang:
Because we can run both sensu-puppet and puppet-uchiwa recipe on the same host we should verify if the other package doesn't declare Package[sensu] before.

### DIFF
--- a/manifests/repo/apt.pp
+++ b/manifests/repo/apt.pp
@@ -10,7 +10,7 @@ class sensu::repo::apt {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  if defined(apt::source) {
+  if (defined(apt::source)) and (!defined(Apt::Source[sensu])) {
 
     $ensure = $sensu::install_repo ? {
       true    => 'present',
@@ -34,7 +34,7 @@ class sensu::repo::apt {
       before      => Package['sensu'],
     }
 
-  } else {
+  } elsif (!defined(Apt::Source[sensu])) {
     fail('This class requires puppetlabs-apt module')
   }
 


jlambert121:
Which puppet-uchiwa module are you referring to?

bleuchtang:
This one: https://github.com/Yelp/puppet-uchiwa

jlambert121:
I don't see that the uchiwa declares the package sensu, and I don't think this PR will change this module installing the package sensu.  I think the issue you're running into is declaring Apt::Source['sensu'] multiple times.  Both the sensu module and this uchiwa module support not managing the repo.  If you set either ::sensu:: install_repo: false or ::uchiwa::install_repo: false I think your repo issue will be resolved.

bleuchtang:
I don't think so, but we have the same discussion here Yelp/puppet-uchiwa/issues/36

jlambert121:
Looking at thread (which also includes the error) that is exactly the issue.  You need to just disable the repo management on one of these modules (which can be done through hiera).  

This PR looks to see if the sensu package is defined and skips installing the repo, which means if someone uses just this repo (without the uchiwa module) the repo will never get installed on debian systems and the module will fail.

bleuchtang:
No, my PR look if Package[sensu] is not already define (by uchiwa for instance) and if it's not sensu recipe install the repository. I try my PR with and without uchiwa and it work like a charm.

bleuchtang:
Ok, I rebased my PR maybe it's more acceptable now.

jlambert121:
I understand that this will have the desired effect when using the uchiwa module, on a clean machine (never had sensu or uchiwa installed) I believe this PR will prevent the installation of sensu.  Thank you for this PR, but I don't think this is one we're going to merge.  Disabling the apt repo in one of these modules with the install_repo param will fix the error you posted on the uchiwa module.

bleuchtang:
Ok closing

PR #358, opened 2015-06-05
Filter attributes are a property, not a param

bashtoni:
Attributes for filters were not being added because parameters don't call methods.  Attributes for filters are now properties.


jlambert121:
Thanks!

PR #359, opened 2015-06-08
allow modification of hasrestart attribute for services

somic:
At my company we ship our own startup scripts that use upstart. When those scripts change, in order for upstart to pick up the changes, a job needs to be stopped and then started (see http://upstart.ubuntu.com/faq.html#reload).

With hasrestart hard coded to true, it's not possible. This change allows us to control hasrestart attribute.

This change is fully backwards compatible - by default value of hasrestart everywhere is true.

### DIFF
--- a/manifests/api/service.pp
+++ b/manifests/api/service.pp
@@ -2,7 +2,17 @@
 #
 # Manages the Sensu api service
 #
-class sensu::api::service {
+# == Parameters
+#
+# [*hasrestart*]
+#   Boolean. Value of hasrestart attribute for this service.
+#   Default: true
+#
+class sensu::api::service (
+  $hasrestart = true,
+) {
+
+  validate_bool($hasrestart)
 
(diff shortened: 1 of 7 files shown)


jlambert121:
Could you add a test to make sure true is set on hasrestart as the default?  (Will help raise a flag if the default changes)

somic:
Looks to me like this test coverage is provided by context default under service:

https://github.com/sensu/sensu-puppet/blob/master/spec/classes/sensu_client_spec.rb#L57
https://github.com/sensu/sensu-puppet/blob/master/spec/classes/sensu_api_spec.rb#L105
https://github.com/sensu/sensu-puppet/blob/master/spec/classes/sensu_server_spec.rb#L19

If by default hasrestart ends up being anything but true, these tests should fail.

jlambert121:
You are absolutely right, my apologies.

PR #362, opened 2015-06-12
ensure plugins installed before client service started

### LINKED ISSUE #353: First run on a new client node fails checks which depend on plugins
I have a client check which depends on a plugin. To install it on the client, i pass it to the 'plugins' property of sensu, like:
  class { 'sensu':
    ...
    plugins                     => [
      'puppet:///modules/profile/.../sensu/plugins/processes/check-procs.sh',
    ],
    ..
  }

However this causes the first run of the check to fail, because the checks and sensu-client service are enabled in the puppet module before the plugins are installed:
    sh: /etc/sensu/plugins/check-procs.sh: No such file or directory 

To fix, the plugins should be installed before the service + checks/subscriptions are created.

jlambert121:
Resolves #353

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -47,6 +47,9 @@ define sensu::plugin(
   $nocheckcertificate = false,
 ){
 
+  Sensu::Plugin[$name] ->
+  Class['sensu::client::service']
+
   validate_bool($purge, $recurse, $force, $nocheckcertificate)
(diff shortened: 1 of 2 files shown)


jamtur01:
It appears Travis is sad.

jlambert121:
I committed the wrong file, it'll be happy again shortly

PR #364, opened 2015-06-12
update travis, gems, lint

jlambert121:


jlambert121:
Need to figure out what's breaking 1.8.7 (or drop 1.8.7 support which likely isn't a good idea because of RHEL6)

PR #367, opened 2015-06-12
Hiera array hash updated for client/config.pp

pdaugavietis:
First cut, works for me since all our configs for this stuff is coming out of hiera.   Not sure how best way to handle local variable instead of/with hiera lookup.


jamtur01:
Thanks for this. Lots of test failures I am afraid though. :(

PR #370, opened 2015-06-22
fix filters docstring in sensu::handler

somic:
Looks to me like filters param of sensu::handler is expected to be an Array, not a Hash, no?


jlambert121:
Thanks!

PR #378, opened 2015-06-26
Relax the apt module version restriction

johnf:
Apt 2.x is now backwards compatible with 1.8.x


jlambert121:
Should this be:
">= 1.8.0 <3.0.0"?

johnf:
Probably that's what most other seem to be doing. I'll update

jlambert121:
Thanks!

PR #393, opened 2015-08-04
remove world readable permissions from <handler>.json

jhoblitt:
The handler configuration may contain "secrets" such as passphrases or
API keys that should not be world readable.


jhoblitt:
Fixed a typo in the commit message.

jlambert121:
Good call

PR #400, opened 2015-08-07
[WIP] Sensu Enterprise & Enterprise Dashboard support

dhgwilliam:
this is very much a WIP. feedback welcome. i'm just about to dive in to the types/providers for config. i think i can probably borrow quite a bit of code.

i've also made some presumably debatable decisions which might need to be modified slightly, especially around service management for sensu-server and sensu-api.

i'm smoke testing on centos 6.6, but all the existing (and new) unit tests pass, including debian ones...


manifests/enterprise/dashboard/package.pp
```diff
@@ -0,0 +1,38 @@
+class sensu::enterprise::dashboard::package {
+
+  if $caller_module_name != $module_name {
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
+
+  if $::sensu::install_repo and $::sensu::enterprise_dashboard {
```
jlambert121: What about breaking this up a bit for the scenario where someone wants to install enterprise ($ensure => present), but hosts the RPM on a private yum server (install_repo == false)?  Instead of $ensure = 'present', $ensure => $::sensu::version I think would be better as well (or if the dashboard version doesn't match the sensu version another parameter).

manifests/enterprise/dashboard/config.pp
```diff
@@ -0,0 +1,18 @@
+class sensu::enterprise::dashboard::config {
+  if $caller_module_name != $module_name {
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
+
+  if $sensu::purge_config and !$sensu::enterprise_dashboard {
```
jlambert121: Likely #401 will end up being merged which changes the purge options (just a heads up)

manifests/enterprise/dashboard/service.pp
```diff
@@ -0,0 +1,38 @@
+# = Class: sensu::enterprise::dashboard
+#
+# Installs the Sensu Enterprise Dashboard
+class sensu::enterprise::dashboard::service (
+  $hasrestart = true,
```
jlambert121: Are there some cases when the init/service script doesn't have a restart option?

manifests/enterprise/dashboard/service.pp
```diff
@@ -0,0 +1,38 @@
...
+        $enable = true
+      }
+      default: {
+        $ensure = 'stopped'
+        $enable = false
+      }
+    }
+
+    service { 'sensu-enterprise-dashboard':
+      ensure     => $ensure,
```
jlambert121: If $::sensu::manage_services is false this will always stop the enterprise dashboard instead of just doing nothing - is that the intended behavior?

manifests/enterprise/package.pp
```diff
@@ -0,0 +1,49 @@
...
+        $repo_require = undef
+      }
+    }
+
+    default: { fail("${::osfamily} not supported yet") }
+
+  }
+
+  package { 'sensu-enterprise':
+    ensure  => $sensu::enterprise_version,
```
jlambert121: Would it be possible to just use $::sensu::version?  Just thinking it is essentially duplicating $::version just with enterprise => true.

manifests/enterprise/dashboard/package.pp
```diff
@@ -0,0 +1,38 @@
...
+    $se_pass = $sensu::enterprise_pass
+
+    validate_string($se_user, $se_pass)
+    if $se_user == undef or $se_pass == undef {
+      fail('The Sensu Enterprise repos require both enterprise_user and enterprise_pass to be set')
+    }
+
+    $url = "http://${se_user}:${se_pass}@enterprise.sensuapp.com/yum/\$basearch/"
+
+    yumrepo { 'sensu-enterprise-dashboard':
```
jlambert121: Maybe combine this repo with the sensu::enterprise::repo classes?

manifests/enterprise/repo/yum.pp
```diff
@@ -0,0 +1,33 @@
...
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
+
+  if $sensu::install_repo  {
+    if $sensu::enterprise {
+      $se_user = $sensu::enterprise_user
+      $se_pass = $sensu::enterprise_pass
+
+      validate_string($se_user, $se_pass)
+      if $se_user == undef or $se_pass == undef {
```
jlambert121: What about moving this validation to init.pp?  Keeps all of the validation in one place then and when you get here you know you can just use the data then.

manifests/plugins_dir.pp
```diff
@@ -14,7 +14,7 @@
       recurse => $recurse,
       purge   => $purge,
       force   => $force,
-      require => Package['sensu'],
+      require => Package[$::sensu::package::package_name],
     }
```
jlambert121: I'm not seeing where the package_name is changed at all - could you help me with what this is for?

jlambert121:
This is great - thanks for your work!  I'd been wondering how easy it would be to support enterprise with this, but not being an enterprise customer I hadn't looked into it at all.  I made some comments in the code with thoughts I had as I read over it and would love to see some more testing to go down all of the code paths/options with the changes.

manifests/enterprise/package.pp
dhgwilliam: my thought here is that SE versioning seems to be decoupled from Sensu OSS version, so (arbitrary made-up scenario warning) there could be a customer who is pinned to a specific client package version but they still want the latest SE version? Let me know if this is ridiculous...

manifests/enterprise/dashboard/service.pp
dhgwilliam: Are you sure? Unless I'm losing track of matching braces, the service resource should only make it in to the catalog if $::sensu::manage_services is true...

I followed the pattern from https://github.com/sensu/sensu-puppet/blob/master/manifests/api/service.pp

(just in case, I'll add a test to check this)

manifests/plugins_dir.pp
dhgwilliam: This was "leftover" from a time when I thought the sensu-enterprise package replaced the sensu package completely. In fact, you need to install sensu and sensu-enterprise. Will clean up.

manifests/enterprise/dashboard/service.pp
jlambert121: Nope, you are correct - I apparently can't count.

manifests/enterprise/package.pp
jlambert121: I'm not familiar with SE at all - is there a SE client that you also run on all of the clients or is it the standard OSS client?

manifests/plugins_dir.pp
jlambert121: Ahhh - this may also explain the above comment.  What components does SE replace?  Is it (essentially) just the dashboard addition to OSS?

dhgwilliam: sensu-enterprise replaces sensu-server and sensu-api. sensu-client (OSS),
among other things, still requires the sensu package.
sensu-enterprise-dashboard replaces uchiwa.

manifests/enterprise/dashboard/config.pp
dhgwilliam: p. sure i just rebased against this PR, will update accordingly

manifests/api/service.pp
```diff
@@ -37,5 +37,10 @@
       hasrestart => $hasrestart,
       subscribe  => [ Class['sensu::package'], Class['sensu::api::config'], Class['sensu::redis::config'] ],
     }
+  } elsif $sensu::manage_services and $::sensu::enterprise {
```
jlambert121: What about something more like this so we don't declare the sensu-api service twice?

And just to confirm, in SE api should not be running correct?

```puppet
if $sensu::manage_services {

    if $sensu::enterprise {
        $ensure = 'stopped'
        $enable = false
    } elsif $sensu::api {
        $ensure = 'running'
        $enable = true
    } else {
        $ensure = 'stopped'
        $enable = false
    }

    service { 'sensu-api':
      ensure     => $ensure,
      enable     => $enable,
      hasrestart => $hasrestart,
      subscribe  => [ Class['sensu::package'], Class['sensu::api::config'], Class['sensu::redis::config'] ],
    }
  }
```

manifests/enterprise/dashboard/package.pp
```diff
@@ -0,0 +1,33 @@
...
+
+  if $::sensu::install_repo and $::sensu::enterprise_dashboard {
+    $ensure = 'present'
+
+    $se_user = $sensu::enterprise_user
+    $se_pass = $sensu::enterprise_pass
+
+    $url = "http://${se_user}:${se_pass}@enterprise.sensuapp.com/yum/\$basearch/"
+
+    yumrepo { 'sensu-enterprise-dashboard':
```
jlambert121: I think this should be moved to sensu::repos::yum so it honors the $install_repo flag and it seems odd to install a repo somewhere else since that exists.

manifests/enterprise/dashboard/package.pp
```diff
@@ -0,0 +1,33 @@
+class sensu::enterprise::dashboard::package {
+
+  if $caller_module_name != $module_name {
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
+
+  if $::sensu::install_repo and $::sensu::enterprise_dashboard {
+    $ensure = 'present'
+
```
jlambert121: I think you need to add a dashboard version param for this or you'll be in a situation where you can't ever upgrade the dashboard through puppet on an existing server (and building a new one you aren't guaranteed to get what you had before with the same manifests)

manifests/enterprise/dashboard/service.pp
```diff
@@ -0,0 +1,38 @@
...
+      enable     => $enable,
+      hasrestart => $hasrestart,
+      subscribe  => [
+        Class['sensu::enterprise::dashboard::package'],
+        Class['sensu::enterprise::dashboard::config'], # TODO
+        Class['sensu::redis::config'],
```
jlambert121: TODO still relevant?

manifests/enterprise/repo/apt.pp
```diff
@@ -0,0 +1,29 @@
+class sensu::enterprise::repo::apt {
```
jlambert121: Is there a reason this and sensu::enterprise::repo::yum couldn't be merged into sensu::repo::yum/apt?  It'd just be a conditional in them to set the url based on enterprise true/false, right?

manifests/init.pp
```diff
@@ -314,6 +339,18 @@
...
   if $purge_plugins_dir { fail('purge_plugins_dir is deprecated, set the purge parameter to a hash containing `plugins => true` instead') }
 
+  # sensu-enterprise supercedes sensu-server and sensu-api
+  if ( $enterprise and $api ) or ( $enterprise and $server ) {
+    fail('Sensu Enterprise: sensu-enterprise replaces sensu-server and sensu-api')
+  }
+  # validate enterprise repo creds
+  if $enterprise or $enterprise_dashboard {
+    validate_string($enterprise_user, $enterprise_pass)
+    if $enterprise_user == undef or $enterprise_pass == undef {
```
jlambert121: Just a thought on this, many organizations do not allow direct access to remote repos through their firewalls so they may be mirroring the packages internally.  They can get around it by setting these settings to junk strings, just a thought though.  Maybe only require these if install_repos is true?

manifests/server/service.pp
```diff
@@ -18,7 +18,7 @@
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  if $sensu::manage_services {
+  if $sensu::manage_services and !$sensu::enterprise {
```
jlambert121: Same thing as above with the api service - can we merge the conditionals together so we aren't defining the service multiple times?

manifests/enterprise/dashboard/package.pp
dhgwilliam: :white_check_mark:

manifests/enterprise/dashboard/package.pp
dhgwilliam: I think the motivation was initially because of the potential complexity of the $repo_require logic (a few lines down), but I'll see if I can merge all the repo stuff in to a single class (per platform).

manifests/enterprise/dashboard/service.pp
dhgwilliam: :white_check_mark:

manifests/init.pp
dhgwilliam: :white_check_mark:

manifests/enterprise/repo/apt.pp
dhgwilliam: On an EL-based system, you actually need three repos installed:

- sensu (OSS),
- sensu-enterprise, and
- sensu-enterprise-dashboard

manifests/api/service.pp
dhgwilliam: Correct, the sensu-enterprise service replaces sensu-api and sensu-server.

Upon reflection, there's actually a [gate in init.pp](https://github.com/sensu/sensu-puppet/pull/400/files?diff=unified#diff-60ae41fd0a31977447947f59940ee9a4R343) that will fail() if $api and $enterprise are both true, so I _think_ we can simplify this back to the original?

~~~
  if $sensu::manage_services {

    case $sensu::api {
      true: {
        $ensure = 'running'
        $enable = true
      }
      default: {
        $ensure = 'stopped'
        $enable = false
      }
    }

...
~~~

should actually work as intended!

dhgwilliam: indeed, unit tests and acceptance tests pass!

manifests/server/service.pp
dhgwilliam: :white_check_mark:

jlambert121:
Thanks for your effort on this!

PR #401, opened 2015-08-09
Add `purge` parameter to control all purging, deprecate `purge_config…

### LINKED ISSUE #328: Unable to purge handlers, extensions, or mutators
Since #302 was reverted in #307, #321 has been merged which supposedly re-added this functionality. However the following directories are not purged with this new implementation:
`/etc/sensu/handlers`, `/etc/sensu/extensions`, `/etc/sensu/mutators`, `/etc/sensu/extensions/handlers`

I'm willing to go and create another PR to re-add this functionality, but before I do I'd like to know your preferred approach. Would you like the new parameter `purge_plugins_dir` to purge these additional directories (possibly also renaming the parameter), or would you like 4 additional parameters to purge these 4 directories? Or something in between?

nhinds:
…` and `purge_plugins`

The `purge` parameter can either be `true`/`false` (in which case it controls purging
everything), or a hash controlling each type of purge-able thing.

Fixes #328

Might need changing if we decide that the old `purge_config`/`purge_plugins` parameters should be handled differently, or if the boolean/hash usage is too confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,8 @@ and configures Sensu on each individual node via
 ### common.yaml
 
     sensu::install_repo: false
-    sensu::purge_config: true
+    sensu::purge:
+      config: true
     sensu::rabbitmq_host: 10.31.0.90
     sensu::rabbitmq_password: password
     sensu::rabbitmq_port: 5672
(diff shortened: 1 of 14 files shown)


jlambert121:
Thanks!

PR #388, opened 2015-07-22
init.pp: get rid of the ugly hack to compute $check_notify

amosshapira:
Proposed cleanup of the "ugly hack".

PR #403, opened 2015-08-24
simplify $check_notify logic

jlambert121:


jlambert121:
Was playing with the solution from #388 - I can make new puppets happy or old puppets happy, but not both.

PR #405, opened 2015-08-26
Specify transport type

thanandorn:
It would be nice if we could specify transport type even only for testing.

Thanks,
Tron


thanandorn:
Redis transport seems to be supported officially now so re-open this.

Issue #406, opened 2015-08-26
Undefined variable "file_ensure" in sensu::handler


somic:
https://github.com/sensu/sensu-puppet/blob/master/manifests/handler.pp#L145

$file_ensure variable seems to be defined within "if $source {" block and trips our handler of type='set' which doesn't have $source.

jlambert121:
@somic can you check and make sure #417 resolves your issue?

jlambert121:
#418 is the fix I should have written.

PR #418, opened 2015-09-13
Move file_ensure out of conditional

hashbrowncipher:
The file_ensure variable is referenced outside of this conditional, which fails under strict_variables.


jlambert121:
Fixes #406

PR #419, opened 2015-09-13
Make the sensu enterprise dashboard not show up unbidden

hashbrowncipher:
The logic on line 10 would drop the dashboard in place with ensure=>present even when dashboard was explicitly disabled.  I believe I've modified it to more closely follow the expectations around the $purge parameter.


jlambert121:
I think this works, thanks!

PR #420, opened 2015-09-13
Make these strings again, so that literal 'undef' doesn't get templated

hashbrowncipher:
Upon bumping to the master branch, I found 'undef' was being templated into my defaults files.  This may be Puppet version specific, but I believe that the empty string achieves our desired result under all circumstances.


jlambert121:
Setting empty strings makes puppet 4 unhappy, what's your thoughts on #421?

PR #421, opened 2015-09-14
set sysconfig parameters when defined

jlambert121:
Closes #420


templates/sensu.erb
```diff
@@ -1,6 +1,10 @@
 EMBEDDED_RUBY=<%= scope.lookupvar("sensu::use_embedded_ruby") %>
 LOG_LEVEL=<%= scope.lookupvar("sensu::log_level") %>
+<% if scope.lookupvar("sensu::rubyopt") -%>
```
fessyfoo: at least in puppet 3.8 these will return :undef  perhaps  `if scope.lookupvar("sensu::rubyopt") == :undef` ??

jlambert121:
Thanks @fessyfoo - didn't think of that, so used to having a real undefined now in future_parser.

PR #422, opened 2015-09-14
allow handle and handlers together

### LINKED ISSUE #391: issue with "Do not use 'handle' and 'handlers' together. Your 'handle' value has been overridden with 'handlers'"
In a nutshell - this doesn't make sens(u)e to me -- why was it added to say that if you have handlers, you can't diable the handling of a check?    I think it was added on commit #a9b4089cb3ff7ae28a56b8d6d535478953c8f6b8 but i don't agree with it.    I should accept whatever i tell it to do, and not impose a logic block - at least remove the 'return' there, and issue the warning for 'style' points or something.

https://github.com/sensu/sensu-puppet/blob/master/lib/puppet/provider/sensu_check/json.rb#L179

jlambert121:
Resolves #391

### DIFF
--- a/lib/puppet/provider/sensu_check/json.rb
+++ b/lib/puppet/provider/sensu_check/json.rb
@@ -175,11 +175,6 @@ Puppet::Type.type(:sensu_check).provide(:json) do
   end
 
   def handle=(value)
-    if value && resource[:handlers]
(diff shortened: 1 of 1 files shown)


countryHick:
that looks good

PR #424, opened 2015-09-14
Add require on apt::update for puppetlabs-apt 2.x

br0ch0n:
This problem doesn't always happen.  I'm unable to reproduce when
just declaring class {'sensu':} by itself.  However, it does seem
to happen when declared in a profile with other classes that use
puppetlabs-apt.  My best guess is that once one of the other
classes triggers an apt-get update, another one doesn't happen
within the same run.  In any case, the official puppetlabs-apt
docs now state that one's package resource must have an explicit
dependency on class apt::update if you expect things to work in one
run:
https://github.com/puppetlabs/puppetlabs-apt#adding-new-sources-or-ppas

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -14,6 +14,10 @@ class sensu::package {
       class { '::sensu::repo::apt': }
       if $sensu::install_repo {
         include ::apt
+        $pkg_require = Class['apt::update']
+      }
+      else {
+        $pkg_require = undef
       }
     }
 
@@ -27,6 +31,7 @@ class sensu::package {
 
   package { 'sensu':
     ensure  => $sensu::version,
+    require => $pkg_require,
   }
 
   if $::sensu::sensu_plugin_provider {


Pryz:
Same issue here with puppetlabs-apt 2.1.1 !

This PR works for me ! Thanks :)

jlambert121:
This whole apt 2.0 stuff is annoying, bet this breaks 1.x versions.  I'm thinking with a sensu 2.0 module maybe we'll update the deps to be apt 2.0 as well.  Thoughts from those who use apt?

br0ch0n:
There's already been a dependency on puppetlabs-apt 2.0 since https://github.com/sensu/sensu-puppet/commit/54692955f95506b22b90fa814b36595c42c19bd9 happened (the "include" param kills puppetlabs-apt 1.x...just confirmed).  My change just requires a class that's been defined since at least  puppetlabs-apt 1.5, so it probably wouldn't break anything.  However, due to the above it doesn't matter anyway.  Requiring 2.x is the only way forward unless the referenced commit (and the subsequent one which added the enterprise repo) is altered somehow to add backwards compatibility.

jlambert121:
I thought we'd jumped back to 1.8.x to make it easier for people to migrate, but that is incorrect - thanks for the response.

PR #431, opened 2015-10-13
Adding option to manage the handlers dir

### LINKED ISSUE #430: Add option not to manage handlers dir
We already have manage_plugins, we should also have manage_handlers imho.

jaxxstorm:
Fixes #430

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -85,6 +85,11 @@
 #   Default: true
 #   Valid values: true, false
 #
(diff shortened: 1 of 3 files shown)


jlambert121:
Can you put some tests around this to make sure the parameter works as expected and I'll merge it?

jaxxstorm:
tests added @jlambert121

jlambert121:
Thanks!

PR #434, opened 2015-10-16
Update package repository URLs

portertech:
Updated the module to use the new Sensu package repository URLs, for example:

http://repositories.sensuapp.org/apt/

http://repositories.sensuapp.org/yum/

The new repositories contain the latest Sensu builds, >= 0.20.0 only. The repositories are part of the new Sensu Core build pipeline.

tumblr_mvrc0evts81swhepeo1_250_zps371b1695

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -53,7 +53,7 @@
 #
 # [*repo_key_source*]
 #   String.  URL of the apt GPG key
-#   Default: http://repos.sensuapp.org/apt/pubkey.gpg
+#   Default: http://repositories.sensuapp.org/apt/pubkey.gpg
 #
 # [*client*]
 #   Boolean.  Include the sensu client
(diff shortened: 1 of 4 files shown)


jlambert121:
Just thinking about this, I'm wondering if we need to put a switch in based on the version of sensu requested unless you're planning on adding (at least) some of the older versions into your new repository.

roobert:
I made these changes independently and found that puppet runs fail due to the new repository key not being added. I had to add some version check hackery to add the key to clients that aren't running 0.21.0 in order to get the key added.

Would it not make more sense to add all the old packages to repositories.sensuapp.org and point repo.sensuapp.org to where the new packages are hosted?

taik0:
How is this still not merged?

jamtur01:
@taik0 There appear to be open questions about the PR and backwards compability. Until those are resolved we're not going to merge the PR. Would welcome your input or addition to the PR! Thanks!

jamtur01:
Ping @portertech re old versions.

taik0:
@jamtur01 Ops, I see, didn't know that @portertech was a Sensu contributor.
So we are waiting to see if Sensu add the old packages to the new repo?
If they didn't do it, should I create a PR like @jlambert121 said configuring the old or the new depending on the version?

nhinds:
It looks like this PR needs the apt key updated as well, the fingerprint doesn't match the key used by the APT repository. Same problem as https://github.com/Yelp/puppet-uchiwa/pull/62

shpoont:
Why not merge those repos as @roobert suggested? If I use current puppet module I get 4 months old version..

portertech:
Updated the key to match Yelp/puppet-uchiwa#62

portertech:
Ping @jlambert121

poolski:
ping @jlambert121 ;)

PR #436, opened 2015-10-18
add ttl to check provider

gsalisbury:
Adds check ttl feature to check provider.


jlambert121:
It looks like there's a typo in the spec file that's causing the tests to fail, would you take a look at it?

gsalisbury:
@jlambert121 how does it look now?

jlambert121:
Thanks!

PR #443, opened 2015-11-12
update client config to use socket hash

gsalisbury:
Use a socket hash of bind address and port in the sensu_client_config provider.


jlambert121:
Thanks!  What versions of sensu support the socket hash?

gsalisbury:
@jlambert121 I think since 0.17, going by the documentation - https://sensuapp.org/docs/0.17/clients has it whereas https://sensuapp.org/docs/0.16/clients does not.

Cheers,
George

jlambert121:
Long enough ago then, thanks.  Just wanted to make sure it wasn't a brand new feature.

lib/puppet/type/sensu_client_config.rb
```diff
@@ -45,14 +45,31 @@ def insync?(is)
...
-    desc "The IP that client will bind to"
-    defaultto '127.0.0.1'
-  end
+  newproperty(:socket) do
+    desc "A set of attributes that configure the Sensu client socket."
+    include PuppetX::Sensu::ToType
+
+    def is_to_s(hash = @is)
+      hash.keys.sort.map {|key| "#{key} => #{hash[key]}"}.join(", ")
+    end
```
zbintliff: This breaks the puppet run for everyone updateing from an old puppet module.  Because the @is is '' or nil.    Is there a reason why just using hte default to_s function of the has isn't enough?

For example when upgrading to this commit running puppet get this error:
` Error: Could not convert change 'socket' to string: undefined method `keys' for nil:NilClass `

When I change is_to_s to just return hash it then updates cleanly with following log line
`Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[sensu1b-af93681f.coreeng.mlbam.dev.us-east-1.bamgrid.net]/socket: socket changed '' to {'bind' => '127.0.0.1', 'port' => '3030'}`

zbintliff:
Fix pending in  #457

PR #451, opened 2015-11-30
Fix dependencies for ruby 2.2.

jguiditta:
Without these changes, running specs fails when using:
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-linux]

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,6 +14,8 @@ group :development, :unit_tests do
   gem 'metadata-json-lint',                               :require => false
   gem 'puppet-blacksmith',                                :require => false
   gem 'rest-client', "1.6.8",                             :require => false
+  gem 'syck',                                             :require => false
+  gem 'safe_yaml',                                        :require => false
   gem 'puppet-lint-absolute_classname-check',             :require => false
   gem 'puppet-lint-appends-check',                        :require => false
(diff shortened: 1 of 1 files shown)


jlambert121:
Ahh... the joys of old ruby.  I believe there is a way to constrain gems in the Gemfile to just certain versions of ruby, could you do that?

Gemfile
```diff
@@ -14,6 +14,8 @@ group :development, :unit_tests do
   gem 'metadata-json-lint',                               :require => false
   gem 'puppet-blacksmith',                                :require => false
   gem 'rest-client', "1.6.8",                             :require => false
+  gem 'syck',                    :require => false
```
EmilienM: can you align it please?

jguiditta: done, though I would still need to make the requested change to constrain this to certain versions of ruby, which I believe I had no success with last time I looked at this patch.  Happy to make changes if there are suggestions how though.

jaxxstorm:
I think we need to update the `.travis.yml` as well here, to actually run tests for the new ruby in sensu 0.25.

jaxxstorm:
i'm going to completely overhaul the gem file and travis matrix next week, so closing this as it's very out of date

PR #453, opened 2015-12-07
transform default client_port into an integer

EmilienM:
By using the old default value which was a string and not an integer,
sensu-client would never start and you would got this message:

{"timestamp":"2015-12-06T21:18:52.299496+0000","level":"fatal","message":"client socket port must be an integer","object":{"name":"instack","address":"192.0.2.1","subscriptions":["undercloud_basic"],"socket":{"bind":"127.0.0.1","port":"3030"},"safe_mode":false,"keepalive":{}}}

The port value (defined by sensu::client_port here) must be an integer
and not a string.
This patch transforms the default value from a string to an integer so
we can run sensu-client out-of-the-box.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -314,7 +314,7 @@ class sensu (
   $api_password                   = undef,
   $subscriptions                  = [],
   $client_bind                    = '127.0.0.1',
-  $client_port                    = '3030',
+  $client_port                    = 3030,
   $client_address                 = $::ipaddress,
   $client_name                    = $::fqdn,
   $client_custom                  = {},
@@ -360,6 +360,7 @@ class sensu (
   validate_re($log_level, ['^debug$', '^info$', '^warn$', '^error$', '^fatal$'] )
   if !is_integer($rabbitmq_port) { fail('rabbitmq_port must be an integer') }
   if !is_integer($redis_port) { fail('redis_port must be an integer') }
+  if !is_integer($client_port) { fail('client_port must be an integer') }
   if !is_integer($api_port) { fail('api_port must be an integer') }
   if !is_integer($init_stop_max_wait) { fail('init_stop_max_wait must be an integer') }
   if $dashboard { fail('Sensu-dashboard is deprecated, use a dashboard module. See https://github.com/sensu/sensu-puppet#dashboards')}


agarstang:
I changed this myself (closed PR as is duplicate) however I have noticed that the port is still getting "stringyfied." Have you encountered this after your patch?

EmilienM:
@agarstang haven't tested yet but I will today.
I noticed your patch is testing if the value is an integer, do you prefer to keep my PR? Should I push -f to also test if it's an integer?

agarstang:
@EmilienM 
I'm happy to keep yours, feel free to pull the integer check (it kind of feels redundant if it gets converted back to a string anyway).

I'll look further into it when I have a moment later, would be nice to know if you experience the same.

An aside, hopefully we'll see Puppet 4 type hinting in the near future so that will be more elegant.

EmilienM:
The bug is more complex, it's in lib/puppet/type/sensu_client_config.rb - I'm working on it, do not merge this patch.

EmilienM:
6d8d665fb7faa558f398358e5be83d95e9d75988 is the source of this bug by the way -- cc @gsalisbury

EmilienM:
@agarstang so I don't have the bug when I configure sensu-puppet with Hiera and set sensu::client_port - but I have the bug if I don't use Hiera and keep the default value (string or integer).

EmilienM:
So quick update after some investigation:

* My patch is fixing the bug with or without Hiera. We can safely merge it, I tested it on my lab.
* d8b9d3270e2871cf3415cb5586c0f96164006f35 is the root cause of the issue, this commit added the parameter as a string, and not an integer cc @superseb 

Please review this patch asap, thanks!

superseb:
Spec tests also declare integer, should be good to go. I don't think we override the parameter in our setup, I can check in a few hours. Is this broken since a new release?

jlambert121:
What version of sensu are you running to get this error?

EmilienM:
@superseb I tested against master, not against a release.
@jlambert121 the latest one provided by the official repo on centos7 systems.

EmilienM:
That's very obvious there is a bug in the code. The value was not an integer. How can I help to get this merged?

jlambert121:
@EmilienM What version is that?

EmilienM:
@jlambert121 sensu-0.20.3-1.x86_64

jlambert121:
Just trying to figure out why this is just now appearing, the check has been present since at least 5/13/15 (https://github.com/sensu/sensu-settings/commit/f95a919e66bb1f0d478b547a6c6a0a59fc879a55).  I'm not disagreeing this should be an int, again, just trying to figure out why we haven't seen this before.  Did you set client_port at one point and now unset it?

EmilienM:
@jlambert121 I never deployed Sensu before this patch.
Do you run a real CI against this module? I'm not sure you actually deploy Sensu.
Moreover, I'm sure most people override this parameter with an integer (which Hiera or not), otherwise, it's just not possible to deploy it.

superseb:
That was exactly my point, I have the code running with '3030' string and not overridden by Hiera and the daemon starts just fine. 

@EmilienM : CentOS 7 with Sensu RPM sensu-0.20.3-1.x86_64 and what Puppet version?

EmilienM:
@superseb 3.6

agarstang:
@superseb and Puppet 3.8.4.

I'm actually pulling through the Integer commit and it's still getting stringyfied in my client.json. I'm trying to find any possible local issues that may be causing this.

EmilienM:
@jlambert121 @superseb Look this demo, how it fails: https://asciinema.org/a/acp59fa3309vjnvvhaaotl2re
We need to run Puppet twice, this is a bug. I'm investigating.

EmilienM:
I noticed some things strange in lib/puppet_x/sensu/to_type.rb but I'm not sure. I'll continue later, feel free to also have a look.

EmilienM:
Closing it in favor of https://github.com/sensu/sensu-puppet/pull/456

PR #454, opened 2015-12-07
client_port must be integer

agarstang:


agarstang:
Close this as is duplicate of #453

Issue #447, opened 2015-11-20
Error: Could not convert change 'socket' to string: undefined method `keys' for nil:NilClass


poolski:
As of the latest commit (merging #443), I get the error above every time I run Puppet. 

Any ideas?

zbintliff:
Me too.

zbintliff:
@poolski  I figured it out.  has to do with the "is_to_s" method when there was no bind previously.  So should only effect people updating.

The module itself is ok, its only when it tries to log (hence why it passes spec tests).  My quick and dirty was just to override the function on the host:

```
Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[sensu1b-af93681f.coreeng.mlbam.dev.us-east-1.bamgrid.net]/socket: socket changed '' to {'bind' => '127.0.0.1', 'port' => '3030'}
```

jlambert121:
This should be resolved, let me know if it isn't!

PR #457, opened 2015-12-10
Fixing regression bug.

zbintliff:
When people are updating from old puppet module the socket is '',
which crashes puppet when it tries to #keys method in the is_to_s
method.  Additionally, I think the default change message is good
enough.

Here is how it looks now when upgrading from old puppet module:
`Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[fqdn]/socket: socket changed '' to '{"bind"=>"128.0.0.1", "port"=>3030}'`

Here is how it will look when changing:
`Notice: /Stage[main]/Sensu::Client::Config/Sensu_client_config[fqdn]/socket:
 socket changed {"bind"=>"1.1.1.1", "port"=3030} to '{"bind"=>"128.0.0.1", "port"=>3030}'`

### DIFF
--- a/lib/puppet/type/sensu_client_config.rb
+++ b/lib/puppet/type/sensu_client_config.rb
@@ -53,14 +53,6 @@ Puppet::Type.newtype(:sensu_client_config) do
       value.each { |k, v| value[k] = to_type(v) }
     end
 
-    def is_to_s(hash = @is)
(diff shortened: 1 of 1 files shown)


zbintliff:
Referencing #447

jlambert121:
Thanks!

Issue #399, opened 2015-08-05
Client.json integers are saved as double quoted strings on first run


hhtpcd:
I've noticed that when I provision a client with this module, and include client_keepalives options, on the first run the services will fail to start because the threshold/critical/warning values are double quoted strings and not integers in the JSON. If I do a second puppet apply run then the values are noticed as different and changed to integers in client.json.

I've poked around the providers/types and can see the logic that compares is vs should and enacts changes, but I can't figure out why the initial config is being written without integers. I can re-create this by deleting the contents of client.json and doing another manual puppet run to get double quoted strings then integers. 

I've included the sensu dec below. This is with the latest module pulled today b3afef6710430d88568c26df36dd3abb7c018768 1.5.5 I believe, on Ubuntu 14.04 x64 and Puppet 3.8.1.

``` ruby
class { '::sensu':
    client                      => true,
    server                      => false,
    rabbitmq_user               => 'sensu',
    rabbitmq_password           => 'password',
    rabbitmq_host               => '192.168.0.123',
    rabbitmq_vhost              => '/sensu',
    rabbitmq_port               => '5671',
    subscriptions               => 'sensu-test',
    rabbitmq_ssl                => true,
    rabbitmq_ssl_cert_chain     => "puppet:///modules/base_config/cert.pem",
    rabbitmq_ssl_private_key    => "puppet:///modules/base_config/key.pem",
    use_embedded_ruby           => true,
    sensu_plugin_provider       => sensu_gem,
    sensu_plugin_version        => 'present',
    rabbitmq_reconnect_on_error => true,
    # install_repo                => true,
    purge_config                => true,
    client_keepalive            => {
        "refresh"    => 120,
        "thresholds" => {
            "critical" => 60,
            "warning"  => 30
        }
    }
}
```

hhtpcd:
I've added a client_custom block to the above ::sensu, and numbers in that block are not being double quoted on first run or when no client.json config exists. 

```ruby
client_custom => {
    'things'  => 'happening',
    'more'    => 20,
    'another' => {
        'thing' => 90,
    },
},
```

```json
{
  "client": {
    "name": "computer-name",
    "address": "192.168.131.33",
    "subscriptions": [
      "Webserver"
    ],
    "bind": "127.0.0.1",
    "port": "3030",
    "safe_mode": false,
    "things": "happening",
    "more": 20,
    "another": {
      "thing": 90
    },
    "keepalive": {
      "refresh": "120",
[...]

PR #458, opened 2015-12-10
fix issue #399

bovy89:


zbintliff:
Was fixed in #456. Personally, like your solution better but oh well.

jlambert121:
Thanks!

Issue #461, opened 2016-01-07
redis.json removed on purge { config => true }


jcochard:
Testing in an ubuntu 12.04 vagrant instance that is both server/client, enterprise enabled, the redis.json config is removed when config purge is set to true.  

==> default: Info: Computing checksum on file /etc/sensu/conf.d/redis.json
==> default: Info: /Stage[main]/Sensu::Redis::Config/File[/etc/sensu/conf.d/redis.json]: Filebucketed /etc/sensu/conf.d/redis.json to puppet with sum ca24608064728ad61e8b846b16671979
==> default: Notice: /Stage[main]/Sensu::Redis::Config/File[/etc/sensu/conf.d/redis.json]/ensure: removed

From redis/config.pp:  
```
if $sensu::_purge_config and !$sensu::server and !$sensu::api {
    $ensure = 'absent'
  } else {
    $ensure = 'present'
  }
```

I believe the issue is that with enterprise, we have to set $sensu::server and $sensu::api = false, so likely just need to add in an and !$sensu::enterprise.  I will submit a pull request shortly.

jcochard:
https://github.com/sensu/sensu-puppet/pull/462

jlambert121:
Closed with #462

PR #462, opened 2016-01-08
support purging with enterprise version

jcochard:


jlambert121:
I haven't ever run sensu enterprise, but why shouldn't these be removed if you want to purge the config on enterprise?

jlambert121:
I just now saw #461 which explains, thanks!

PR #464, opened 2016-01-21
Add support for slack channel configuration / check.

oszknagy:
Hi there,

I've added support to the sensu_check provider to support custom channel configuration pre check   
It enables one or more channel to be specified per check.

Please see the multi channel handler script for the reference.
https://github.com/sensu-plugins/sensu-plugins-slack/blob/master/bin/handler-slack-multichannel.rb

Regards,
Oszkar

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,61 @@ This will create the following check definition for Sensu:
       }
     }
 
+
+## Slack multi channel configuration
+It enables to send alert to specific channel per check.
+Please see for the reference the multi channel handler script.
+https://github.com/sensu-plugins/sensu-plugins-slack/blob/master/bin/handler-slack-multichannel.rb
+
+For example:
+Check from hieradata:
+
+
+sensu.yaml:
+```
+::sensu_checks:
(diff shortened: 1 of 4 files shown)


jaxxstorm:
Personally, I'd rather see this not merged. The options that are defined already are directly available inside sensu and are params in the sensu documentation, and anything else can be covered by sensu_custom.

Merging this opens up things like hipchat channels, and other things being added.

jbehrends:
I'd have to agree.  I have Hipchat integrated without any special code in the sensu module.  I don't see why we would need to deviate and add slack specific code to this module.

jlambert121:
Thank you for your time on this, I think this can be achieved using the $config parameter already and opens the door to a nearly limitless set of properties someone may want to define.

oszknagy:
That's right that it is possible to implement with the custom parameter. 
But it looks like the custom parameter has a bug.  
```      custom:
       'slack':
         channels: 
           - '#avc'```
If you pass the above config via hiera it will changes the file on every puppet run .

PR #466, opened 2016-01-21
Redaction support

jaxxstorm:
This PR adds support for password redaction which is part of the sensu-core and has been since the early days.

I've also added a README to make use of it, as this isn't well documented.

I've set the value to undef for the time being but I think it could be debated that we should set "password" as the default value.

Any thoughts or feedback is very welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,6 +437,47 @@ by using the `sensu_gem` package provider:
       provider => sensu_gem,
     }
 
+## Sensitive String Redaction
+
+Redaction of passwords is supported by this module. To enable it, pass a value to `sensu::redact`
+and set some password values with `sensu::client_custom`
+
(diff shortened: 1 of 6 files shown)


jaxxstorm:
ping @jamtur01 or @jlambert121 ?

jlambert121:
I've never used redact because I couldn't find good documentation on it. :)  It does seem like `password` would be a valid default, but I'm a bit torn since it does change existing functionality.

jaxxstorm:
I'd like to get this merged as we rely on the mainline sensu module for this. We can add defaults later.

ping @jlambert121

jaxxstorm:
ping again @jamtur01 or @jlambert121

PR #467, opened 2016-01-26
add create_resources() for mutators the same way for handlers, checks

EslamElHusseiny:
[x] add create_resources() for mutators the same way for handlers, checks


jlambert121:
Thanks!

PR #471, opened 2016-02-01
Add support for using the same source for  different sensu handlers

salimane:
Currently it's impossible to use the same source for different handlers. a duplicate file definition exception  is thrown. This PR fixes that. With this PR, you can now do:

```yaml
---
sensu::handlers:
  'hipchat_main_room':
    type: pipe
    source: "puppet:///modules/sensu_module/community-plugins/handlers/notification/hipchat.rb"
    config:
      apikey: my_long_key
      room: Big Alerts
  'hipchat_other_room':
    type: pipe
    source: "puppet:///modules/sensu_module/community-plugins/handlers/notification/hipchat.rb"
    config:
      apikey: my_other_long_key
      room: 'Small Alerts'
```

cc @jlambert121 
Thanks

### DIFF
--- a/manifests/handler.pp
+++ b/manifests/handler.pp
@@ -123,13 +123,13 @@ define sensu::handler(
     $filename = inline_template('<%= scope.lookupvar(\'source\').split(\'/\').last %>')
     $handler = "${install_path}/${filename}"
 
-    file { $handler:
+    ensure_resource('file', $handler, {
       ensure => $file_ensure,
       owner  => 'sensu',
       group  => 'sensu',
       mode   => '0555',
       source => $source,
-    }
+    })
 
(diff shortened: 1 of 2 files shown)


jlambert121:
I see the issue, but I'm struggling on putting each handler into its own directory for this somewhat edge case (I say edge because I haven't seen anything about hitting this issue before).  What about just wrapping the file resource in handler.pp with an ensure_resource instead?  I don't normally like using that because I don't think the code is as explicit, but this seems like it could be a prime use case.

salimane:
@jlambert121 , I've updated the changes and use ensure_resource instead.
Thanks

jlambert121:
Thanks!

PR #477, opened 2016-02-22
Add tests when using checks parameter in init

dmsimard:
This adds missing tests that the parameter "checks" and "check_defaults" used in init work as intended.


jlambert121:
I love tests!  Could you rebase?

PR #478, opened 2016-02-22
Add filters and filter_defaults to init with create_resources, missing puppetdoc

dmsimard:
This allows to create a batch of filters just like we have support
for doing so with checks and handlers.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -355,6 +355,8 @@ class sensu (
   $handler_defaults            = {},
   $checks                      = {},
   $check_defaults              = {},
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -355,6 +355,8 @@
   $handler_defaults            = {},
   $checks                      = {},
   $check_defaults              = {},
+  $filters                     = {},
```
EmilienM: it would be great to have documentation for the 2 new parameters.

jlambert121:
Thanks!

PR #479, opened 2016-02-24
Feature prefetch attribute

chrissav:
I was getting an error trying to set the [Rabbitmq prefetch attribute](https://sensuapp.org/docs/latest/rabbitmq#anatomy-of-a-rabbitmq-definition) in the ::sensu class.
Added the [Rabbitmq prefetch attribute](https://sensuapp.org/docs/latest/rabbitmq#anatomy-of-a-rabbitmq-definition), following the same format as the other attributes (such as rabbitmq_port), and got it to update the /etc/sensu/conf.d/rabbitmq.json with prefetch successfully when running on this branch.  
Set the default to 1 since that's what it is in the Sensu docs.
Added rspec tests and verified they're passing.


jlambert121:
Thanks!

PR #480, opened 2016-02-25
add support for configuring sensu-enterprise-dashboard ssl listener

cwjohnston:
Uchiwa 0.14.2 and Sensu Enterprise Dashboard support a native SSL listener. This change adds support for specifying ssl configuration in sensu_enterprise_dashboard_config as a Hash.


jlambert121:
Thanks!

Issue #269, opened 2014-11-14
Support for multiple broker connection options with RabbitMQ


royalj:
Since Sensu 0.15 multiple broker connections with RabbitMQ have been supported to provide cluster failover without the need for a load balancer. It would be nice if this could be supported in the puppet module as well so we can start utilizing this feature.

jamtur01:
I haven't had a chance to address this and it's unlikely I will. I'd be happy to take a patch replacing the provider with a template.

agoddard:
@dblessing did you end up having a play with this?

roobert:
Hi, I came here looking for this feature since it's now the recommended method to do HA Sensu.

I forked 2.0.0 (I tried forking master but was getting various errors) and adjusted the `sensu::rabbitmq::config` class to use a template if `sensu::rabbitmq_hosts` is set which should maintain backwards compatibility for existing deployments and allow people to define multiple servers if they need to.

In terms of implementation detail: the template first checks for unique configuration per-node and then falls back to checking for common values since it's assumed that when using rmq in a cluster configuration you'll have configured each server with the same certs, etc.

I'm not sure whether I should open a PR for this or not since it probably needs rebase'ing and I haven't added any tests or anything. Opinions welcome.

Cheers,

dalesit:
I've spun up the Vagrant instance, copied in the changes to the module, and added the lines
``` 
   use_embedded_ruby => true,
   rabbitmq_hosts    => ['localhost','puppet'],
```
to the sensu-server.pp file in tests.

Doing a `puppet apply sensu-server.pp` generates this error:
```
Error: Evaluation Error: Error while evaluating a Function Call, Failed to parse template sensu/rabbitmq.conf.erb:
  Filepath: /etc/puppet/modules/sensu/templates/rabbitmq.conf.erb
  Line: 5
  Detail: undefined method `has_key?' for "localhost":String
 at /etc/puppet/modules/sensu/manifests/rabbitmq/config.pp:125:18 on node sensu-server.example.com
Error: Evaluation Error: Error while evaluating a Function Call, Failed to parse template sensu/rabbitmq.conf.erb:
  Filepath: /etc/puppet/modules/sensu/templates/rabbitmq.conf.erb
  Line: 5
  Detail: undefined method `has_key?' for "localhost":String
 at /etc/puppet/modules/sensu/manifests/rabbitmq/config.pp:125:18 on node sensu-server.example.com
```
Any thoughts on where the error might be?

[...]

PR #485, opened 2016-03-08
Configuring sensu to utilise a rabbitmq cluster

dalesit:
This addresses #269

Utilises a boolean `$rabbitmq_cluster` to determine whether to configure a rabbitmq cluster using either `$rabbitmq_cluster_hosts` as an array of servers, (with all parameters the same for all servers as used with `$rabbitmq_host` as at present) or `$rabbitmq_cluster_custom` as an array of hashes which allows specific parameters to be set per host. The `$rabbitmq_cluster_hosts` also works as an array of one, so could eventually subsume the requirement for `$rabbitmq_host`, but has been left in so as to be backwardly compatible.

Both methods utilise a template to build the `/etc/sensu/conf.d/rabbitmq.json config` file.

Vagrant has been configured to create rabbitmq servers on the machines sensu-client and sensu-server, and join them into a cluster. The sensu-client vm utilises the `$rabbitmq_cluster_hosts` method, and the sensu-server vm utilises the `$rabbitmq_cluster_custom` method, which verifies that both methods work.

The `provision_server.sh` script  builds a local copy of the sensu-sensu module to install in the VMs, as that is what we are testing.


roobert:
@dalesit my only criticisms are that it seems somewhat confusing to offer multiple types as a parameter and I think if the `rabbitmq_hosts` parameter is set then it's implied that the configuration is a cluster deployment which negates the need for a boolean switch. I think my original patch is a bit simpler and less confusing than this, but I'd be interested to hear other peoples opinions on this. To recap, my patch does the following:

1) if `rabbitmq_hosts` is a hash then it overrides `rabbitmq_host` and the configuration is considered to be a cluster
2) all original parameters are used as the defaults for each server but can optionally be overridden per-server by settings in the `rabbitmq_hosts` hash

I think the only adjustments I'd make to my original patch are the ones mentioned by @jlambert here: https://github.com/sensu/sensu-puppet/issues/269#issuecomment-189286292 the bug fix you pointed out which i made here: https://github.com/roobert/sensu-puppet/commit/3d8aed119ea1fe437a3783c5258cbd726bab15b8 and any additional tests..

dalesit:
@roobert I agree that we could remove the boolean switch - I was including it to be specific about what was going on. I found your patch confusing, because I was expecting to put in a simple array of hosts (it would be worth changing the description in your init.pp to show `Array of hashes` rather than just `Array` to minimise confusion).  

I think it would be rare that someone would want to have different parameters per cluster node, but I can see it might happen. Thus it seemed to make sense to have a simple array of server names as `$rabbitmq_cluster_hosts` but allow more complex configuration with an array of hashes using `$rabbitmq_cluster_custom` (which is essentially your patch rolled in).

roobert:
@dalesit I can see why it'd be confusing and why you'd expect to only specify a list of hosts. I also agree that in the majority of configurations your rmq servers will all have the same configuration which is why I fall back to reading the defaults.

Personally I wouldn't choose to have two ways to configure the same feature because you could get unexpected behaviour/confusion if both parameters are configured, e.g: which config option takes precedence?

I think the decision that needs to be made is whether the flexibility of allowing per-cluster node customization is considered important or not, and if it isn't, then the `rabbitmq_hosts` parameter can be simplified to accept a list of hosts and the template can be updated to remove the extra complexity of allowing per-host overrides.

gbolo:
Would love to have support for rabbitmq clusters.

dalesit:
Failing checks appear to be due to separate wider issue (see #511), not related to this change.

jaxxstorm:
@dalesit can you rebase with master to make sure the tests pass and then I think this is good to go.

jaxxstorm:
@dalesit it looks to me like #518 achieves the same thing but in a slightly more "ruby like" way. Any thoughts? That one covers both redis/sentinels AND rabbitmq clusters, and if the tests pass I'd be more inclined to accept it.

dalesit:
@jaxxstorm In #269 there was a lot of discussion about the merits of moving the config from a custom provider to a template based configuration - I developed my change on the work that had been done by @roobert. That said, #518 does seem to provide the relevant changes in a nice way that continues down the provider route, and does have the merit of covering redis/sentinels as well.

Much of the work I did on the Vagrant config to start up with a cluster can be changed across easily enough, and I might well do this if it is decided to go with #518.

One change I would make to #518 would be to be explicit in the documentation in init.pp that the parameters `rabbitmq_cluster` and `redis_sentinels` are arrays of hashes, rather than simply putting down `Array`. This would seem to fit more with the format as used elsewhere in the file which explicitly refers to `Array of strings`.

jaxxstorm:
Deprecated in favour of #581

PR #486, opened 2016-03-10
add source to remote_file for urls in plugin.pp

chrissav:
Getting an error when trying to add a new sensu plugin.  I'm defining the plugin as:
```
$url = 'https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb'

  sensu::plugin { $url:
    type               => 'url',
    nocheckcertificate => true,
  }
```
And the error in puppet:
```
Error: Could not set 'present' on ensure: bad URI(is not URI?): at 103: modules/sensu/manifests/plugin.pp
```
* Add the source param to remote_file in the url block to fix this.
* Add rspec tests to check for source


chrissav:
bump

jlambert121:
I'm trying to understand why your URL is having this issue as it does seem to be a valid URI.  I had an odd thing once where there was a unicode character in a URL (not sure how it got there) that caused weird errors - can you make sure the file type is ASCII and that there aren't any "invisible" escape/control characters in the URL?

atrepca:
The remote_file module needs a `source` parameter ([Forge link](https://forge.puppet.com/lwf/remote_file#parameters)):

> `source`: The source location of the file, or where to get it from if it is needed. This should be a URI.

Using the remote_file module directly:
```
remote_file { 'https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb':
  ensure => present,
  path   => '/tmp/check-puppet-last-run.rb',
}
```
Generates the same error:
```
Error: /Stage[main]/Test/Remote_file[https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb]/ensure: change from absent to present failed: Could not set 'present' on ensure: bad URI(is not URI?)
```
Adding `source`:
```
remote_file { 'https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb':
  ensure => present,
  source => 'https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb',
  path   => '/tmp/check-puppet-last-run.rb',
}
```
Works as intended:
```
Notice: /Stage[main]/Test/Remote_file[https://raw.githubusercontent.com/sensu-plugins/sensu-plugins-puppet/master/bin/check-puppet-last-run.rb]/ensure: created
```

The call in [package.pp](https://github.com/sensu/sensu-puppet/blob/c53ac5c36e2b76ae3e646c5643291765ee600548/manifests/package.pp#L51-L55) looks right. The call in `plugin.pp` should probably be:
```
remote_file { "${install_path}/${filename}":
  ensure   => present,
  source   => $name,
  checksum => $pkg_checksum,
  require  => File[$install_path],
} ->
```

atrepca:
Bump.

chrissav:
@jlambert121 I've encountered that issue before - with "invisible" characters in the url that sometime cause failures, but that doesn't seem to be the case here.  Puppet runs are failing on this resource only when I add a new plugin, so when it has to fetch it.  For previously existing ones, it sees it's there and moves on I guess.  But I can't get it to work at all without providing the 'source' param to the remote_file module.  By setting it to $name only for url types, the way we define a sensu::plugin would remain the same.

PR #488, opened 2016-03-15
updating package provider to upgrade gems

smbambling:
Currently the sensu_gems package provider doesn't allow gems to be updated, this fixes that.  So if you are bold you can ensure 'latest' on packages or set a newer version.


smbambling:
1.8.7 checks failed need #487 before they pass.

PR #490, opened 2016-03-30
Updated config.pp to add base_path

r0b0tAnthony:
sensu_rabbitmq_config resource type does not a base_path param being passed to it. This causes it default to /etc/sensu/conf.d which breaks Windows support. Added base_path to resource using $sensu::conf_dir, used on the above file resource to ensure the directory.


jlambert121:
Thanks!

PR #494, opened 2016-04-06
Use 127.0.0.1 instead of localhost for hosts, it could resolve to ::1

portertech:
This should stop issues like https://github.com/sensu/sensu/issues/1209


jlambert121:
Thanks!

PR #501, opened 2016-04-21
plugins should be hash and use create_resources

r0b0tAnthony:
The current use of plugins param is an array for files only. This isn't necessarily a problem but the install_path is not dynamic so you can't use it on windows. In general, a better solution is for plugins to be like checks and be a hash. This allow multiple plugin types and allows the install_path to be overwritten on per plugin basis or using plugins_defaults.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -549,7 +549,7 @@ class sensu (
   if $plugins_dir {
     sensu::plugin { $plugins_dir: type => 'directory' }
   } else {
-    create_resources('::sensu::plugin', $plugins, $plugins_default)
+    create_resources('::sensu::plugin', $plugins, $plugins_defaults)
   }
 
 }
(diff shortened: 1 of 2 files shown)


jaxxstorm:
This changes default behavior and therefore needs some tests adding, I can help if required

ghoneycutt:
@Phil-Friderici could you please have a look. I think we should continue to support having a string or array using Puppet4's lambdas. For a hash, we could do the same or use create_resources(), as you see fit.

Phil-Friderici:
Working on this in https://github.com/sensu/sensu-puppet/pull/703

ghoneycutt:
Thanks everyone!

@Phil-Friderici rebased @r0b0tAnthony 's work in PR #703 which has been merged.

ghoneycutt:
Released in v2.11.0

PR #508, opened 2016-06-08
Add FreeBSD support

alexanderdavidsen:
This add FreeBSD support. The only disadvantage is that the sensu package needs to be installed manually or be available in a private repo (since the sensu package is not available via the official FreeBSD pkg sources)


alexanderdavidsen:
The travis builds seems to be failing because of the changes to handler.pp. This will probably be "fixed" by adding 
`let(:facts) { { :osfamily => 'RedHat' } }`
to sensu_handler_spec.rb

This is already done to sensu_check_spec.rb - probably to have the build pass on Windows?

PR #517, opened 2016-07-05
Fix tests

jaxxstorm:


jaxxstorm:
This fixes a bunch of the tests, including the apt failures and failures on older ruby versions.

I'm still working on the `that_comes_before` issues

PR #518, opened 2016-07-11
Rabbitmq Cluster and Redis Sentinels

whermann:
I have created a rabbitmq cluster so that I can add several rabbitmq, the same I did for redis. it runs in both richtunge from one host to two host and vice versa .

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -151,6 +151,10 @@
 #   Integer.  The integer value for the RabbitMQ prefetch attribute
 #   Default: 1
 #
+# [*rabbitmq_cluster*]
+#   Array. Rabbitmq Cluster configuration and connection information for one or more Cluster
+#   Default: Not configured
+#
 # [*redis_host*]
 #   String.  Hostname of redis to be used by sensu
 #   Default: 127.0.0.1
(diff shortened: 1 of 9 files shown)


jaxxstorm:
@whermann can you rebase with master to make sure the tests pass?

jaxxstorm:
@whermann looks like the tests are failing puppet-lint checks, can you update the PR to fix those and then I think this is good to go?

manifests/init.pp
```diff
@@ -151,6 +151,10 @@
 #   Integer.  The integer value for the RabbitMQ prefetch attribute
 #   Default: 1
 #
+# [*rabbitmq_cluster*]
+#   Array. Rabbitmq Cluster configuration and connection information for one or more Cluster
```
dalesit: Would explicitly state that it is an `Array of hashes` to avoid confusion. I would also put in some validation to error with an explanation if it is not an array of hashes (having been bitten by this before).

manifests/init.pp
```diff
@@ -173,6 +177,10 @@
 #   Integer.  The Redis instance DB to use/select
 #   Default: 0
 #
+# [*redis_sentinels*]
+#   Array. Redis Sentinel configuration and connection information for one or more Sentinels
```
dalesit: As above, explicitly state it is an `Array of hashes`. Fits with elsewhere in file where it explicitly refers to `Array of strings` eg subscriptions.

jaxxstorm:
this looks to be the same as #527 but that one has tests. Can you choose one of them, make sure the tests are included and passing, then I'll happily merge

/cc @whermann

mnaser:
I vote to close this one because Sentinel support was already added by another patch.  Let me know if you need help with this project @jaxxstorm

jaxxstorm:
Deprecated in favour of #581

PR #521, opened 2016-07-21
skip ssl config if rabbitmq_ssl = false

eholzbach:
Allow rabbitmq_ssl to be explicitly defined as false. The modules default behavior is that if ssl cert/keys are defined anywhere in hiera, then ssl is reconfigured, regardless if desired or not.

### DIFF
--- a/manifests/rabbitmq/config.pp
+++ b/manifests/rabbitmq/config.pp
@@ -16,88 +16,92 @@ class sensu::rabbitmq::config {
 
   $ssl_dir = "${sensu::etc_dir}/ssl"
 
-  if $sensu::rabbitmq_ssl_cert_chain or $sensu::rabbitmq_ssl_private_key {
-    file { $ssl_dir:
-      ensure  => directory,
-      owner   => $sensu::user,
-      group   => $sensu::group,
-      mode    => '0755',
-      require => Package['sensu'],
-    }
-
-    # if provided a cert chain, and its a puppet:// URI, source file form the
-    # the URI provided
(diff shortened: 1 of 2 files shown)


jaxxstorm:
Can we have some test coverage for these changes?

manifests/rabbitmq/config.pp
```diff
@@ -16,88 +16,92 @@
...
-
-    # if provided a cert chain, and its a puppet:// URI, source file form the
-    # the URI provided
-    if $sensu::rabbitmq_ssl_cert_chain and $sensu::rabbitmq_ssl_cert_chain =~ /^puppet:\/\// {
-      file { "${ssl_dir}/cert.pem":
-        ensure  => file,
-        source  => $sensu::rabbitmq_ssl_cert_chain,
+  if $sensu::rabbitmq_ssl == false {
+    # do nothing
+  } else {
```
fessyfoo: suggestion: use `elseif` and blend into the following conditional?   the existing layers of nesting are complex enough.

eholzbach: You would still need to wrap every reference to rabbitmq_ssl_* in this class with with if rabbitmq_ssl = false do nothing, or change the default behavior of the module.

eholzbach:
With recent additions, there are more problems than this resolves. Need to look into why defining a master with ssl in node/%{::fqdn} hierarchy that hits far before common.yaml which configures clients without ssl config results in the ssl cert/keys defined on clients.

PR #525, opened 2016-07-23
Switch default vhost to /sensu

### LINKED ISSUE #473: Wrong default value of rabbitmq_vhost
Hi,

It seems like a slash has dropped off the default value in https://github.com/sensu/sensu-puppet/blob/master/manifests/init.pp#L304 -- it should probably be `/sensu` not `sensu`.

Setting the value explicitly when calling `class { 'sensu': ... }` solved my keepalive problems :)

 ~llaurén

jaxxstorm:
Fixes #473

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -118,7 +118,7 @@
 #
 # [*rabbitmq_vhost*]
 #   String.  Rabbitmq vhost to be used by sensu
-#   Default: 'sensu'
+#   Default: '/sensu'
 #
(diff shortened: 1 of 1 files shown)


jaxxstorm:
I will merge this in a couple of days unless I see some opposition to the proposed change.

dalesit:
Think this a good idea - at present the workaround has been to specify /sensu - makes sense for the default vhost to match the documentation

PR #529, opened 2016-08-02
Pin listen to a working pre-ruby2.2 version

thejandroman:
It looks like guard requires listen. Listen has been updated to only
support ruby >= 2.2.x. The last version of listen to work on ruby >=
1.9.3 was 3.0.x. This pins the version of listen so development
dependencies can be installed on ruby 2.1.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,8 +35,9 @@ group :system_tests do
 end
 
 group :development do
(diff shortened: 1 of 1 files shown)


thejandroman:
Rebased.

PR #530, opened 2016-08-02
Pin the package provider for RedHat osfamily

thejandroman:
Local tests fail when run on a Mac. This commit forces the package
provider to use yum. Since we're already assuming the osfamily is RedHat
and the default package provider for RedHat is yum, side effects from
this change should be minimal. And local tests will pass on a Mac.

Original error:
`Parameter ensure failed on Package[sensu]: Provider must have features
'upgradeable' to set 'ensure' to 'latest'`


jaxxstorm:
Can you rebase now the tests are passing?

thejandroman:
Rebased.

Issue #533, opened 2016-08-03
Tests still failing


jaxxstorm:
All new PR's are still failing. I thought this was fixed in #517 but it seems not. It needs further investigation.

PR #536, opened 2016-08-16
validate subdue is a hash

fessyfoo:
If subdue is passed, it is expected to be a hash. Validate this is true
to prevent accidentally passing other values which will prevent
sensu-client from starting.


fessyfoo:
> An error occurred while installing json_pure (2.0.2), and Bundler cannot

#533  :(

jaxxstorm:
@fessyfoo yep, it should be fixed now, can you rebase?

PR #537, opened 2016-08-18
Add the ability to add ::apt module from external source

gpenaud:
Hi !

I need to import ::apt class from my own custom modules, because of special parameters. So your way to manage this was a problem for me.
Now, the module is able to use apt functionnalities without  import the module.

This can be inconsistency, but far more flexible.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -18,7 +18,7 @@ class sensu::package {
       if $sensu::manage_repo {
         class { '::sensu::repo::apt': }
       }
-      if $sensu::manage_repo and $sensu::install_repo {
+      if $sensu::manage_repo and $sensu::install_repo and $sensu::manage_apt {
         include ::apt
         $pkg_require = Class['apt::update']
       }
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -60,10 +60,11 @@ class sensu::package {
   }
 
   package { $pkg_title:
-    ensure  => $sensu::version,
-    name    => $pkg_name,
-    source  => $pkg_source,
-    require => $pkg_require,
+    ensure          => $sensu::version,
+    name            => $pkg_name,
+    source          => $pkg_source,
+    install_options => ['--allow-unauthenticated'],
+    require         => $pkg_require,
   }
 
   if $::sensu::sensu_plugin_provider {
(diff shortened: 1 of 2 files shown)


jaxxstorm:
What is this providing that isn't already covered with `manage_repo` ?

gpenaud:
If we want the repo to be managed by the module BUT not the apt external module.

Without manage_repos, we have to install manually the sensu apt source repo. The other option i see is to remove  the line "if defined(apt::source) {" in ::sensu::repo::apt.

jaxxstorm:
I'm not sure I'm following? 

> If we want the repo to be managed by the module BUT not the apt external module.

You've just added a boolean parameter which essentially does the same thing as `$manage_apt` - I don't think this is giving anyone any additional functionality.

gpenaud:
No. The parameter i added ($manage_apt) impacts only the inclusion of external module ::apt.
Your parameter, $manage_repo, impacts both inclusion of external module ::apt AND the management of apt sensu repository. 

There's the difference.

My need is to manage myself the way to include ::apt. Instead of doing 
    include ::apt
I must do
    class { apt:
      purge => {
        'sources.list' => true,
        'preferences' => true
      }
    }

But I still need $manage_repo at true, to automatically manage sensu upstream repository.
I don't know how to explain myself better than this.

manifests/package.pp
```diff
@@ -60,10 +60,11 @@
...
 
   package { $pkg_title:
-    ensure  => $sensu::version,
-    name    => $pkg_name,
-    source  => $pkg_source,
-    require => $pkg_require,
+    ensure          => $sensu::version,
+    name            => $pkg_name,
+    source          => $pkg_source,
+    install_options => ['--allow-unauthenticated'],
```
GhostLyrics: You might want to reconsider this and sign your repository instead of making the default less secure for everyone.

GhostLyrics:
I'm pretty sure you do not need to to this in order to achieve what you want, @gpenaud. 

```puppet
include sensu
# alternatively class{something::sensu: some_param => value}
class 
{ 
  apt:
  purge => 
  {
    'sources.list' => true,
    'preferences' => true
  }
}
contain 'apt'
```

PR #547, opened 2016-09-09
add timeout support for handlers

lobeck:


lib/puppet/provider/sensu_handler/json.rb
```diff
@@ -135,4 +135,12 @@ def subdue
...
   end
+
+  def timeout
+    conf['handlers'][resource[:name]]['timeout']
+  end
+
+  def timeout=(value)
+    conf['handlers'][resource[:name]]['timeout'] = value
+  end
 end
```
jaxxstorm: Can you add a trailing newline here?

manifests/handler.pp
```diff
@@ -62,6 +62,10 @@
 #   Hash.  Handler subdue configuration
 #   Default: undef
 #
+# [*timeout*]
+#   Hash.  Handler timeout configuration
+#   Default: 10
```
jaxxstorm: default is undef as far as I can see?

lobeck: You're right, needs to be undef. We've taken the value from the sensu documentation, but the default only applies for tcp handlers (https://github.com/sensu/sensu/blob/master/lib/sensu/server/handle.rb#L63)

Will switch to `undef` here.

jaxxstorm:
@lobeck can you rebase this and fix the conflicts?

lobeck:
@jaxxstorm done.

PR #555, opened 2016-09-21
Add support for filters option to sensu::mutator

johanek:


johanek:
This is not really needed is it? The new filters behaviour means that the filtering isn't happening in the handler, it's happening in the server and it's stopping the handler & mutator from ever running...

At least I think that's what the case is.

Issue #428, opened 2015-10-05
Unable to set RabbitMQ Heartbeat option


warmfusion:
The RabbitMQ configuration allows a `heartbeat` option to be provided which lets the sensu EM connection to RMQ to be held open even if the particular server gets no traffic, or runs through HAProxy (for example)

Can this option be added to the module?

jlambert121:
This seems like a good parameter - would you take a stab at a PR?

derkgort:
I just added the PR.

jaxxstorm:
resolved by #559

jeffmccune:
Closing this issue via #559 #596 and #623

PR #559, opened 2016-09-30
add support for setting the rabbitmq heartbeat value

derkgort:
This adds the heartbeat setting for rabbitmq, see #428

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,27 @@
+<!--
+Hi there! If you are reporting a bug or a problem, please use this template so that we can collect the information that we need in order to help. If you are opening an issue for a reason other than reporting a problem, e.g. to make a feature request or start a discussion of new functionality, then you do not need to follow the template.
+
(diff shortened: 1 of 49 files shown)


jaxxstorm:
Can you default this to undef? A lot of setups won't have this configured, and I don't think we should change it by default.

jbehrends:
I think having a default value of 30 is ok.  

This is a setting that appears in the [official Sensu documentation](https://sensuapp.org/docs/latest/reference/rabbitmq.html#sensu-rabbitmq-configuration-examples) as a default configuration.  Recently we went through a number of alert keepalive storms in our environment that this setting would have addressed.  As a result I've butt-patched my puppet code with this PR and verified it's working as expected.   

Adding this setting to most installs should have no net effect other than adding a few tcp packets every 30 seconds or so from each client during a internal AMQP heartbeat check.

+1 for this PR as is.

derkgort:
Should I change it to undef?
I've moved the changed to a separate branch and updated to the current master branch.

https://github.com/derkgort/sensu-puppet/tree/feature/heartbeat

jeffmccune:
Closing this pull-request via #596 and #623 

After reviewing this pull request, the desired functionality has been merged and released via #596 and #623.

PR #562, opened 2016-10-18
Filter changes should restart sensu-server service

Oded-B:


solarkennedy:
I believe this will fail on windows due to:
https://github.com/sensu/sensu-puppet/blob/7f60b2c5f708ed03b7769b69a1e86018db598aa9/manifests/server/service.pp#L34

or if someone asked puppet not to manage the services:
https://github.com/sensu/sensu-puppet/blob/7f60b2c5f708ed03b7769b69a1e86018db598aa9/manifests/server/service.pp#L21


What if instead you did this at the puppet (not provider) layer, similar to how extensions notify?
https://github.com/sensu/sensu-puppet/blob/7f60b2c5f708ed03b7769b69a1e86018db598aa9/manifests/extension.pp#L37-L41

ghoneycutt:
@jeffmccune This seems very similar to the code you wrote in ae4885952830868e346eec60b52acfa4c8a28853

ghoneycutt:
Fixed in release v2.20.1

PR #570, opened 2016-10-27
[enterprise dashboard] move package resource inside conditional

cwjohnston:
This change avoids defining the package resource when the enterprise_dashboard parameter is false.

I believe this is desirable because it mirrors the behavior of the enterprise package manifest, making it possible to include this module in another module which manages the sensu-enterprise-dashboard package itself.

cc @dhgwilliam as he committed the original implementation


jaxxstorm:
Looks much better as well 👍

cwjohnston:
@jaxxstorm any reason we ought to hold off on merging this?

Issue #535, opened 2016-08-16
sensu::check unable to remove a check property


fessyfoo:
We (accidentally) set the property subdue (to an empty string) on a `sensu::check` resource and then later tried to remove it by setting the property to `undef`,  however setting a property to undef seems to make the property go unmanaged.   Puppet runs resulted in "no change" to that resource.     I _think_ this is because puppet is [treating an unspecified property as "insync"](https://github.com/puppetlabs/puppet/blob/78e897b/lib/puppet/property.rb#L274) I could be wrong there, hard to follow the puppet code.  

is there a way to remove a property from a previously defined check?

note: setting subdue to something other than a hash prevents sensu-client from restarting.   so as a workaround we have to set subdue to an empty hash.

fessyfoo:
validating that subdue is a hash seems a sensible protection against getting into this situation,  i submitted a PR https://github.com/sensu/sensu-puppet/pull/536

fessyfoo:
given the somewhat vague terms from the documentation.  I think it may be expected behavior for   an unspecified property to go "unmanaged" 

> The `undef` value is usually useful for testing whether a variable has been set. It can also be used as the value of a resource attribute, which can let you un-set any value inherited from a resource default and cause the attribute to be unmanaged

[1] https://docs.puppet.com/puppet/latest/reference/lang_data_undef.html
[2] https://docs.puppet.com/puppet/3.8/reference/lang_datatypes.html#undef

however one might interpret that "unmanaged" in practice dose allow for defaults as some custom types use `defaultto` from the property dsl.    Since a default behavior is acceptable i suggest that `sensu_check`  for properties that are undefined should delete the key from the corresponding sensu json config.

fessyfoo:
in the [latest sensu changes](https://github.com/sensu/sensu-settings/commit/54e6b10988aebac4786e009b28a6b5bdc2e725b3)  subdue is now checked more deeply  and setting to an empty hash is no longer a viable workaround. 

we really need to be able to remove the subdue parameter via puppet.

ttarczynski:
I've used an empty hash as a workaround (to cleanup subdue settings) with sensu version `0.25`:
```
subdue => {},
```

With sensu version `0.26` sensu doesn't accept empty hash for subdue:
```
..."level":"fatal","message":"check subdue days must be a hash","object":{... "subdue":{}...
[...]

PR #572, opened 2016-10-31
use constant SENSU_CHECK_PROPERTIES instead of hardcoded check_args in sensu_check provider

ttarczynski:
This change follows the discussion in #535
And it uses a pattern similar to this:
* https://github.com/puppetlabs/puppet/blob/4.7.0/lib/puppet/provider/yumrepo/inifile.rb#L24


jaxxstorm:
@ttarczynski can you rebase this so the tests pass?

ttarczynski:
I've just rebased, but the tests on Ruby 1.8.7 fails again (another gem dependency is broken).

ttarczynski:
@jaxxstorm this PR is now rebased and ready to merge.

PR #573, opened 2016-10-31
transorm input with munge in type rather than in sensu_check/json.rb provider

ttarczynski:
This change follows the discussion in #535

I don't know how to properly test the provider functionality.
I think that spec tests in this module don't cover the provider implementation (only the type is tested by verifying the resulting catalog).


jaxxstorm:
The test failures here will be fixed by #574 gem pinning

jaxxstorm:
@ttarczynski can you rebase this so the tests pass?

jaxxstorm:
@ttarczynski can you rebase this so the tests pass?

PR #527, opened 2016-07-25
added cluster

whermann:
everything was updated
rebase to master tests ok

PR #574, opened 2016-10-31
Adding support for array of brokers in RabbitMQ config

### LINKED ISSUE #269: Support for multiple broker connection options with RabbitMQ

cwjohnston:
This PR includes the changes proposed in #527 (h/t @whermann), which I have rebased on master and added some spec coverage.

Closes #527
Closes #518
Closes #485
Closes #269

### DIFF
--- a/lib/puppet/provider/sensu_rabbitmq_config/json.rb
+++ b/lib/puppet/provider/sensu_rabbitmq_config/json.rb
@@ -34,7 +34,7 @@ Puppet::Type.type(:sensu_rabbitmq_config).provide(:json) do
   end
 
   def ssl_transport
-    if conf['rabbitmq'].class != Array 
+    if conf['rabbitmq'].class != Array
       if conf['rabbitmq'].has_key? 'ssl'
         :true
       else
@@ -62,8 +62,8 @@ Puppet::Type.type(:sensu_rabbitmq_config).provide(:json) do
       else
         ''
       end
-    else 
-      '' 
+    else
+      ''
     end
   end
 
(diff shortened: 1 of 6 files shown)


cwjohnston:
In reviewing the build failures, it looks like there are a couple of problems here:

1. All Ruby 1.8.7 builds fail. I think this is because metadata-json-lint is pulling in semantic_puppet 0.1.4 which uses the `File.absolute_path` method not available prior to Ruby 1.9.1. I am trying to work out the solution here but specifying semantic_puppet "< 0.1.4" in Gemfile has only produced new errors.

2. Unit test added here fail on builds using Puppet gem version 3.5 without FUTURE_PARSER enabled. Any suggestions here are welcome, as I am a Puppet neophyte. :wink:

portertech:
@whermann @jaxxstorm @jlambert121 any help that you can provide here would be much appreciated :-)

jaxxstorm:
@cwjohnston @portertech I've only taken a cursory look here very quickly, but for whatever reason  the code is evaluating the type/provider config as an array of arrays with the future parser disabled, rather than the expected array of hashes. I've no idea why that would be. I can only imagine this a bug in the code.

Thankyou very much for adding the tests for this, it would clearly have caused some unrest if this was merged without it!

cwjohnston:
Closing this PR in deference to #581 which I think is ready for merge 👍

PR #580, opened 2016-11-08
Use default redact

paramite:
Module should not use different default value than the service
is actually using. This patch changes to default redact value
according to [1]

[1] https://sensuapp.org/docs/0.24/reference/clients.html

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -290,7 +290,7 @@
 #
 # [*redact*]
 #   Array of strings. Use to redact passwords from checks on the client side
-#   Default: []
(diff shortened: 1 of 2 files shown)


jaxxstorm:
Perhaps it's better to set this to `undef` rather than an empty array? I'm happy to set it to the defaults as well, just thinking it's better to avoid unnecessary params in the JSON

paramite:
Sorry, this felt off my radar. Yes, you are probably right. Avoiding explicit configuration will achieve the default value too. Ok, will change the PR.

PR #592, opened 2016-12-02
sensu_check provider: fix missed value

pjfbashton:
This was missed in 4f1fa9fb0ae9e1a029b04a8da4f0dec85e18e9aa


ttarczynski:
@pjfbashton thank you for catching this.
@jaxxstorm this is a needed fix to #573. 

I will try to learn about testing the provider code. Maybe we can catch such bugs with some beaker tests or unit tests for providers.

PR #593, opened 2016-12-09
Fix for enabling strict_variables

madAndroid:
This missing variable is causing failures in our spec tests, since we've recently enabled `strict_variables`  - https://docs.puppet.com/puppet/latest/lang_variables.html#unassigned-variables-and-strict-mode

The change ensures the variable is explicitly set to `undef` if the `source` parameter is not set.
This will have no negative affect on existing implementations.

### DIFF
--- a/manifests/mutator.pp
+++ b/manifests/mutator.pp
@@ -61,6 +61,8 @@ define sensu::mutator(
       mode   => '0555',
       source => $source,
     }
+  } else {
+    $file_ensure = undef
   }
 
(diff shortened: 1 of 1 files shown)


xiii:
Other than Changelog missing looks ok , simple change :) @madAndroid

madAndroid:
@calebhailey - any chance we can get a merge here? It's a really small change to address failures when `strict_variables` config option is enabled - but I don't think the spec tests in travis currently cover that.

PR #594, opened 2016-12-12
Add transport type support for redis/rabbitmq

### LINKED ISSUE #556: Doesn't create a transport.json file
It would seem that the latest version of the sensu requires a transport.json file, configured for either redis or rabbitmq. I suppose in the past it defaulted to rabbitmq and this file wasn't needed, but now it does. 

`{
  "transport": {
    "name": "redis",
    "reconnect_on_error": true
  }
}`

liv3d:
Allows the selection of either Redis or RabbitMQ as the backend for
Sensu.

Fixes #556
Taken the code from #405 and merged in to a single commit

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -192,6 +192,11 @@
 #   Boolean.  Reconnect to Redis in the event of a connection failure
 #   Default: true
 #
+# [*transport_type*]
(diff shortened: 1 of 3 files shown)


liv3d:
I'm going to come back and look at this later; there are issues with updating the `$_transport_type`

PR #602, opened 2017-01-13
Remove sensu check property with absent

### LINKED ISSUE #535: sensu::check unable to remove a check property

### LINKED ISSUE #601: Source parameter not purged when removed from check
## Description of problem

When adding the source parameter to a check, then removing it later, the source parameter hangs around in the JSON
I expected when refreshing the JSON on the check after removing the source parameter, the source parameter remained. In order to fix this, I had to remove the check entirely

To reproduce, simply define a check:

```
sensu::check { 'test':
  command => '/bin/true'
  source => 'something'
}
```

Then update it to remove the source parameter:

```
sensu::check { 'test':
  command => '/bin/true'
}
```

- Your OS: Linux
- Your Ruby version: 2.1.1
- Your version of Puppet: 3.8.7
- Your version of Sensu: 0.25.2

ttarczynski:
This will allow to remove a check property by setting it's value to `'absent'`.
I've applied the same logic as in #565 (subdue param only) for all params.

This follows the same logic as puppet type `yumrepo`:
* `property => undef` – no changes are made
*  `property => 'absent'` – property is removed from json file

I've based this on puppet type `yumrepo` implementation:
* https://github.com/puppetlabs/puppet/blob/4.7.0/lib/puppet/type/yumrepo.rb#L47-L53
* https://github.com/puppetlabs/puppet/blob/4.7.0/lib/puppet/provider/yumrepo/inifile.rb#L251-L295

This fixes #601, fixes #535


ttarczynski:
I'ts a work in progress, I'll try to add some test coverage for 'absent' values.

ttarczynski:
@jaxxstorm I think this is now ready for merge.

I've added a simple rspec test in [sensu_check_spec.rb](https://github.com/sensu/sensu-puppet/pull/602/files#diff-fb207f496b73a3aece974fb2840289c1).
And ensured that `'absent'` value is handled properly for boolean and array properties (2edd88f).
Also tested locally if values `'absent'` and `undef` behaves as expected for each property.

This PR only adds `:absent` handling, but default values (in [manifests/check.pp](https://github.com/ttarczynski/sensu-puppet/blob/2edd88fdd6b64c077248fb387d11221517472811/manifests/check.pp#L107-L129)) are still `undef` which means "unmanaged".

PR #611, opened 2017-02-27
Select debian/ubuntu release for apt repo

johanek:
It appears that sensu packages are no longer created in the `sensu` apt release on repositories.sensuapp.org - the latest there is 0.2.6-5.

Newer builds are being created for each of the releases ala. trusty, xenial, jessie etc.

Update the apt repo config to support this, and update relevant tests.


jgnagy:
Bump!
Without this simple fix, we can't move beyond 0.26.5

PR #612, opened 2017-02-27
add dist to windows msi url

andyroyle:
Similarly to #611 the windows install url also now has the dist codename in it.

Unfortunately only 2012r2 is supported, so it's simplest (for now) to hard-code the url

🏖

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -52,7 +52,7 @@ class sensu::package {
 
       remote_file { $pkg_source:
         ensure   => present,
-        source   => "http://repositories.sensuapp.org/msi/2012r2/sensu-${sensu::version}.msi",
+        source   => "http://repositories.sensuapp.org/msi/2012r2/sensu-${sensu::version}-x64.msi",
         checksum => $::sensu::package_checksum,
       }
     }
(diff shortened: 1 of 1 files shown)


jaxxstorm:
I would rather not hard code this, it's going to bite us down the line if/when we forget

andyroyle:
So I've updated it to use the `$::architecture` fact for the `x64` suffix, but i'm puzzling over the best way to do the `2012r2` in the path.

1) use `$::os_maj_version` and transform it (i.e. remove the spaces) to use as the url segment
- pros: might not break when new dists are added
- cons: we can't guarantee that new dist codes will follow the same pattern, so it may bite us down the line anyway. Transforming fact values could have unintended consquences.

2) use `$::os_maj_version` inside a case and explicitly map the value to the url segment
- pros: mapping in a case allows us to add new dists with minimal interference
- cons: Will *definitely* need code change to add cases for new dists. Looks nastier.

3) adding a new fact to get the equivalent of `$::lsbdistcodename` for windows
- pros: cleaner, more isolated, can use the same logic for linux and windows
- cons: more invasive and a lot more work (to cover all possible windows versions)

andyroyle:
ok, seems like #619 has solved this problem in a better way 😄 

closing this

PR #613, opened 2017-02-27
Support `ensure` property on sensu::enterprise::dashboard::api

cwjohnston:
This change adds an `ensure` parameter to sensu::enterprise::dashboard::api defined type, making it possible to specify that particular instances should be absent.

Perhaps I am thinking about this the wrong way, but adding the ability explicitly setting ensure to `absent` with these changes is the only way I've found to remove the 'Site 1' and 'Site 2' examples from the default dashboard.json.


jaxxstorm:
LGTM

PR #615, opened 2017-02-28
redhat version fix

andyroyle:
The redhat version numbers now have the os version appended, which when combined with this [version number validation](https://github.com/sensu/sensu-puppet/blob/master/manifests/init.pp#L446) means that `0.28.0-1.el7` is rejected by the module, but `0.28.0-1` won't select the package correctly from the yum repo.

The options are: either remove the version number validation in init.pp, or transform the centos version number to include the `.el5`, `.el6` or `.el7` suffix.

In the interests of making it easier to have one version that applies to all environments (debian, redhat, windows, etc) I've opted for the latter.

I'm not sure what side-effects this might have, so let me know if you think this is a terrible idea.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -36,7 +36,7 @@ class sensu::package {
         'installed' => $sensu::version,
         'present'   => $sensu::version,
         'absent'    => $sensu::version,
-        default     => "${sensu::version}.el${::os_maj_version}"
+        default     => "${sensu::version}.el${::operatingsystemmajrelease}"
       }
 
       $pkg_source = undef
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -31,7 +31,14 @@ class sensu::package {
     'RedHat': {
       $pkg_title = 'sensu'
       $pkg_name = 'sensu'
-      $pkg_version = $sensu::version
+      $pkg_version = $sensu::version ? {
+        'latest'    => $sensu::version,
+        'installed' => $sensu::version,
+        'present'   => $sensu::version,
+        'absent'    => $sensu::version,
+        default     => "${sensu::version}.el${::os_maj_version}"
+      }
+
       $pkg_source = undef
 
       if $sensu::manage_repo {
--- a/spec/classes/sensu_package_spec.rb
+++ b/spec/classes/sensu_package_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe 'sensu' do
-  let(:facts) { { :fqdn => 'testhost.domain.com', :osfamily => 'RedHat', :os_maj_version => 7 } }
+  let(:facts) { { :fqdn => 'testhost.domain.com', :osfamily => 'RedHat', :operatingsystemmajrelease => 7 } }
   directories = [ '/etc/sensu/conf.d', '/etc/sensu/conf.d/handlers', '/etc/sensu/conf.d/checks',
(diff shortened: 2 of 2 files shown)


rgeniesse:
Nice! I couldn't figure out why my runs were failing after updating the repo with yum seeing the new software fine.

The only piece I had to make a modification was here:

`default     => "${sensu::version}.el${operatingsystemmajrelease}"`

Not sure if my puppet version (1.7.1) is causing it, but I don't have a `os_maj_version` fact.

andyroyle:
@rgeniesse cheers for the heads up! `$::operatingsystemmajrelease` seems to work fine for me as well so I'll use that instead 👍

andyroyle:
blurgh. I feared this might happen. I'll take a look and see if I can work out whats wrong

cwjohnston:
@andyroyle @rgeniesse sorry for the confusion here, allow me to explain.

Sensu 0.27 is the first Sensu release built by us using Chef's Omnibus tools. One effect of this change is that version strings in the RPM metadata now include a `.elX` suffix, where X is the major version of Redhat. Switching to Omnibus builds also necessitated this change of repository structures, and in our restructured repository we have renamed old package files for consistency, but have not modified the old package version strings to add the new suffix.

This means

* From 0.27 onward, both file names and version strings in RPM metadata will include the `.elX` suffix
* Versions prior to 0.27 include the `.elX` suffix in their file name but do not include the suffix in RPM metadata

In the Sensu Chef cookbook I've implemented [a helper method for generating appropriate version strings](https://github.com/sensu/sensu-chef/blob/5eed3ed8a64f5e50045e6b5b8e92fe9916f4377e/libraries/sensu_helpers.rb#L138), perhaps a similar implementation might be useful here.

rgeniesse:
Thanks @cwjohnston, that clears up the behavior I am seeing. I just upgraded all my Sensu agents today, mix of CentOS 5/6/7, using the code @andyroyle provided. They all worked.

My only issue now is we have a few Amazon Linux nodes out there, but I haven't had time to dig into how to resolve those. Maybe a fact override somewhere. Or maybe they'll just have to sit on 0.26.5 for now.

Open to any final solution and happy to test people's ideas.

jaxxstorm:
This looks great to me. Can you fix the conflicts?

andyroyle:
rebased 👍

cwjohnston:
@andyroyle @jaxxstorm I think merging this has broken tests introduced in https://github.com/sensu/sensu-puppet/pull/645/:

```
  1) sensu package setting version with redhat platform version suffix should contain Package[sensu] with ensure => "0.29.0-1.el7"
     Failure/Error:
       it { should contain_package('sensu').with(
         :ensure => '0.29.0-1.el7'
       ) }
     
       expected that the catalogue would contain Package[sensu] with ensure set to "0.29.0-1.el7" but it is set to "0.29.0-1.el7.el7"
     # ./spec/classes/sensu_package_spec.rb:53:in `block (5 levels) in <top (required)>'
```

Automatically appending the OS major version without the ability to override it makes it impossible for folks like @rgeniesse to install the `el7` package on many RedHat derivatives, e.g. Amazon Linux where the osmajorversion would be `2007`.

Changes in #645 were intended to allow the version parameter value to include the `elX` suffix so that users of RedHat derivatives can specify the `elX` suffix that aligns with their platform's major version . Maybe there's some middle ground possible between these two approaches?

PR #616, opened 2017-03-01
use gem.cmd instead of gem.bat

andyroyle:
yet more fun and games from the 0.27 update for windows

I've yet to find exactly out why using `gem.bat` fails but using `gem.cmd`, but it's *something* to do with the gems environment.

It's worth noting that `gem.cmd` seems to have been added in 0.27, but it's not clear if it's supposed to replace `gem.bat` or not

### DIFF
--- a/lib/puppet/provider/package/sensu_gem.rb
+++ b/lib/puppet/provider/package/sensu_gem.rb
@@ -13,7 +13,7 @@ Puppet::Type.type(:package).provide :sensu_gem, :parent => :gem do
 
   commands :gemcmd =>
     if RUBY_PLATFORM =~ /cygwin|mswin|mingw|bccwin|winse|emx/
-      "#{ENV['SYSTEMDRIVE']}\\opt\\sensu\\embedded\\bin\\gem.bat"
+      "#{ENV['SYSTEMDRIVE']}\\opt\\sensu\\embedded\\bin\\gem.cmd"
     else
       "/opt/sensu/embedded/bin/gem"
     end


cdenneen:
@andyroyle what failure were you seeing?
First run as noted in #629 this test won't pass because package isn't installed.
After that it should work using the gem.bat just fine. #632 fixes that issue and I'm guessing the reason your original run was failing.

jpuskar:
:+1:  that .bat for me fails, .cmd for me works.
0.28.4.1, fresh install

cdenneen:
@jpuskar can you rebase from current master?
@hprins @cwjohnston think this needs to be fixed or the gem.bat needs to be fixed because it appears to be using wrong Ruby/Gem (not embedded)
Right now the sensu_gem scripts get installed under puppet embedded bin... using the cmd fixes it...

jpuskar:
@andyroyle, see @cdenneen's rebase request.

andyroyle:
I've rebased, but not had a chance to test the latest head of master to see if the issue has been resolved

cdenneen:
I tested this change and it seems to fix the issue of where plugins get installed.

cwjohnston:
Aside from incompatibility with Sensu versions < 0.27., this change seems okay to me 

In my manual testing (without Puppet involved) of latest Sensu 0.29 package, both `\opt\sensu\embedded\bin\gem.cmd` and `\opt\sensu\embedded\bin\gem.bat` successfully install gems into the Sensu embedded Ruby. I am not able to explain why this isn't the case when the .bat file is used by Puppet.

cdenneen:
puppet-agent AIO ships with similar vendored ruby just like sensu
When run via puppet it uses the puppet vendored ruby and places the gems in Puppet rather than embedded.

cdenneen:
@cwjohnston So there is a puppet bug open on this:
https://tickets.puppetlabs.com/browse/PUP-6134
If we use the cmd if fixes issues for > 0.27 so maybe that is acceptable at this point since 0.29 is released?
cc @hprins

jpuskar:
We could use an if, and use the .bat if the .cmd does not exist. That would
cover older sensu versions.

cdenneen:
@jpuskar yeah they would just install in the wrong directory too :)

cdenneen:
#636 hopefully should fix the issue and allow for < 0.27 versions
@cwjohnston @andyroyle @hprins

Issue #606, opened 2017-02-06
Update repository URLs and release new module version


GhostLyrics:
Please update the repository URLs according to the changes to packaging introduced with sensu 0.27.
Currently they still run with e.g. `sensu/main` instead of `trusty` for Ubuntu.

https://sensuapp.org/blog/2017/01/31/sensu-core-v0-27-0.html

GhostLyrics:
Is there any update on this?

jaxxstorm:
@GhostLyrics please send a pull request

goodwolf:
See my PR for yum. I will submit another for deb repos

law:
Any word on this for Ubuntu/debian systems?

jaxxstorm:
@law a PR would be appreciated

goodwolf:
Sorry humans, I was distracted by a field mouse. I will get on this.

dzeleski:
I assume we will need to do the same for windows? 

Repo path now includes a 2012r2 folder: https://sensu.global.ssl.fastly.net/msi/2012r2/ 

https://github.com/sensu/sensu-puppet/blob/master/manifests/package.pp#L55

dzeleski:
@jaxxstorm added some code in my fork to support windows repos: 
https://github.com/dzeleski/sensu-puppet/blob/master/manifests/package.pp#L53-L75

This uses the kernel version so we dont need to rely on stdlib methods, but would require an update to the case if a new kernel was released. Default is set to 2012r2. Any issues with this method of setting the repo?

dzeleski:
@goodwolf The repo adds correctly for yum but the filename is incorrect so it cannot install properly. The version now includes .el(5-7). I had to append that to the sensu pkg_version var for redhat as seen here:
```
'RedHat': {
      $pkg_title = 'sensu'
      $pkg_name = 'sensu'
      $pkg_version = "${sensu::version}.el${::operatingsystemmajrelease}"
      $pkg_source = undef

      if $sensu::manage_repo {
        class { '::sensu::repo::yum': }
      }

      $pkg_require = undef
    }
```

https://github.com/dzeleski/sensu-puppet/blob/test_win_log_rotate/manifests/package.pp#L34

jothoma1:
Any news on this ? it didnt install the latest version of sensu for me

jothoma1:
Someone has an idea ? do i have to use @dzeleski fork ?
Thanks in advance

dzeleski:
[...]

PR #610, opened 2017-02-22
update repository urls yum

goodwolf:
repositories for new repository infrastructure

https://sensuapp.org/blog/2017/01/31/sensu-core-v0-27-0.html

See #606

PR #619, opened 2017-03-15
Add support for new windows repo structure

dzeleski:
Referencing this issue: https://github.com/sensu/sensu-puppet/issues/606

This adds support for the new repo structure for windows systems.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -51,16 +51,16 @@ class sensu::package {
       $pkg_require = "Remote_file[${pkg_source}]"
       
       case $::kernelmajversion {
-        6.0: {
+        '6.0': {
           $win_os_code = "2008"
         }
-        6.1: {
+        '6.1': {
           $win_os_code = "2008r2"
         }
-        6.2: {
+        '6.2': {
           $win_os_code = "2012"
         }
-        6.3: {
+        '6.3': {
           $win_os_code = "2012r2"
         }
         default: {
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -49,10 +49,28 @@ class sensu::package {
       $pkg_name = 'Sensu'
       $pkg_source = "C:\\Windows\\Temp\\sensu-${sensu::version}.msi"
       $pkg_require = "Remote_file[${pkg_source}]"
+      
+      case $::kernelmajversion {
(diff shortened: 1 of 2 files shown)


dzeleski:
@jaxxstorm This is failing on the metadata validation, looks like this line is no longer supported: https://github.com/dzeleski/sensu-puppet/blob/win_log_rotate/metadata.json#L19

Should I pull that line out? Or how would you like to handle that?

cdenneen:
@jaxxstorm are you planning on creating these other folders in the repo in order for this to work?

cwjohnston:
@cdenneen You are correct to pointing out that this change assumes the presence of a number of platform directories which aren't currently part of the repository structure.

FWIW @jaxxstorm is a Sensu user whose contributions are valued by both the Core team and the larger Sensu community, but to my knowledge he is not in a position to manage the core repository directly.

Presently we only build MSIs on Windows 2012r2, with the expectation that the resulting artifacts work both on older and newer Windows systems. Should the need arise to build per-platform-version packages for Windows, we would add additional directories to the repository. With this in mind, the changes in this PR are probably not the right approach at this time.

P.S. If it is the case that these artifacts built on 2012r2 aren't working on other versions of Windows, that something we'd like to address. Please report issues of this nature against the [sensu-omnibus](http://github.com/sensu/sensu-omnibus) project.

cdenneen:
Biggest issue right now is old versions don't have architecture and new releases do in the filename.
So that would need to be tested for as well.
Either that or do away with it all and just have a `windows_download_url` and you just set the full url to filename exactly. This takes care of different folders, naming conventions, etc.

dzeleski:
The same thing happened on rhel . A change for yum was merged that already broke support for versions older then .27. Because those changes were merged I follow the same concept with this PR. 

https://github.com/sensu/sensu-puppet/pull/610

dzeleski:
Also this change: https://github.com/sensu/sensu-puppet/commit/789d915696a5591b6f9df93c13d16d821a422da6

would need to be tweaked a bit to support this PR as there is a conflict. Let me know what you guys want to do to support the new repo structure.

dzeleski:
This was resolved with a redirect in the repo

Issue #618, opened 2017-03-15
Windows - sensu-client.log does not rotate


dzeleski:
## Description of problem

Sensu wrapper service is not configured to rotate logs on windows

## Platform and version information

Occurs on Windows Server 2012R2

## Info

erb template that causes the issue: https://github.com/sensu/sensu-puppet/blob/master/templates/sensu-client.erb

winsw info around logging: https://github.com/kohsuke/winsw/blob/master/doc/loggingAndErrorReporting.md#rotate-mode

Im thinking we add the following to the erb template, this would keep 10x 10MB log files and delete the oldest.
```
<log mode="roll-by-size">
      <sizeThreshold>10240</sizeThreshold>
      <keepFiles>10</keepFiles>
</log>
```

dzeleski:
Adding the logpath option as well:
```
  <logpath>C:\opt\sensu\</logpath>
  <log mode="roll-by-size">
        <sizeThreshold>10240</sizeThreshold>
        <keepFiles>10</keepFiles>
</log>
```

Going to let this burn for a few days and we will see what happens.

jothoma1:
Hi, can you explain where do you put these options ?

dzeleski:
@jothoma1 sure: `C:\opt\sensu\bin\sensu-client.xml` is where it gets added. Youll also need to remove the -l option off of the arguements, if you click on the pull request above you can see the changes that were made.

ghoneycutt:
Thank you @dzeleski for opening this issue and @jothoma1 for your help.

This appears to be solved by PR #621 

Please re-open if this is not the case.

PR #621, opened 2017-03-17
Add support to rotate windows logs

dzeleski:
The windows agent was not rotating logs correctly. I added support for that here, some changes to how logging is done was altered to add this support. Im open to discussion if we think we need to adjust it briefly but there are now separate log files for each type of output.

https://github.com/sensu/sensu-puppet/issues/618

https://github.com/sensu/sensu/issues/1617

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -15,8 +15,7 @@
     { "operatingsystem": "Ubuntu", "operatingsystemrelease": [ "12.04", "14.04" ] }
   ],
   "requirements": [
-    { "name": "puppet", "version_requirement": ">=3.3.0 <5.0.0" },
-    { "name": "pe", "version_requirement": ">=3.7.0 < 2015.3.0" }
+    { "name": "puppet", "version_requirement": ">=3.3.0 <5.0.0" }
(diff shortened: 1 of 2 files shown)


cdenneen:
@dzeleski sent you PR

dzeleski:
@cdenneen Cool, ty. Were you able to test those changes out? I wont be able to get to testing it until Monday.

cdenneen:
@dzeleski I didn't write specific tests but it's passing travis tests that are there.

dzeleski:
@cdenneen sorry I mean did you run the code in puppet on some nodes? I assume it works based on looking at it. Ill be able to pull the code down Monday and test in my lab regardless.

cwjohnston:
I think this looks good. 

For posterity I'll point out that this log rotation implementation relies on the behavior of the winsw service wrapper. Instead of specifying `--log FILE` as a flag for the Sensu daemon, the rendered XML will start the service daemon with stdout and stderr logging captured by winsw instead of being written directly by the daemon as is done on Linux platforms.

PR #596, opened 2016-12-21
sensu-puppet-add heartbeat feature

derkgort:
add support for setting the rabbitmq heartbeat value #559

PR #623, opened 2017-03-19
Add fix to resolve rabbitmq cluster heartbeat config failure.

dzeleski:
This PR fixes a bug introduced by this change: https://github.com/sensu/sensu-puppet/pull/596

When a cluster is used the single node heartbeat var was conflicting with the cluster setting. This PR fixes that issue by mimicking what was done with the prefetch var. 

The default is set to undef as I believe it should not be set by default. Sensu's documentation states that its default is to not be set which disables heartbeats and I would say we should follow that. The config example in the docs shows how to set it if needed: https://sensuapp.org/docs/latest/reference/rabbitmq.html#rabbitmq-attributes

### DIFF
--- a/lib/puppet/provider/sensu_rabbitmq_config/json.rb
+++ b/lib/puppet/provider/sensu_rabbitmq_config/json.rb
@@ -150,11 +150,11 @@ Puppet::Type.type(:sensu_rabbitmq_config).provide(:json) do
   end
 
   def heartbeat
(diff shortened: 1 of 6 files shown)


cwjohnston:
This change seems 👍  to me, and 11a4769 itself will unblock a bunch of PRs currently failing due to changes in metadata-json-lint. @jaxxstorm any thoughts?

Issue #617, opened 2017-03-08
Windows - attempts to create a local 'sensu' user


Justin-DynamicD:
## Description of problem

Running a basic class to install the client on windows will result in and attempt to create a local user named "sensu".  Any password complexity rules set on the host will force an error,  yet I can't find any parameters within the class to manually override the defaults for this account (neither name nor password).  furthermore, manual install of the agent succeeds without the need to even create said local account, which leads me to question it's purpose? (I am new to sensu, so any insight would be helpful if I'm mistaken)

- Is this a master based puppet setup, btw, with a simple single puppet master.
 - module version 2.2.0

dzeleski:
This was fixed here: https://github.com/sensu/sensu-puppet/pull/588 

New version has not been cut however but you can pull that commit and at least have a working module for windows.

Justin-DynamicD:
Thank you for this.  I was able to "temporarily" get around the problem by using $manage_user false then creating the account with other modules where i could control the password which was enough to get myself going (dsc_user class, call the user "sensu" and then the module works fine).

dzeleski:
In that case you would need to make sure the sensu user has the correct system permissions to run the checks needed. Of course it will work fine and I do plan to add better support for that in a few weeks, just have not had a moment to spare for it. 

Regardless no problem and glad you at least got it working :)

cdenneen:
@dzeleski  I don't think this is fixed... I'm pulling latest (with the above mentioned fix) and getting the following:
```
  Error: /Stage[main]/Sensu::Package/User[NT Authority\SYSTEM]: Could not evaluate: ADSI connection error: failed to parse display name of moniker `WinNT://./SYSTEM,user'
      HRESULT error code:0x800708ad
        The user name could not be found.
  Wrapped exception:
  failed to parse display name of moniker `WinNT://./SYSTEM,user'
      HRESULT error code:0x800708ad
        The user name could not be found.
```
cc @cwjohnston

dzeleski:
@cdenneen need to set manage_user to false for windows systems. I plan on adding support for sensu to be able to manage a local user but I have not gotten around to it. I mentioned that setting on the pull request, the default I believe is true.

cdenneen:
Can you update PR to have false based on osfamily in params?

dzeleski:
[...]

Issue #586, opened 2016-11-21
Windows: Fails to create sensu user


dzeleski:
@jaxxstorm Issue found after resolving this issue: https://github.com/sensu/sensu-puppet/issues/569 

Not getting that error after a restart but now its failing on either creating the sensu user, or if I disable manage user it cannot find the sensu user or use the built in administrator user. 

With manage_user set to false: 
```
Error: Could not find user Sensu
Error: /Stage[main]/Sensu::Package/File[C:/opt/sensu/conf.d]/owner: change from BUILTIN\Administrators to Sensu failed:
Could not find user Sensu
```

true:
```
Error: User update failed: (in OLE method `SetInfo': )
    OLE error code:800708C5 in Active Directory
      The password does not meet the password policy requirements. Check the minimum password length, password complexit
y and password history requirements.

    HRESULT error code:0x80020009
      Exception occurred.
Wrapped exception:
(in OLE method `SetInfo': )
    OLE error code:800708C5 in Active Directory
      The password does not meet the password policy requirements. Check the minimum password length, password complexit
y and password history requirements.

    HRESULT error code:0x80020009
      Exception occurred.
Error: /Stage[main]/Sensu::Package/User[Sensu]/ensure: change from absent to present failed: User update failed: (in OLE
 method `SetInfo': )
    OLE error code:800708C5 in Active Directory
      The password does not meet the password policy requirements. Check the minimum password length, password complexit
y and password history requirements.

    HRESULT error code:0x80020009
      Exception occurred.
```

The biggest issue is that I cannot find any documentation on how to setup a windows client via the sensu puppet module, I also dug through some of the code and I dont see any parameters on setting the sensu user/password for the service account. I can obviously create a sensu user but I feel like the sensu module should be able to handle that no?

Also my module that I wrote to get around this issue registers the sensu service as local system, is there any reason why this shouldnt be done? Most solutions I see on windows should be totally happy as a local system service.

```
  exec {'create_sensu_client_service':
    command  => 'new-service -name sensu-client -BinaryPathName "c:\\opt\\sensu\\bin\\sensu-client.exe" -DisplayName "Sensu Client"',
    unless   => 'if (Get-Service sensu-client -ErrorAction SilentlyContinue) {exit 0} else {exit 1}',
    provider => powershell,
    require  => Dsc_package["sensu-${sensu_version}.msi"],
  }
```

dschaaff:
[...]

PR #628, opened 2017-03-29
Disable user creation on osfamily = windows by default

cdenneen:
This is to address #617

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -162,7 +162,7 @@ class sensu::package {
     }
   }
 
-  if $sensu::manage_user {
+  if $sensu::manage_user and $osfamily != 'windows' {
     user { $sensu::user:
       ensure  => 'present',
       system  => true,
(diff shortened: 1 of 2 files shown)


cwjohnston:
I understand that the change here is motivated by factors described in https://github.com/sensu/sensu-puppet/issues/586#issuecomment-290190265. 

@jaxxstorm do you have any concerns about this from a SemVer perspective? I'm not sure if this change would require a major version bump, or what the plans are for the next release.

jaxxstorm:
I've been trying to track major versions with sensu versions, I think we can probably get away with doing a minor bump here as long as we make it clear what's happened in the changelog.

cwjohnston:
@dzeleski I take it from [your comment in #617](https://github.com/sensu/sensu-puppet/issues/617#issuecomment-289966990) that you're 👍  on this change? If so I think we can merge it.

dzeleski:
@cwjohnston In my lab, running on about 30 systems. I try not to run things that are not in the release branch in prod. No issues thus far.

PR #635, opened 2017-04-11
Windows Support Fixes

jmkeyes:
This PR provides fixes for two issues when using this module in a Windows environment:

1. The package title in `Class[sensu::package]` is "Sensu" instead of "sensu" which causes a compilation failure due to resource references to `Package[sensu]`. Commit 69b01e7 unifies the use of `$sensu::package::pkg_title` across the module.
2. The RabbitMQ configuration class manages a directory for storing SSL certificates, but hard-codes the permissions on that directory to `0755` on all platforms which results in unusable folder permissions and a catalog application failure. Commit c1c6ef5 switches the hard-coded permissions to use `$sensu::dir_mode` instead.

With these two fixes applied this module works perfectly on my Windows server deployments.

Let me know if there's anything else that needs to change.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -97,7 +97,7 @@ class sensu::package {
       owner   => '0',
       group   => '0',
       mode    => '0444',
-      require => Package['sensu'],
+      require => Package[$pkg_title],
     }
   }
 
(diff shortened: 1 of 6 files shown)


cwjohnston:
Closes #646

PR #636, opened 2017-04-12
Adding a gem.bat to fix the ruby.exe path issue

cdenneen:


cwjohnston:
We've merged changes in https://github.com/sensu/sensu-omnibus/pull/203 that I believe will make it unnecessary for puppet to directly manage this file. I've kicked off builds for 0.29.0-9 packages so we may test that theory.

cdenneen:
@cwjohnston great news but I'm slightly confused.
- you didn't want to move to gem.cmd because this would break installs for versions <0.27
- without doing this change it will only fix versions >0.29

So this fix will work with all older versions... and would still work with 0.29
At the point that you've decided to deprecate older than 0.27 the bat file wouldn't be necessary and the cmd PR #616 would work for >0.27

I knew you wouldn't go back and release older versions updating the bat file so figured this was a better solution to cover all scenarios... you're call obviously.

cwjohnston:
> you didn't want to move to gem.cmd because this would break installs for versions <0.27

To be clear, I'm not fully 👎  on using gem.cmd over gem.bat, I am merely calling out risks I see and trying figure out what will cause the least confusion. 

To me there's still an open question of how `gem.cmd` came to be part of the package build. I believe it was added by upstream Ruby when we moved from 2.0 to 2.3.0 circa Sensu 0.27. As I continue to look into this I see that `gem.cmd` is present in 0.27, 0.28 and 0.29 packages and isn't being explicitly added by our omnibus build toolchain. I think we can merge #616.

> So this fix will work with all older versions... and would still work with 0.29

I am reluctant to endorse unconditionally overwriting a file provided by official packages, as it adds a corner case we have to consider should changes to that aspect of packaging be needed in the future.

> I knew you wouldn't go back and release older versions updating the bat file so figured this was a better solution to cover all scenarios...

I agree, we won't be back-porting a fix and releasing new packages for these older versions.

cdenneen:
@cwjohnston ok whatever the internal team feels is best... 
#616 will break versions older than 0.27 as you've pointed out.
Yes this creates a corner case but until < 0.27 are officially deprecated from support I think it might become a necessary evil in order to have a module that works in those cases. 

I see 3 options:


**1. Officially stop supporting versions < 0.27**
- Merge #616 

**2. Support older versions**
- Merge #636 
  a. When official support deprecates for versions < 0.27****
    - Revert #636 
    - Merge #616 

  b. When official support deprecates for versions < 0.29.0-9****
    - You can either leave #616 since cmd is present or Revert #616 since the bat file is now fixed.

**3. Drop support for all versions < 0.29.0-9**
- No PR merge necessary

@jaxxstorm @hprins

cwjohnston:
> I see 3 options

I agree with your analysis. I've merged #616 to address this.

PR #644, opened 2017-04-24
feat: update apt/yum repos

bdossantos:
Hi,

I've updated yum/apt repository location, according documentation.

This also permit to get latest sensu version, v0.29

https://sensuapp.org/docs/0.29/platforms/sensu-on-ubuntu-debian.html#install-sensu-core-repository

Next step: ensure apt/yum https transport package is installed then use the HTTPS endpoint by default.

Thank you

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -61,7 +61,7 @@
 #
 # [*repo_key_source*]
 #   String.  URL of the apt GPG key
-#   Default: http://repositories.sensuapp.org/apt/pubkey.gpg
+#   Default: http://sensu.global.ssl.fastly.net/apt/pubkey.gpg
 #
 # [*client*]
 #   Boolean.  Include the sensu client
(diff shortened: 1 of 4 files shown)


bdossantos:
Oups, broken specs. I'll try to fix it.

jaxxstorm:
Any luck here?

ghoneycutt:
Looks good, though tests are failing. Think it is related to http/https change.

manifests/init.pp
```diff
@@ -360,7 +360,7 @@
   $repo                           = 'main',
   $repo_source                    = undef,
   $repo_key_id                    = 'EE15CFF6AB6E4E290FDAB681A20F259AEB9C94BB',
-  $repo_key_source                = 'http://repositories.sensuapp.org/apt/pubkey.gpg',
+  $repo_key_source                = 'https://sensu.global.ssl.fastly.net/apt/pubkey.gpg',
```
ghoneycutt: We need to keep this http for now to ensure compatibility.

bdossantos: Hi, sorry for the late reply, I'll fixup tomorrow.

cwjohnston: @ghoneycutt https with `sensu.global.ssl.fastly.net` is OK. 

We need to be careful with https on `repositories.sensuapp.org`, as the cert being provided on that CDN map uses SNI (subjectAltName, not supported by all package managers) and requires TLS 1.2 (not supported by all openssl versions)

bdossantos:
@ghoneycutt @cwjohnston done

ghoneycutt:
Rebasing and testing

ghoneycutt:
Thank @bdossantos for your contributions! Much appreciated.

cwjohnston:
I believe the URL scheme needs to change from http to https if we want to realize a benefit from this.

ghoneycutt:
This gets us on the correct hosts. I'll change the protocol next.

ghoneycutt:
Rebased in #696

PR #653, opened 2017-05-09
[WIP] check type should notify sensu-enterprise

### LINKED ISSUE #495: $check_notify does not load sensu::enterprise::service
When you set `sensu::check`, this module does not notify `Service["sensu-enterprise"]`.

I've look at [code](https://github.com/sensu/sensu-puppet/blob/master/manifests/init.pp#L442), there is "ugly hack" but that hack does not work with sensu-enterprise. 
If needed, I can make PR for this but just tell me if I'm on right direction. 
How `Service["sensu-enterprise"]` should get notified when check is added?

EDIT: 
Maybe I should be more descriptive, so if you have understand what is the problem, you might stop reading after this point. 

When I add new check ex:

``` puppet
sensu::check { 'check_cpu':
    command     => 'check-cpu.rb -w 85 -c 95',
  }
```

and run `puppet apply` returns this 

``` log
Notice: /Stage[main]/Fb_sensu::Checks::System/Sensu::Check[check_cpu]/Sensu_check[check_cpu]/ensure: created
Notice: /Stage[main]/Sensu::Client::Service/Service[sensu-server]: Triggered 'refresh' from 2 events
```

The problem here is that `sensu-server` is being refreshed and not `sensu-enterprise` 

EDIT: relevant PR #388

dhgwilliam:
I believe this resolves #495 in spirit, but I'm still working on writing a good test. It may be preferable to implement Puppet 4's `autonotify()` now that we've dropped support for Puppet 3...

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -24,6 +24,7 @@ Puppet::Type.newtype(:sensu_check) do
     self[:notify] = [
       "Service[sensu-client]",
       "Service[sensu-server]",
+      "Service[sensu-enterprise]",
     ].select { |ref| catalog.resource(ref) }
   end
 


ghoneycutt:
Released in v2.19.1

PR #667, opened 2017-06-17
[WIP] Add a sensu::write_json defined resource type, for custom configs

robbyt:


manifests/write_json.pp
```diff
@@ -0,0 +1,86 @@
...
+#   Default: true
+#
+# [*content*]
+#   Hash. The hash content that will be converted to json and written into
+#         the target config file.
+#   Default: {}
+#
+define sensu::write_json(
+  Enum['present', 'absent'] $ensure = 'present',
+  Optional[String]          $mode = undef,
```
ghoneycutt: why not just

```puppet
String $mode = '0755'
```

robbyt:
@ghoneycutt I added that selector for default file mode to work around some past oddness with file modes on Windows. It's probably unnecessary with modern versions of Puppet, so I'll simplify it as you suggested.

Additionally, the tests are failing because of the change introduced in https://github.com/sensu/sensu-puppet/pull/615 (see the last comment in the PR).

ghoneycutt:
That fix is for versions older than .27 and we're at .29. Wonder if we should even support those old versions.

robbyt:
I hit an issue with my rspec error matchers- Puppet changed the error message in `assert_type`, and so the rspec tests are failing only on some Puppet versions. My latest change should fix this.

If my latest changes to the new tests pass, this will be ready for a squash/merge.

manifests/write_json.pp
```diff
@@ -0,0 +1,73 @@
...
+#   Default: true
+#
+# [*content*]
+#   Hash. The hash content that will be converted to json and written into
+#         the target config file.
+#   Default: {}
+#
+define sensu::write_json(
+  Enum['present', 'absent'] $ensure = 'present',
+  Optional[String]          $mode = '0755',
```
ghoneycutt: I don't think this should be optional after you refactor. Could you order it after group, so it reads owner/group/mode. Same in the file resource below.

manifests/write_json.pp
```diff
@@ -0,0 +1,73 @@
...
+  # Write the config file, using the native file resource and the
+  # sensu_sorted_json function to format/sort the json.
+  file { $title :
+    ensure  => $ensure,
+    mode    => $mode,
+    owner   => $owner,
+    group   => $group,
+    content => sensu_sorted_json($content, $pretty, 4),
+  }
+
```
ghoneycutt: style: no newline here

spec/defines/sensu_write_json_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        :pretty => false,
+        :content => CUSTOM_CFG,
+      } }
+
+      it { should contain_file(CFG_PATH).with(
+        :ensure => 'present',
+        :mode => '0755',
+        :owner => 'root',
+        :group => 'root',
+        :content => "{\"custom\":\"data\"}",
```
ghoneycutt: Here's a way to do the HEREDOC so that you dont have all the escaping.

https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/defines/acl_spec.rb#L22-L35

tests/sensu-write-json.pp
```diff
@@ -0,0 +1,21 @@
+$h = {
```
ghoneycutt: Could you please update the acceptance test to apply this code and check that the file is formatted correctly. You could use diff, md5, however you want to do comparison.

robbyt: @ghoneycutt I'm not sure how to do this- do you know of an example?

robbyt:
@ghoneycutt other than the file check in the one test, I think this is ready for merge. Can you please confirm and squash/merge when you can?

ghoneycutt:
Closing in favor of #671 which has been rebased with additional changes.

PR #668, opened 2017-06-19
TravisCI to explicitly test supported versions of Puppet

ghoneycutt:
Puppet and ruby versions taken from
https://docs.puppet.com/puppet/4.10/about_agent.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,45 @@
 ---
 language: ruby
+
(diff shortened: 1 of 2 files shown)


ghoneycutt:
Seems to not respect the environment variable and use the newest version of puppet

ghoneycutt:
Currently Travis is using the newest puppet and not honoring the intent of using different versions.

PR #692, opened 2017-06-28
Add support for Ubuntu 16.04 LTS

ghoneycutt:
This commit adds Ubuntu 16.04 and 14.04 as clients in Vagrant.


ghoneycutt:
Released in v2.6.0

Issue #688, opened 2017-06-27
the sensu-plugin gem is incorrectly installed with the system ruby instead of the embedded ruby


ghoneycutt:
We should install the plugin and anything else using the embedded ruby and not the system ruby.

jeffmccune:
@ghoneycutt Is the scope of this the default behavior of the module to be released to end users, or restricted to the vagrant environment, e.g. `tests/sensu-client.pp` which would need `sensu_plugin_provider => sensu_gem` added?

jeffmccune:
[PR-694](https://github.com/sensu/sensu-puppet/pull/694) is ready for review.

ghoneycutt:
Not just vagrant, you're on the right track here by only using the embedded ruby.

ghoneycutt:
Released in v2.7.0

PR #694, opened 2017-06-28
(GH-688) Default sensu-plugin gem to use sensu_gem provider

jeffmccune:
Without this patch the behavior of sensu-plugin is managed by the default
package provider for the system, e.g. gem.  This patch changes the default
behavior to use the embedded ruby version at `/opt/sensu/embedded/bin/gem` via
the built in provider.

### DIFF
--- a/spec/classes/sensu_package_spec.rb
+++ b/spec/classes/sensu_package_spec.rb
@@ -40,7 +40,6 @@ describe 'sensu' do
 
         it { should contain_package('sensu-plugin').with(
           :ensure   => 'installed',
-          :provider => 'gem'
         ) }
       end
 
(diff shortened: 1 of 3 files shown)


jeffmccune:
resolves #688

jeffmccune:
Note, this PR also changes `Sensu::Package/File[/etc/default/sensu]/content`, setting `EMBEDDED_RUBY=true`

spec/classes/sensu_package_spec.rb
```diff
@@ -56,15 +55,26 @@
 
         it { should contain_package('sensu-plugin').with(
           :ensure   => 'installed',
-          :provider => 'gem'
```
ghoneycutt: should this be changed to sensu_gem instead of removed?

spec/classes/sensu_package_spec.rb
```diff
@@ -40,7 +40,6 @@
...
```
ghoneycutt: should this be changed to sensu_gem instead of removed?

spec/classes/sensu_package_spec.rb
```diff
@@ -56,15 +55,26 @@
...
+        it { should contain_package('sensu-plugin').with(:provider => 'sensu_gem') }
+      end
+
+      context 'with use_embedded_ruby => true' do
+        let(:params) { { :use_embedded_ruby => true } }
+
+        it { should contain_package('sensu-plugin').with(:provider => 'sensu_gem') }
+      end
+
+      context 'with use_embedded_ruby => true' do
```
ghoneycutt: context should read `false`

ghoneycutt:
The docs should also mention that we use the embedded ruby by default and do not recommend using the system ruby.

@cwjohnston @agoddard should we even allow system ruby?

spec/classes/sensu_package_spec.rb
jeffmccune: I thought about that, but I think specifying the provider here is superfluous since the context of the test, "setting version with / without platform suffix" doesn't seem to relate behaviorally to the provider.  I would have left it the provider wasn't fully covered in the embedded_ruby context.  Happy to add them back if you prefer.

spec/classes/sensu_package_spec.rb
jeffmccune: same reasoning as line 43

spec/classes/sensu_package_spec.rb
jeffmccune: fixed

jeffmccune:
> The docs should also mention that we use the embedded ruby by default and do not recommend using the system ruby.

Docs updated in 44aae14

spec/classes/sensu_package_spec.rb
ghoneycutt: I agree.

ghoneycutt:
Thanks @jeffmccune !

Released in v2.7.0

jeffmccune:
> Thanks @jeffmccune !

You're welcome!  Glad to help out, thanks for the quick review!

PR #696, opened 2017-06-28
(GH-644) Use the new apt and yum repositories

ghoneycutt:
See: https://sensuapp.org/docs/0.29/platforms/sensu-on-ubuntu-debian.html#install-sensu-core-repository


ghoneycutt:
Released in v2.7.0

Issue #697, opened 2017-06-28
Change repos to use HTTPS by default


ghoneycutt:
Currently they are using HTTP.

ghoneycutt:
Fixed in v2.8.0

PR #698, opened 2017-06-28
(GH-697) Use https with public package repositories

ghoneycutt:


ghoneycutt:
Closes #697

ghoneycutt:
Released in v2.8.0

PR #702, opened 2017-07-03
Use rspec-puppet 2.5.x until 2.6.x is fixed

Phil-Friderici:
Latest update to rspec-puppet 2.6.x doesn't work correctly.
Pin to use 2.5.x until 2.6.x is fixed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ end
 
 group :development, :unit_tests do
   gem 'rake',                                             '< 11.0.0'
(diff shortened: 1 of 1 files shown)


jaxxstorm:
LGTM

Phil-Friderici:
This will fix the spec tests. Check https://github.com/Phil-Friderici/sensu-puppet/pull/1 to see that the spec will fail with no change at all.

Phil-Friderici:
@ghoneycutt please have a look for this fix too

PR #703, opened 2017-07-03
Working on PR501

Phil-Friderici:
- Keep backward compatibility
- Updating docs
- Adding tests for https://github.com/sensu/sensu-puppet/pull/501

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,31 @@ are managed with the server, and API parameters.
 
     }
 
+If you need only one plugin you can also use a simple string:
+
+    node 'sensu-server.foo.com' {
+      class { 'sensu':
+        plugins           => 'puppet:///data/sensu/plugins/ntp.rb'
+        ...
+    }
+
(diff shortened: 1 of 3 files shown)


ghoneycutt:
@Phil-Friderici Looks good. Could you please update the README to show that hashes are also allowed. BTW, when referring to an issue, use `(GH-###)` in the commit where ### is the issue number (GH is for GitHub).

You can see previous commits here - https://github.com/sensu/sensu-puppet/commits/master

Phil-Friderici:
@ghoneycutt README is updated. I didn't found a matching issue for this new feature. Did I miss it ?

Phil-Friderici:
Maybe we shouldn't propagate the usage of strings for plugins in the docs explicitly?
I guess on the long run arrays and hashes should keep supported only.

ghoneycutt:
Released in v2.11.0

Issue #710, opened 2017-07-06
Add support for Debian 7 and 8


ghoneycutt:
Module should support Debian 7 (wheezy) and 8 (jessie) which are supported by Sensu.

ghoneycutt:
Released in v2.12.0

Issue #711, opened 2017-07-06
Add ability to specify a different release for apt::source


ghoneycutt:
As described in Issue #708

ghoneycutt:
@debilski

ghoneycutt:
Released in v2.12.0

PR #712, opened 2017-07-06
(GH-710) support debian 7 and 8

ghoneycutt:


ghoneycutt:
Solves #710 and #711

ghoneycutt:
Released in v2.12.0

PR #716, opened 2017-07-06
(#649) Enumerate sensu_enterprise_dashboard_config instances

### LINKED ISSUE #649: implement an `instances` method for the sensu_enterprise_dashboard_api_config `json` provider
This would facilitate, e.g, using the `resources` resource to `purge` any unmanaged API endpoints from the config

See https://docs.puppet.com/puppet/4.10/provider_development.html#providerresource-api

jeffmccune:
Without this patch the `resources` resource cannot be used to purge unmanaged
API endpoints as requested in #649.  This patch implements the instances class
method providing this functionality.

resolves #649

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,4 +50,8 @@ group :development do
   gem 'listen', '~> 3.0.0', :require => false
 end
 
+if File.exists? 'Gemfile.local'
+  eval(File.read('Gemfile.local'), binding)
+end
+
 # vim:ft=ruby
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,8 @@ spec/reports
 test/tmp
 test/version_tmp
 tmp
-Gemfile.lock
+/Gemfile.lock
+/Gemfile.local
 
 # YARD artifacts
 .yardoc
--- a/README.md
+++ b/README.md
@@ -31,7 +31,9 @@ Sensu-Puppet module.
 
 ## Installation
 
-    $ puppet module install sensu/sensu
+```bash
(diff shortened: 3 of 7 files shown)


ghoneycutt:
What's with the empty commit?

jeffmccune:
> What's with the empty commit?

Made the empty commit to trigger Waffle so I'd be assigned the issue and it'd move over to in-progress.

jeffmccune:
This is ready for review / merge.

spec/spec_helper.rb
```diff
@@ -1,4 +1,5 @@
 require 'puppetlabs_spec_helper/module_spec_helper'
+require 'pry'
```
ghoneycutt: This should be removed. It is what's causing Travis to fail.

Gemfile
```diff
@@ -50,4 +50,8 @@ group :development do
   gem 'listen', '~> 3.0.0', :require => false
 end
 
+if File.exists? "#{__FILE__}.local"
```
ghoneycutt: This would be more clear if it read `Gemfile.local` instead of being variable, since the name of the Gemfile will not change.

ghoneycutt:
Could you please add documentation of this new feature in the README.

ghoneycutt:
Thanks @jeffmccune 

Could you add the README portion and then this is good to go.

jeffmccune:
@ghoneycutt README updated, review comments have been addressed.  Should be ready for merge / release now.

ghoneycutt:
Released in v2.16.0

PR #502, opened 2016-05-05
sensu-enterprise (if in manifest) also needs a notify for config changes

dtweet:

PR #720, opened 2017-07-07
(#495) Notify Service[sensu-enterprise] from Sensu::Check resources

### LINKED ISSUE #495: $check_notify does not load sensu::enterprise::service

jeffmccune:
Without this patch The sensu-enterprise service is not notified
automatically when `sensu::check` resources are declared.  This patch
addresses the problem by checking the catalog for the service and
automatically adding the notify relationship.

resolves #502 
resolves #495
closes #653

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -21,11 +21,13 @@ Puppet::Type.newtype(:sensu_check) do
   def initialize(*args)
     super *args
 
-    self[:notify] = [
-      "Service[sensu-client]",
-      "Service[sensu-server]",
-      "Service[sensu-enterprise]",
-    ].select { |ref| catalog.resource(ref) }
+    if c = catalog
+      self[:notify] = [
+        "Service[sensu-client]",
+        "Service[sensu-server]",
+        "Service[sensu-enterprise]",
+      ].select { |ref| c.resource(ref) }
+    end
   end
 
   ensurable do
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -24,6 +24,7 @@ Puppet::Type.newtype(:sensu_check) do
     self[:notify] = [
       "Service[sensu-client]",
       "Service[sensu-server]",
+      "Service[sensu-enterprise]",
     ].select { |ref| catalog.resource(ref) }
   end
 
--- a/tests/sensu-server-enterprise.pp
+++ b/tests/sensu-server-enterprise.pp
@@ -9,18 +9,19 @@ node 'sensu-server' {
     mode   => '0600',
   }
 
+  # NOTE: When testing sensu enterprise, provide the SE_USER and SE_PASS to use
+  # with the online repository using the FACTER_SE_USER and FACTER_SE_PASS
+  # environment variables.  An effective way to manage this is with `direnv`
   class { '::sensu':
     install_repo              => true,
-    server                    => true,
+    enterprise                => true,
+    enterprise_user           => $facts['se_user'],
+    enterprise_pass           => $facts['se_pass'],
     manage_services           => true,
     manage_user               => true,
     rabbitmq_password         => 'correct-horse-battery-staple',
     rabbitmq_vhost            => '/sensu',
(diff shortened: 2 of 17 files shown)


lib/puppet/type/sensu_check.rb
```diff
@@ -21,10 +21,13 @@ def should
...
 
-    self[:notify] = [
-      "Service[sensu-client]",
-      "Service[sensu-server]",
-    ].select { |ref| catalog.resource(ref) }
+    if c = catalog
+      self[:notify] = [
+        "Service[sensu-client]",
+        "Service[sensu-server]",
+        "Service[sensu-enterprise]",
```
ghoneycutt: if Service[sensu-enterprise] is not in the catalog, because we aren't using enterprise, does this cause a problem?

jeffmccune: I expect no problem because`Array#select` filter out the resource ref when absent from the catalog.

ghoneycutt: Have you verified with `vagrant up` ?

jeffmccune: No, will do so now.

jeffmccune: @ghoneycutt I modified the provisioning manifest for `sensu-server-enterprise` to manage Service[sensu-enterprise] instead of Service[sensu-server].  Take a look at this change in 1d694e4, I assumed `tests/sensu-server-enterprise.pp` is intended to setup sensu-enterprise.

With `vagrant up sensu-server`, the notification doesn't cause a problem as we expect.  With `vagrant up sensu-server-enterprise`, the notification does properly notify `Service[sensu-enterprise]' when a managed `sensu::check` resource changes state.

ghoneycutt:
@jeffmccune see PR #502 where they made similar changes, but to other files as well.

jeffmccune:
@ghoneycutt Got it, I'll update this PR to add the notify to all of the files referenced in #502

jeffmccune:
@ghoneycutt Notifications and spec coverage added to all types listed in #502  This is ready for another review.

tests/sensu-server-enterprise.pp
```diff
@@ -9,18 +9,19 @@
...
     install_repo              => true,
-    server                    => true,
+    enterprise                => true,
+    enterprise_user           => $facts['se_user'],
+    enterprise_pass           => $facts['se_pass'],
     manage_services           => true,
     manage_user               => true,
     rabbitmq_password         => 'correct-horse-battery-staple',
     rabbitmq_vhost            => '/sensu',
-    api                       => true,
```
ghoneycutt: Why was the API portion removed?

jeffmccune:
The api portion was removed because the API is mutually exclusive with setting sensu::enterprise => true.  See init.pp L573:

```puppet
  if ( $enterprise and $api ) or ( $enterprise and $server ) {
    fail('Sensu Enterprise: sensu-enterprise replaces sensu-server and sensu-api')
  }
```

ghoneycutt:
Released in v2.19.1

PR #664, opened 2017-06-15
Add support for Chocolaty package manager on Windows

jpuskar:
Windows package support for non-default providers and titles.

PR #723, opened 2017-07-08
(#589) Add Chocolatey support for Windows

### LINKED ISSUE #589: REQ - Windows - Support chocolatey as a package manager
I'm looking to deploy on windows, into a pretty locked down environment - I've noticed that the package.pp gets the client MSI from the internet with no override.

I'd propose using chocolatey as a method of retrieving this for windows, however a package would be needed to be created for the MSI. Ideally, sensu would host a nuget feed with packaged versions of their MSIs, then we could rely on that or a local package feed if required.

I imagine there aren't many windows users of sensu, especially with more locked down environments, but I just thought I'd make my thoughts known. 

I'll take a look at the work required.

jeffmccune:
Without this patch package resource used to install sensu on Windows defaults
to the `msi` provider and can't easily be changed to the Chocolatey provider.
This patch addresses the problem by adding a new class parameter,
`sensu::windows_package_provider` which defaults to `undef`, the default
behavior of Puppet on windows, and can be set to `'chocolatey'`.

The chocolatey package provider may be exercised with `vagrant up
win2012r2-client-chocolatey`.  There is no `sensu` package in the public
chocolatey repository, however, so installation will fail with the following error:

    ==> win2012r2-client-chocolatey: Error: Execution of 'C:\ProgramData\chocolatey\bin\choco.exe install sensu -y ' returned 1: Ch
    ocolatey v0.10.7
    ==> win2012r2-client-chocolatey: Installing the following packages:
    ==> win2012r2-client-chocolatey: sensu
    ==> win2012r2-client-chocolatey: By installing you accept licenses for the packages.
    ==> win2012r2-client-chocolatey: sensu not installed. The package was not found with the source(s) listed.
    ==> win2012r2-client-chocolatey:  If you specified a particular version and are receiving this message, it is possible that the
     package name exists but the version does not.
    ==> win2012r2-client-chocolatey:  Version: ""
    ==> win2012r2-client-chocolatey:  Source(s): "https://chocolatey.org/api/v2/"
    ==> win2012r2-client-chocolatey: Chocolatey installed 0/1 packages. 1 packages failed.
    ==> win2012r2-client-chocolatey:  See the log for details (C:\ProgramData\chocolatey\logs\chocolatey.log).

Even with this error, the behavior of the chocolatey package provider is
exercised.

Closes #664
Resolves #589

Original Author: john.puskar <jpuskar@us.ibm.com>

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -83,7 +83,33 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     client.vm.provision :shell, :path => "tests/provision_basic_win.ps1"
     ## Symlink module into place, run puppet module install for puppet apply
     client.vm.provision :shell, :path => "tests/provision_basic_win.2.ps1"
+    ## Install Sensu using the default Windows package provider (MSI)
     client.vm.provision :shell, :inline => 'iex "puppet apply -v C:/vagrant/tests/sensu-client-windows.pp"'
+  end
+
+  config.vm.define "win2012r2-client-chocolatey", autostart: false do |client|
+    client.vm.box = "opentable/win-2012r2-standard-amd64-nocm"
+    client.vm.provider :virtualbox do |vb|
+      vb.customize ["modifyvm", :id, "--memory", "2048"]
+      vb.customize ["modifyvm", :id, "--cpus", "1"]
+    end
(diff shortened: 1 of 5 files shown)


ghoneycutt:
Released in v2.18.0

Issue #701, opened 2017-06-30
Stop using scope.lookupvar() in templates


ghoneycutt:
We should bring the variables into the local scope and use `@variable`

https://docs.puppet.com/puppet/4.10/lang_template_erb.html#accessing-puppet-variables

Phil-Friderici:
If we move the variables to the classes where they are really used, we will break backward compatibility.
For example the template sensu.erb is used in sensu::package but the variables are sourced in from the main class.

To maintain backward compatibility in the first step, I could add the variables to sensu::package and source them in from the main class. This way you get the same results using the old way.

``` puppet
class sensu::package (
  $log_dir = $::sensu::log_dir,
  ) {
```

ghoneycutt:
Let's go the second route and do variable assignment that references variables in the main class. Eventually I'd like to collapse all these extra classes and this will make it easier to understand where the variables come from.

Phil-Friderici:
Working on this in PR https://github.com/sensu/sensu-puppet/pull/724

Phil-Friderici:
this issue can be closed now :)

ghoneycutt:
Released in v2.22.0

PR #724, opened 2017-07-09
(GH-701) Stop using scope.lookupvar() in templates

Phil-Friderici:
Add tests for existing functionality as preparation and to proof backward compatibility.


Phil-Friderici:
I have separated the commits to proof that the existing functionality is not changed. If prefered I can squash them into one commit.

Phil-Friderici:
ready for review

ghoneycutt:
Please squash into one commit. Thanks!

jeffmccune:
This looks good to me.  The tests are great!  Only question is why scope.lookup var is being left in `sensu-client.erb` and `sensu-windows-package-version.erb`?

Phil-Friderici:
Thanks for the positive feedback, will do the same for the other templates as well :)

Phil-Friderici:
Here we go:
- all appearances of scope.lookupcvar() have been removed
- all commits squashed into one

manifests/package.pp
```diff
@@ -49,7 +108,7 @@
       # $pkg_version is passed to Package[sensu] { ensure }.  The Windows MSI
       # provider translates hyphens to dots, e.g. '0.29.0-11' maps to
       # '0.29.0.11' on the system.  This mapping is necessary to converge.
-      $pkg_version = template('sensu/sensu-windows-package-version.erb')
+      $pkg_version = regsubst($::sensu::version, '-', '.')
```
ghoneycutt: good catch!

ghoneycutt:
@Phil-Friderici looks like we still have a few

```
sensu-puppet$ grep -ri scope.lookupvar .
./manifests/extension.pp:  $filename = inline_template('<%= scope.lookupvar(\'source\').split(\'/\').last %>')
./manifests/handler.pp:    $filename = inline_template('<%= scope.lookupvar(\'source\').split(\'/\').last %>')
./manifests/mutator.pp:    $filename = inline_template('<%= scope.lookupvar(\'source\').split(\'/\').last %>')
./manifests/plugin.pp:      $filename = inline_template('<%= scope.lookupvar(\'name\').split(\'/\').last %>')
./manifests/plugin.pp:      $filename = inline_template('<%= scope.lookupvar(\'name\').split(\'/\').last %>')
```

While you are changing these, could you have them use .erb files. I prefer to use .erb files, even for one liners, so they can be validated.

manifests/package.pp
Phil-Friderici: same is true for sensu-version.erb. I have created the issue https://github.com/sensu/sensu-puppet/issues/731 to get it changed later on too.

Phil-Friderici:
have added the grep results to https://github.com/sensu/sensu-puppet/issues/731 as I think they should be refactored to Puppet code instead.

ghoneycutt:
Yes, please address that in this PR.

Solves #731

ghoneycutt:
@Phil-Friderici please use vagrant to confirm that all platforms still work. Here's a bash function I use to see what clients are connected, instead of poking around through the Dashboard.

```bash
function sensuclients {
}
```

Phil-Friderici:
Think we should replace Ruby with Puppet code where possible. So I have replaced all occurrences of
```inline_template('<%= scope.lookupvar(\'source\').split(\'/\').last %>')``` with the much simplier to read ```basename()``` function.

Sorry, haven't used vagrant yet. Will need some time to get into the details and understand it.

File: manifests/package.pp [hunk 2]
```diff
@@ -69,7 +128,7 @@
         # The OS Release specific sub-folder
         $os_release = $facts['os']['release']['major']
         # e.g. '2012 R2' => '2012r2'
-        $pkg_url_dir = template('sensu/sensu-version.erb')
+        $pkg_url_dir = regsubst($os_release, '^(\d+)\s*[rR](\d+)', '\\1r\\2')
```
Phil-Friderici: same as for line 111 above

Phil-Friderici:
sweet :)

ghoneycutt:
Released in v2.22.0

PR #726, opened 2017-07-09
WIP - Freebsd

ghoneycutt:


ghoneycutt:
handler definitions, which are JSON configuration data, live under /etc/sensu/conf.d/handlers

/etc/sensu/handlers is one place to put handler plugins, the actual executable

madelaney:
I'd be more then happy to help, if you guys need it.

ghoneycutt:
@madelaney We would love to have your input! Basing your changes on this PR would be the way to start. You should be able to do `vagrant up sensu-server` then `vagrant up freebsd10-client` and see that the freebsd system has registered. I use this bash function from my workstation to help with that.

```bash
function sensuclients {
}
```

I started with FreeBSD10 as 11 had a bunch of problems with Vagrant. Would suggest we get 10 to work first, then add support for 11 in a different PR.

madelaney:
@ghoneycutt No problem. I'm not 100% sure how I'd "merge" any of my changes with this PR though. Would it be more advantageous to create a freebsd branch where we can work off of?

jeffmccune:
> @ghoneycutt No problem. I'm not 100% sure how I'd "merge" any of my changes with this PR though. Would it be more advantageous to create a freebsd branch where we can work off of?

There are quite a few different ways to do it, I suggest taking your work and rebase it on top of Garrett's.  This is usually how I do it:

Configure Git to pull down PR's: https://gist.github.com/piscisaureus/3342247

Assuming you're on your own branch named "my_topic" checkout a new branch we'll use to rebase on Garrett's work:

    git checkout -b my_topic_rebase

Fetch the pull requests, you should see something like `* [new ref]         refs/pull/726/head -> upstream/pr/726`

    git fetch --all

Now you have the HEAD of Garrett's PR, go ahead and [interactively rebase] your branch on top of Garrett's.

    git rebase -i upstream/pr/726

As your commits are replayed resolve any merge conflicts and `git rebase --continue` until the rebase completes.  After doing this once it will be reasonably straight forward to keep rebasing to stay current.

Hope this helps, Garrett may have a different approach which will be better since you'll be working with him.

madelaney:
@ghoneycutt May I ask why the service type is set to `bsd` and not the default for FreeBSD/DragonFly?

I have a working example, off an older version, that uses augeas. See https://github.com/madelaney/sensu-puppet/blob/master/manifests/client/service.pp#L80 for the full example.

PR #728, opened 2017-07-11
(#597) Add sensu::contact type (Enterprise Only)

### LINKED ISSUE #597: Unable to define Contact Routing for Sensu Enterprise
## Description of problem

Per [the Sensu Enterprise documentation](https://sensuapp.org/docs/0.26/enterprise/contact-routing.html), contacts need to be defined within the `"contacts": {}` configuration scope. An example of the way the contacts config should be defined is below: 

```
{
  "contacts": {
    "support": {
      "email": {
        "to": "support@sensuapp.com"
      },
      "slack": {
        "channel": "#support"
      }
    }
  }
}
```

Using this puppet module, I don't see a way to define a contacts hash at the top level configuration scope. I tried tricking the module and used `sensu::config` to specify the contact hash, but I was unable to get this to work, when specifying the hash as the event or the config. Even if this did work, it was not an ideal solution, as I would need to masquerade the config as additional check configuration, and not as a contacts file.

I was finally able to find a solution by adding a handler using `sensu::handler` named contacts, and providing the contacts as a hash to the handler config. I needed to provide a dummy command to this handler, but since I am not going to be using it, I guess it's not a big deal. This added the `"contacts": {}` configuration scope and seems to be working, although it feels a bit hacky.

Is there any way to specify arbitrary configuration at the top level? Is there any need to add a contacts resource? How are others implementing the contact configuration with this puppet module?

- Your OS: CentOS 6
- Your Ruby version: 2.3.0
- Your version of Puppet: 4.8.1
- Your version of Sensu: 0.26.5
- Your version of this module: 2.2.0

jeffmccune:
Without this patch the Puppet module doesn't provide an affordance for [contact
routing](https://sensuapp.org/docs/0.29/enterprise/contact-routing.html)  This
patch addresses the problem by adding a new defined type, sensu::contact backed
by a custom type and provider for sensu_contact.

Example usage:

```puppet
sensu::contact { 'support':
  config => { 'email'  => { 'to' => 'support@example.com' } },
}
sensu::contact { 'ops':
  config => { 'email'  => { 'to' => 'ops@example.com' } },
}
sensu::check { 'check_ntp':
  command  => 'PATH=$PATH:/usr/lib64/nagios/plugins check_ntp_time -H pool.ntp.org -w 30 -c 60',
  handlers => 'email',
  contacts => ['ops', 'support'],
}
```

Resolves #597

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -69,6 +69,17 @@ Puppet::Type.newtype(:sensu_check) do
     end
   end
 
+  newproperty(:contacts, :array_matching => :all, :parent => SensuCheckArrayProperty) do
+    desc "Contact names to override handler configuration via Contact Routing"
+    # Valid names documented at
+    # https://sensuapp.org/docs/0.29/enterprise/contact-routing.html#contact-names
+    newvalues(/^[\w\.-]+$/, :absent)
+    def insync?(is)
+      return is.sort == should.sort if is.is_a?(Array) && should.is_a?(Array)
+      is == should
+    end
+  end
+
   newproperty(:high_flap_threshold) do
     desc "A host is determined to be flapping when the percent change exceedes this threshold."
     newvalues(/.*/, :absent)
--- a/manifests/contact.pp
+++ b/manifests/contact.pp
@@ -1,0 +1,55 @@
+# = Define: sensu::routing
+#
+# Manage [Contact
+# Routing](https://sensuapp.org/docs/latest/enterprise/contact-routing.html)
+# configuration with Sensu Enterprise.
+#
+# Note:  If the `sensu::purge_config` class parameter is `true`, unmanaged
+# sensu::contact resources located in /etc/sensu/conf.d/contacts will be purged.
+#
+# == Parameters
+#
+# [*ensure*]
+#   String. Whether the check should be present or not
+#   Default: present
+#   Valid values: present, absent
+#
+# [*base_path*]
+#   String.  Where to place the contact JSON configuration file.  Defaults to
+#   `undef` which defers to the behavior of the underlying sensu_contact type.
+#   Default: undef
+#
+# [*config*]
+#   Hash. The configuration data for the contact.  This is an arbitrary hash to
+#   accommodate the various communication channels. For example, `{ "email": {
+#   "to": "support@example.com" } }`.
+#   Default: {}
+define sensu::contact(
+  $ensure = 'present',
+  $base_path = undef,
+  $config = {},
+) {
+  validate_re($ensure, ['^present$', '^absent$'] )
+  validate_hash($config)
+
+  $file_ensure = $ensure ? {
+    'absent' => 'absent',
+    default  => 'file'
+  }
+
+  # handler configuration may contain "secrets"
+  file { "/etc/sensu/conf.d/contacts/${name}.json":
+    ensure => $file_ensure,
+    owner  => 'sensu',
+    group  => 'sensu',
+    mode   => '0440',
+    before => Sensu_contact[$name],
+  }
+
+  sensu_contact { $name:
+    ensure    => $ensure,
+    config    => $config,
+    base_path => $base_path,
+    require   => File['/etc/sensu/conf.d/contacts'],
+  }
+}
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -144,7 +144,7 @@ class sensu::package {
     }
   }
 
-  file { [ $::sensu::conf_dir, "${sensu::conf_dir}/handlers", "${sensu::conf_dir}/checks", "${sensu::conf_dir}/filters", "${sensu::conf_dir}/extensions", "${sensu::conf_dir}/mutators" ]:
+  file { [ $::sensu::conf_dir, "${sensu::conf_dir}/handlers", "${sensu::conf_dir}/checks", "${sensu::conf_dir}/filters", "${sensu::conf_dir}/extensions", "${sensu::conf_dir}/mutators", "${sensu::conf_dir}/contacts" ]:
     ensure  => directory,
     owner   => $::sensu::user,
     group   => $::sensu::group,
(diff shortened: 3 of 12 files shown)


jeffmccune:
Tested in Vagrant using sensu-server-enterprise

lib/puppet/type/sensu_check.rb
```diff
@@ -69,6 +69,15 @@ def insync?(is)
     end
   end
 
+  newproperty(:contacts, :array_matching => :all, :parent => SensuCheckArrayProperty) do
+    desc "Contact names to override handler configuration via Contact Routing"
+    newvalues(/.*/, :absent)
```
ghoneycutt: regex for valid contact name - https://sensuapp.org/docs/0.29/enterprise/contact-routing.html#contact-names

manifests/contact.pp
```diff
@@ -0,0 +1,52 @@
...
+  validate_re($ensure, ['^present$', '^absent$'] )
+  validate_hash($config)
+
+  $file_ensure = $ensure ? {
+    'absent' => 'absent',
+    default  => 'file'
+  }
+
+  # handler configuration may contain "secrets"
+  file { "/etc/sensu/conf.d/contacts/${name}.json":
```
ghoneycutt: do we need a resource to manage `/etc/sensu/conf.d/contacts` ?

manifests/contact.pp
```diff
@@ -0,0 +1,52 @@
...
+    ensure => $file_ensure,
+    owner  => 'sensu',
+    group  => 'sensu',
+    mode   => '0440',
+    before => Sensu_contact[$name],
+  }
+
+  sensu_contact { $name:
```
ghoneycutt: instead of a type/provider, what about using sensu_sorted_json() as is done in sensu::write_json? The current layout is to used defined types that then call custom type/providers. What about still calling defined types, but then replacing the type/provider code with sensu_sorted_json()? If possible, I think that would greatly simplify things.

tests/sensu-server-enterprise.pp
```diff
@@ -36,4 +36,33 @@
...
+      'slack' => {
+        'channel' => '#support',
+      },
+    },
+  }
+  sensu::contact { 'ops':
+    ensure => 'present',
+    config => { 'email'  => { 'to' => 'ops@example.com' } },
+  }
+  sensu::contact { 'hacker':
```
ghoneycutt: suggest using another name.

manifests/contact.pp
ghoneycutt: Could we still use `resources {}` to purge?

jeffmccune:
Changed to “departed"

manifests/contact.pp
jeffmccune: I'll take a look.  This also reminds me, `self.instances` needs to be implemented in the type if we don't implement this using sensu_sorted_json().

tests/sensu-server-enterprise.pp
jeffmccune: Changed to “departed"

manifests/package.pp
```diff
@@ -144,7 +144,7 @@
     }
   }
 
-  file { [ $::sensu::conf_dir, "${sensu::conf_dir}/handlers", "${sensu::conf_dir}/checks", "${sensu::conf_dir}/filters", "${sensu::conf_dir}/extensions", "${sensu::conf_dir}/mutators" ]:
+  file { [ $::sensu::conf_dir, "${sensu::conf_dir}/handlers", "${sensu::conf_dir}/checks", "${sensu::conf_dir}/filters", "${sensu::conf_dir}/extensions", "${sensu::conf_dir}/mutators", "${sensu::conf_dir}/contacts" ]:
```
ghoneycutt: purge is there

manifests/contact.pp
```diff
@@ -0,0 +1,52 @@
...
+    ensure    => $ensure,
+    config    => $config,
+    base_path => $base_path,
+    require   => File['/etc/sensu/conf.d/contacts'],
```
ghoneycutt: Should this notify the service?

ghoneycutt: i see it is in the type.

ghoneycutt:
Please add the example usage to the README. Besides that, else I think it's ready.

lib/puppet/type/sensu_check.rb
jeffmccune: Regex has been added both to sensu_check's `contacts` property and sensu_contact's `name` parameter.  Spec tests have been updated to exercise the expected behavior of invalid names throwing Puppet::ResourceError's.

manifests/contact.pp
jeffmccune: Filed #729 to track this independently of this PR and related enhancement.

jeffmccune: Added this documentation to the `sensu::contact` defined type:

> Note:  If the `sensu::purge_config` class parameter is `true`, unmanaged `sensu::contact` resources located in /etc/sensu/conf.d/contacts will be purged.

manifests/package.pp
jeffmccune: Got it, added documentation to sensu::contact to note this purge behavior.

manifests/contact.pp
jeffmccune: Correct

jeffmccune:
> Please add the example usage to the README. Besides that, else I think it's ready.

README updated with a similar example to the vagrant manifest in `tests/sensu-server-enterprise.pp`.

manifests/contact.pp
```diff
@@ -0,0 +1,55 @@
+# = Define: sensu::routing
+#
+# Manage [Contact
+# Routing](https://sensuapp.org/docs/0.26/enterprise/contact-routing.html)
```
ghoneycutt: we're on 0.29. Best to use latest here though

`https://sensuapp.org/docs/latest/enterprise/contact-routing.html`

jeffmccune: Updated to latest.

ghoneycutt:
Released in v2.20.0

PR #730, opened 2017-07-11
(#433) Reload Service[sensu_api] on RabbitMQ config changes

### LINKED ISSUE #433: sensu-api service should subscribe to sensu::rabbitmq::config class
It appears that currently the `sensu-api` service [1] doesn't subscribe to the `sensu::rabbitmq::config` class, which means that it doesn't get restarted automatically if the rabbitmq config is changed. 

From my testing I believe it should do, otherwise the API doesn't work. 

[1] https://github.com/sensu/sensu-puppet/blob/master/manifests/api/service.pp#L34-L39

jeffmccune:
Without this patch the Sensu API service doesn't reload itself when RabbitMQ
configuration changes occur.  This patch addresses the problem by adding a
relationship between the rabbitmq configuration class and the service resource.

Note, both sensu::rabbitmq::config and sensu::redis::config are in the catalog
regardless of which is actually being configured.

Resolves #433

### DIFF
--- a/manifests/api/service.pp
+++ b/manifests/api/service.pp
@@ -40,6 +40,7 @@ class sensu::api::service (
           Class['sensu::package'],
           Class['sensu::api::config'],
           Class['sensu::redis::config'],
+          Class['sensu::rabbitmq::config'],
         ],
       }
     }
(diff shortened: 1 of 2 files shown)


ghoneycutt:
Released in v2.19.2

PR #733, opened 2017-07-11
(#562) Sensu_filter resources notify Sensu Server and Sensu Enterprise

jeffmccune:
Without this patch, Service[sensu-server] and Service[sensu-enterprise] do not
reload when Sensu_filter resources change.

This patch addresses the problem by adding a notify relationship if the service
resources exist in the catalog.

Resolves #562


ghoneycutt:
Released in v2.20.1

PR #738, opened 2017-07-12
(#735) Fix sensu-server-enterprise Vagrant VM

### LINKED ISSUE #735: Vagrant ports for sensu-enterprise-server are off
Currently one cannot test the API when using sensu-enterprise-server from the host. I believe that the redirected ports should be different in the Vagrantfile.

jeffmccune:
Without this patch the sensu-enterprise server is quickly crashing with a JVM
out of memory error when started inside the sensu-server-enterprise VM.

This patch addresses the problem by bumping up the allocated RAM for the VM
from 512m to 768m while also bringing down the JVM heap size from 2GB to 256m.

The API has been tested with the following commands, with no other VM's
running:

```bash
vagrant up sensu-server-enterprise
curl -s http://admin:secret@127.0.0.1:4567/clients \
  | jq -r 'map(.name | split(".") | .[0]) | sort | .[]'
```

Resolves #735

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,23 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
-
+#
+# Environment variables may be used to control the behavior of the Vagrant VM's
+# defined in this file.  This is intended as a special-purpose affordance and
+# should not be necessary in normal situations.  In particular, sensu-server,
+# sensu-server-enterprise, and sensu-server-puppet5 use the same IP address by
+# default, creating a potential IP conflict.  If there is a need to run multiple
+# server instances simultaneously, avoid the IP conflict by setting the
+# ALTERNATE_IP environment variable:
+#
+#     ALTERNATE_IP=192.168.56.9 vagrant up sensu-server-enterprise
+#
+# NOTE: The client VM instances assume the server VM is accessible on the
+# default IP address, therefore using an ALTERNATE_IP is not expected to behave
+# well with client instances.
+#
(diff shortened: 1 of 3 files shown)


tests/sensu-server-enterprise.pp
```diff
@@ -1,5 +1,14 @@
 node 'sensu-server' {
 
+  Ini_setting {
```
ghoneycutt: please add that curl command to test in the comments here.

Vagrantfile
```diff
@@ -25,7 +25,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.define "sensu-server", primary: true, autostart: true do |server|
     server.vm.box = "centos/7"
     server.vm.hostname = 'sensu-server.example.com'
-    server.vm.network :private_network, ip: "192.168.56.10"
+    server.vm.network :private_network, ip: ENV['VAGRANT_SENSU_SERVER_IP'] || '192.168.56.10'
```
ghoneycutt: Could you please document this in the Vagrantfile

tests/sensu-server-enterprise.pp
jeffmccune: Added!

Vagrantfile
jeffmccune: Sure, thing, documentation added to the top of the Vagrantfile.  I also renamed the variables and consolidated to `ENV['ALTERNATE_IP']` for clarity.

ghoneycutt:
Released in v2.22.0

Issue #578, opened 2016-11-04
etc_dir should be configurable


jaredledvina:
## Description of problem
We're customizing the configuration directory of sensu and want to be able to set the etc_dir variable to a custom value. Looks like right now it's hard coded based on the OS type. 

This is the section I'm looking at: https://github.com/sensu/sensu-puppet/blob/master/manifests/init.pp#L525-L549

## Platform and version information

- Your OS: CentOS 6
- Your Ruby version: 
- Your version of Puppet: Open Source Puppet 3.8
- Your version of Sensu: 0.26.5
- Your version of this module: Latest

jaxxstorm:
Agreed, a PR would be appreciated!

Phil-Friderici:
This is the section jaredledvina was looking at the 4th November 2016:
https://github.com/sensu/sensu-puppet/blob/93b2c2bdddb4ab7b221f11002f85ddd9bce04675/manifests/init.pp#L525-L549

Phil-Friderici:
working on this in PR https://github.com/sensu/sensu-puppet/pull/741

ghoneycutt:
Thanks for all your help!

Released in v2.28.0

PR #741, opened 2017-07-13
(GH-578) etc_dir should be configurable

Phil-Friderici:


spec/spec_helper.rb
```diff
@@ -26,5 +26,6 @@
     :ipaddress   => '127.0.0.1',
     :kernel      => 'Linux',
     :osfamily    => 'RedHat',
+    :fqdn        => 'testfqdn',
```
ghoneycutt: your fqdn isnt fully qualified, perhaps `testfqdn.example.com`

Phil-Friderici:
Wow, this small parameter does touch a lot of resources.

The sensu::check class does not source in $etc_dir from sensu main class. It does use it's own (same) default values. Should I change senus::check to also source values from main class ?

spec/spec_helper.rb
Phil-Friderici: fixed

Phil-Friderici:
will rebase

Phil-Friderici:
rebased & ready for review

manifests/init.pp
```diff
@@ -22,6 +22,12 @@
 #   Default: undef
 #   Valid values: sensu_gem, apt, aptitude, yum
 #
+# [*sensu_etc_dir*]
+#   String. Absolute path to the sensu etc directory.
+#   Default: '/etc/sensu' on RedHat and Debian
+#            'C:/opt/sensu' on windows
+#            undef on all other platforms
```
ghoneycutt: what happens if this is undef? Maybe the default should be `/etc/sensu` ?

Phil-Friderici: That would make sense in most cases I guess.

Phil-Friderici:
@tuxmea Let's add support for new OSfamilies in specific PRs/commits. Guess there is more work to do than just adding different paths for $sensu_etc_dir.

manifests/init.pp
Phil-Friderici: changed the behaviour to default to '/etc/sensu'

spec/classes/sensu_init_spec.rb
```diff
@@ -30,6 +136,106 @@
       }
     end
 
+    # FIXME: The following resource checks are only testing $sensu_etc_dir specific values
```
ghoneycutt: what else should they test?

Phil-Friderici: These tests only checks for the $sensu_etc_dir specific details. They don't test the other file resource details that should be tested as well. Eg:

``` ruby
      it do
        should contain_file('C:/opt/sensu/bin/sensu-client.xml').with({
          'ensure'  => 'file',
          'content' => content,
          ... and other parameters....
        })
      end
```
There is still a lot of test to add for all the managed resources and functionality. But this is out of scope for this PR.

Phil-Friderici:
will rebase

Phil-Friderici:
rebased

ghoneycutt:
Sorry to ask you to rebase again, though we've implemented data types and all PR's will need to be rebased.

Phil-Friderici:
et voilà, rebased for a good thing ;)

ghoneycutt:
Solves #578

ghoneycutt:
Thank you! Released in v2.28.0

PR #743, opened 2017-07-13
(#737) Add cron attribute to sensu::check type

### LINKED ISSUE #737: sensu::check resources should support cron scheduling
## Description of problem

In Sensu 0.28 the `cron` attribute was added to the [check definition specification](https://sensuapp.org/docs/latest/reference/checks#check-attributes) as an alternative to `interval` scheduling. 

When present, the value of this attribute supersedes the `interval` attribute.

When `cron` is configured with a value, `interval` should not be required nor present in the configuration written to disk.

jeffmccune:
Without this patch the sensu::check type does not manage the
[cron](https://sensuapp.org/docs/latest/reference/checks#check-attributes)
attribute.  This patch adds the `cron` parameter.  If specified, the interval
parameter will be set to `'absent'` which will remove the interval attribute
from the JSON check configuration if present.

Resolves #737

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -89,6 +89,11 @@ Puppet::Type.newtype(:sensu_check) do
     end
   end
 
+  newproperty(:cron) do
+    desc 'When the check should be executed, using the Cron syntax.'
+    newvalues(/.*/, :absent)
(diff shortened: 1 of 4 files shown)


ghoneycutt:
Released in v2.23.0

PR #744, opened 2017-07-13
(#727) Add sensu::spawn_limit class parameter

### LINKED ISSUE #727: support for setting spawn limit via puppet
we have hit the bottleneck of too many checks (JIT) from a single server leading to issues addressed by https://github.com/sensu/sensu/pull/1265. i know there are other workarounds to deal with this, but since this is public(?) config, i guess it should be configurable via  puppet.

but it seems this is not yet configurable via puppet.
any plans to include this configurable option ?

jeffmccune:
Without this patch the Sensu spawn limit setting cannot easily be configured by
this module.  This patch addresses the problem by adding a sensu::spawn_limit
class parameter.  If unset, the parameter defaults to the behavior of Sensu
which is to have a limit of 12 in version 0.29.0.

Resolves #727

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -71,6 +71,15 @@
 #   with systems that use apt.
 #   Default: $::lsbdistcodename
 #
+# [*spawn_limit*]
+#   Integer.  Tune concurrency of the sensu-server pipe handler and the
+#   sensu-client check execution.  This setting should not need to be tuned
+#   except in specific situations, e.g. when there are a large number of JIT
+#   clients.  See [#727](https://github.com/sensu/sensu-puppet/issues/727) for
+#   more information.  The default is undefined, which does not manage
+#   `/etc/sensu/conf.d/spawn.json`
+#   Default: undef
+#
 # [*client*]
 #   Boolean.  Include the sensu client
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -71,6 +71,15 @@
...
 #   Default: $::lsbdistcodename
 #
+# [*spawn_limit*]
+#   Integer.  Tune concurrency of the sensu-server pipe handler and the
+#   sensu-client check execution.  This setting should not need to be tuned
+#   except in specific situations, e.g. when there are a large number of JIT
+#   clients.  See [#727](https://github.com/sensu/sensu-puppet/issues/727) for
+#   more information.  The default is undefined, which does not manage
+#   `/etc/sensu/conf.d/spawn.json`
+#   Default: undef
```
ghoneycutt: the commit reads a default of 12 though this is undef.

jeffmccune: Updated the commit message to read:

> If unset, the module behavior defers to the default behavior of Sensu which uses an effective spawn limit of 12 in version 0.29.0.

ghoneycutt:
Released in v2.24.0

Issue #582, opened 2016-11-11
to_type helper's handling of numbers is too loose


threadwaste:
## Description of problem

- What did you do?

Tried to pass a numerical identifier with leading zeroes (e.g. "012345") through client_custom.

- What happened?

The identifier was converted to an integer, lost its leading zeroes, and stored as a JSON number.

- What did you expect to happen?

The identifier would retain its leading zeroes, and be stored as a JSON string.

- How can someone reproduce the problem?

```puppet
class { 'sensu':
  client_custom => {
    some_value => "0123"
  }
}
```

## Anything else to add that you think will be helpful?

I'm trying to understand this helper better. The original commit suggests its tied to several type checks inside Sensu, which I'm guessing pose a problem when facts are used as parameter values. A quick `ag` of the source shows only a couple of these type checks present in HEAD for very specific values.

Similarly, the use of #to_type in this module is inconsistent. For example, the sensu_client_config type calls it for keepalive, but not safe_mode.

I have two branches of this local that deal with this. One uses more specific matchers for numeric types, the other simply removes the calling #to_type over value for the client_custom parameter. Happy to PR one, but figured I'd seek clarification and input on direction first.

ghoneycutt:
This is related to issue #682 to use Puppet v4's data types. We should enforce data types for certain values while respecting data types such as a string that looks like an integer, instead of transforming them.

ghoneycutt:
@cwjohnston Does Sensu have a need for integers to be cast as strings or to allow an integer with a leading zero?

alvagante:
So, current situation, after migration to puppet 4 parameters validation, is that when integers are passed as strings they are accepted and should not be converted.
@threadwaste can you verify if you have your issue with current master?

ghoneycutt:
@cwjohnston We need your guidance on what the right thing to do is regarding numbers.

@threadwaste why would you have a number that begins with zero?

cwjohnston:
@ghoneycutt Sensu implements what we call strong configuration validation, meaning Sensu will refuse to start if it determines configuration is invalid. This extends to a number of configuration attributes where integer values are required, e.g. port numbers, check intervals and keepalive or timeout thresholds. This strong validation does not extend to custom client or check attributes.

> Does Sensu have a need for integers to be cast as strings or to allow an integer with a leading zero?

I'm not aware of any case where an attribute requires both an integer value and a leading zero.

[...]

PR #746, opened 2017-07-14
682 data types

alvagante:
Checked syntax. Not spec tests.
Some choices are worth discussion.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -129,94 +129,44 @@
 #   matches the defined client attributes.  See the documentation for the format
 #   of the Hash value.
 #   Default: undef
-define sensu::check(
-  $command             = undef,
-  $ensure              = 'present',
-  $type                = undef,
-  $handlers            = undef,
-  $contacts            = undef,
-  $standalone          = true,
-  $cron                = 'absent',
-  $interval            = 60,
-  $occurrences         = undef,
-  $refresh             = undef,
-  $source              = undef,
-  $subscribers         = undef,
-  $low_flap_threshold  = undef,
-  $high_flap_threshold = undef,
-  $timeout             = undef,
-  $aggregate           = undef,
-  $aggregates          = undef,
-  $handle              = undef,
-  $publish             = undef,
-  $dependencies        = undef,
-  $custom              = undef,
-  $ttl                 = undef,
-  $subdue              = undef,
-  $proxy_requests      = undef,
+define sensu::check (
+  Optional[String]                      $command = undef,
+  Enum['present','absent']              $ensure = 'present',
+  Optional[String]                      $type = undef,
+  Variant[Undef,String,Array]           $handlers = undef,
+  Optional[Array]                       $contacts = undef,
+  Variant[Boolean,Enum['absent']]       $standalone = true,
+  String                                $cron = 'absent',
+  Variant[Integer,Enum['absent']]       $interval = 60,
+  Variant[Undef,Pattern[/^(\d+)$/],Integer,Enum['absent']] $occurrences = undef,
+  Variant[Undef,Enum['absent'],Integer] $refresh = undef,
+  Variant[Undef,String,Integer]         $source = undef,
+  Variant[Undef,String,Array]           $subscribers = undef,
+  Variant[Undef,Enum['absent'],Integer] $low_flap_threshold = undef,
+  Variant[Undef,Enum['absent'],Integer] $high_flap_threshold = undef,
+  Variant[Undef,Enum['absent'],Numeric] $timeout = undef,
+  Optional[String]                      $aggregate = undef,
+  Variant[Undef,String,Array]           $aggregates = undef,
+  Variant[Undef,Enum['absent'],Boolean] $handle = undef,
+  Variant[Undef,Enum['absent'],Boolean] $publish = undef,
+  Variant[Undef,String,Array]           $dependencies = undef,
+  Optional[Hash]                        $custom = undef,
+  Variant[Undef,Enum['absent'],Integer] $ttl = undef,
+  Variant[Undef,Enum['absent'],Hash]    $subdue = undef,
+  Variant[Undef,Enum['absent'],Hash]    $proxy_requests = undef,
 ) {
-
-  validate_re($ensure, ['^present$', '^absent$'] )
 
   if $ensure == 'present' and !$command {
     fail("sensu::check{${name}}: a command must be given when ensure is present")
   }
-  if !is_bool($standalone) and  $standalone != 'absent' {
-    fail("sensu::check{${name}}: standalone must be a boolean or 'absent' (got: ${standalone})")
-  }
-  if $handle and !is_bool($handle) and  $handle != 'absent' {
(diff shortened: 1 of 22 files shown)


spec/classes/sensu_init_spec.rb
```diff
@@ -402,7 +402,7 @@
           context "when #{var_name} (#{type}) is set to invalid #{invalid} (as #{invalid.class})" do
             let(:params) { [mandatory_params, var[:params], { :"#{var_name}" => invalid, }].reduce(:merge) }
             it 'should fail' do
-              expect { should contain_class(subject) }.to raise_error(Puppet::Error, /#{var[:message]}/)
+              expect { should contain_class(subject) }.to raise_error(Puppet::PreformattedError)
```
Phil-Friderici: This would remove the test for the specific error messages provided in the hash as 'message'.
Please don't change.

Phil-Friderici:
Good to see this :)

manifests/check.pp
```diff
@@ -129,89 +129,39 @@
...
-    fail("sensu::check{${name}}: subscribers must be an array or a string (got: ${subscribers})")
-  }
-  if $subdue {
-    if is_hash($subdue) {
-      if !( has_key($subdue, 'days') and is_hash($subdue['days']) ){
-        fail("sensu::check{${name}}: subdue hash should have a proper format. (got: ${subdue}) See https://sensuapp.org/docs/latest/reference/checks.html#subdue-attributes")
-      }
-    } elsif !($subdue == 'absent') {
-      fail("sensu::check{${name}}: subdue must be a hash or 'absent' (got: ${subdue})")
+  if $subdue =~ Hash {
```
ghoneycutt: do we need a struct for this?

manifests/init.pp
```diff
@@ -445,153 +445,130 @@
 #   Default: 'Sensu'
 
 class sensu (
-  $version                        = 'installed',
-  $sensu_plugin_name              = 'sensu-plugin',
-  $sensu_plugin_provider          = $::osfamily ? {
+  Pattern[/^absent$/, /^installed$/, /^latest$/, /^present$/, /^[\d\.\-el]+$/] $version = 'installed',
```
ghoneycutt: Not sure we should be so picky here. Perhaps `String` is right here. What do you think?

spec/classes/sensu_package_spec.rb
```diff
@@ -294,7 +294,7 @@
         let(:params) { { :purge => 'a_string' } }
 
         it 'should fail' do
-          expect { should create_class('sensu') }.to raise_error(/not a Hash/)
+          expect { should create_class('sensu') }.to raise_error(Puppet::PreformattedError)
```
ghoneycutt: I'm not familiar with `Puppet::PreformattedError` what is this?

manifests/check.pp
alvagante: Here yes, because value can also be the string 'absent' and this check has to be done only on hashes.

manifests/init.pp
alvagante: Some tests might fail, and this is probably the same validation which was there before

spec/classes/sensu_package_spec.rb
alvagante: The error we were having in failing spec tests. Related to type validation

manifests/check.pp
```diff
@@ -129,89 +129,39 @@
...
-  $proxy_requests      = undef,
+  String $command,
+  Enum['present','absent'] $ensure              = 'present',
+  Optional[String] $type                = undef,
+  Variant[Undef,String,Array] $handlers            = undef,
+  Optional[Array] $contacts            = undef,
+  Variant[Boolean,Enum['absent']] $standalone          = true,
+  String $cron                = 'absent',
+  Variant[Integer,Enum['absent']] $interval            = 60,
+  Variant[Undef,Pattern[/^(\d+)$/],Integer,Enum['absent']] $occurrences         = undef,
```
alvagante: RFC: In the module in most parts, numbers as accepted both as string and numbers. I've used this approach (Pattern[/^(\d+)$/],Integer ) in these cases, there could be something better maybe.

ghoneycutt: Asking for clarification in related ticket #582

spec/classes/sensu_init_spec.rb
alvagante: Oops, looks my my reply here got lost... 
so yes, I should update the message texts and check on them.
Wonder it it's actually really necessary as they are mostly normal Puppet type validation error messages, so not actually under the module code control.

manifests/check.pp
ghoneycutt: The module has many type/providers though we are looking to minimize the number through using sensu_sorted_json().

alvagante: Sorry, I was meaning data types, to write in the types dir

ghoneycutt: If you think that adding module specific data types will simplify things, then let's do it.

spec/classes/sensu_init_spec.rb
alvagante: @Phil-Friderici I couldn't do any better that this.
To match the actual error messages according to data type would need a refactoring of the test code.... maybe later ;-)

alvagante:
So @ghoneycutt from my pov the PR is basically ready.
Some choices have been done which need review:
- How to manage numbers.
Current approach accepts digits both passed as strings (but still digits) or Numeric.
This is needed to pass some of the existing tests, which would fail otherwise.
Current behaviour is somehow more permissive and this is better, IMHO, at is limits the risk to avoid breaking existing setups.
- On the same spirit, In various cases the exact accepted type of parameters was not clean (to me), generally I've tried to follow the less restrictive approach. (Any)
- Verification of the error message on some input validation checks have been simplified/generalized to avoid the need to refactor whole spec classes.  

Needless to say that whatever the tests may cover, with such an invasive commit we may break existing setups using some edge case not tested.

spec/classes/sensu_init_spec.rb
Phil-Friderici: @alvagante I still like to test that input parameter validation is as expected. Think this makes even sense with Puppet v4 and the new data type validations. It is not mandatory to have validation in Puppet v4 so it's still a module feature to take care for.

Testing for the actual error messages does proof that there is no other error popping up. With the general ```PreformattedError``` it would accept any error message regardless of the real reason.

spec/classes/sensu_package_spec.rb
Phil-Friderici: ```Puppet::PreformattedError``` will accept any error thrown, independently from the reason for the fail. The intention here is to ensure that the module really fails on the validation errors.

alvagante:
@ghoneycutt  I squashed this one in a single commit.
I'll leave the Sensitive part the the relevant PR, to merge once this one is merged.

Can eventually fix later some details like the ones based on pending decisions, or the ones pointed out by @Phil-Friderici  and @jeffmccune 

I'd prefer to merge and consolidate asap a common code base based on puppet 4 types.

ghoneycutt:
Released in v2.26.0

PR #751, opened 2017-07-15
re-defining these variables for sensu-enterprise

csoleimani:
/etc/default/sensu-enterprise lost these values because the module doesn't use scope.lookupvar() anymore. I tested this and it worked

### DIFF
--- a/manifests/enterprise/package.pp
+++ b/manifests/enterprise/package.pp
@@ -2,7 +2,17 @@
 #
 # Installs the Sensu packages
 #
-class sensu::enterprise::package {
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Thank you for reporting this. We are looking into it.

Phil-Friderici:
@csoleimani is right, I missed these occurencies. Thanks for the report and fix. I have added docs and test in PR https://github.com/sensu/sensu-puppet/pull/752 to ensure this is not missed at a later time.

ghoneycutt:
Thanks everyone!

Merged in #752

PR #753, opened 2017-07-17
Add handle_silenced parameter to handler defined type

madAndroid:
We have a handler that needs to operate on events, regardless of whether they've been silenced or not - this change implements the `handle_silenced` handler config option, as outlined in https://sensuapp.org/docs/0.29/reference/handlers.html

### DIFF
--- a/manifests/handler.pp
+++ b/manifests/handler.pp
@@ -67,6 +67,11 @@
 #   Default: false.
 #   Valid values: true, false
 #
+# [*handle_silenced*]
+#   Boolean.  If events in the silenced state should be handled.
+#   Default: false.
+#   Valid values: true, false
+#
 define sensu::handler(
   Enum['present','absent'] $ensure = 'present',
   Enum['pipe','tcp','udp','amqp','set','transport'] $type = 'pipe',
@@ -86,6 +91,7 @@ define sensu::handler(
   Any $subdue                      = undef,
   Optional[Integer] $timeout       = undef,
   Boolean $handle_flapping         = false,
+  Boolean $handle_silenced         = false,
 ) {
 
   if $subdue{ fail('Subdue at handler is deprecated since sensu 0.26. See https://sensuapp.org/docs/0.26/overview/changelog.html#core-v0-26-0')}
@@ -163,6 +169,7 @@ define sensu::handler(
     config          => $config,
     timeout         => $timeout,
     handle_flapping => $handle_flapping,
+    handle_silenced => $handle_silenced,
     notify          => $notify_services,
     require         => File['/etc/sensu/conf.d/handlers'],
   }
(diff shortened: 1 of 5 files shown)


spec/defines/sensu_handler_spec.rb
```diff
@@ -177,4 +177,9 @@
     it { should contain_sensu_handler('myhandler').with_handle_flapping( true ) }
   end
 
+  context 'handle_silenced' do
```
ghoneycutt: we also need a test for when this is false

manifests/handler.pp
```diff
@@ -67,6 +67,11 @@
 #   Default: false.
 #   Valid values: true, false
 #
+# [*handle_silenced*]
```
ghoneycutt: Please add an example to https://github.com/sensu/sensu-puppet/blob/master/tests/sensu-server.pp and document which file should be modified what json should be expected. That way we can functionally test this behaves as expected.

ghoneycutt:
@madAndroid Awesome work, thanks for adding this!

madAndroid:
@ghoneycutt - thanks :) Comments addressed and changes made

manifests/handler.pp
madAndroid: I've added that in d05b1c7, wasn't quite sure how to go about it

spec/defines/sensu_handler_spec.rb
madAndroid: I've added another spec to confirm that `handle_silenced` when it's left as the default of `false`

spec/defines/sensu_handler_spec.rb
```diff
@@ -177,4 +177,14 @@
...
+    let(:params) { { :command => 'mycommand.rb', :type => 'pipe', :handle_silenced => true } }
+    it { should contain_sensu_handler('myhandler').with_handle_silenced( true ) }
+  end
+
+  context 'handle_silenced set to default' do
```
ghoneycutt: this is the default behavior, which is great. We also need to explicitly test for setting it to false like you do with the test above, setting it to true.

manifests/handler.pp
ghoneycutt: Looks good!

madAndroid:
@ghoneycutt - Comments addressed, and all tests now passing :)

spec/defines/sensu_handler_spec.rb
madAndroid: I've now also added a spec to check when `handle_silenced` is explicitly set to `false`

ghoneycutt:
Sorry to ask you to change some of the things I asked here for, like type validation, though we've implemented data types and all PR's will now need to be rebased.

madAndroid:
@ghoneycutt - No worries, I've rebased, should be ready to go now

ghoneycutt:
Thanks for adding this new functionality!

Released in v2.28.0

madAndroid:
@ghoneycutt - Thanks! :)

Issue #515, opened 2016-07-05
redacting passwords from catalogue output


jaxxstorm:
I previously added the ability to use sensu's redaction feature to this module. I'd now like to stop the passwords that are set in `sensu::client_custom` being output into the catalogue output. Currently, you see things like this:

```
custom changed 'healthcheck => {"password"=>"a_password"}' to 'sensu => {"api"=>{"password"=>"another_password"}}'
```

This is suboptimal :)

There are 2 options, and I'd like to see which people think are best.

I can add an option so that the diff's aren't shown on `sensu::custom` or add another parameter, maybe `sensu::sensitive_values` and then have that ALWAYS not show diffs.

Thoughts?

ghoneycutt:
Related to issue #682 to use Puppet v4's data types. We can use the sensitive data type for passwords.

https://docs.puppet.com/puppet/4.10/lang_data_sensitive.html

jeffmccune:
With the use of `Sensitive`, we'll need to drop support for Puppet 4.5.0.

ghoneycutt:
I'm OK with that as we still support the last five versions.

ghoneycutt:
Puppet does not yet have the ability to test the usage of the Sensitive data type. Without this, we can't test and ultimately support it. Once this is testable, let's support it.

PR #754, opened 2017-07-17
WIP: 515 redact pw

alvagante:
This introduced Sensitive type for parameters that contain passwords. Related (but not resolutive) to #515


alvagante:
Based on branch 682-data-types

ghoneycutt:
@alvagante please squash your commits and rebase. I will hold off on merging anything else.

@jeffmccune could you please review.

jeffmccune:
> @jeffmccune could you please review.

Reviewing now.  Hello @alvagante !  Nice to be working with you, hope to see you soon at PuppetConf!

We need to make Travis happy, I'm able to reproduce the errors locally on my workstation with ruby 2.1.9.  I will add any other comments in the code review section of github.

manifests/init.pp
```diff
@@ -445,153 +445,130 @@
...
+  Boolean            $api = false,
+  Boolean            $manage_services = true,
+  Boolean            $manage_user = true,
+  Boolean            $manage_plugins_dir = true,
+  Boolean            $manage_handlers_dir = true,
+  Boolean            $manage_mutators_dir = true,
+  Variant[Undef,Integer,Pattern[/^(\d+)$/]] $rabbitmq_port = undef,
+  Optional[String]   $rabbitmq_host = undef,
+  Optional[String]   $rabbitmq_user = undef,
+  Optional[Sensitive[String]]   $rabbitmq_password = undef,
```
jeffmccune: With the vagrant workflow, `vagrant up sensu-server` is throwing:

```
==> sensu-server: Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Sensu]:                                                                        ==> sensu-server:   parameter 'rabbitmq_password' expects a value of type Undef or Sensitive, got String
```

Suggest taking a look at https://github.com/sensu/sensu-puppet/blob/v2.25.0/tests/sensu-server.pp#L7 which is how customers will likely be using the module.  I don't think we should need to change tests/sensu-server.pp as that vagrant provisioning manifest is expected to be compatible with Puppet 4.10 and this change set.

manifests/init.pp
```diff
@@ -445,153 +445,130 @@
...
+  Boolean            $redis_auto_reconnect = true,
+  Optional[Array]    $redis_sentinels = undef,
+  Optional[String]   $redis_master = undef,
+  Enum['rabbitmq','redis'] $transport_type = 'rabbitmq',
+  Boolean            $transport_reconnect_on_error = true,
+  String             $api_bind = '0.0.0.0',
+  String             $api_host = '127.0.0.1',
+  Integer            $api_port = 4567,
+  Optional[String]   $api_user = undef,
+  Optional[Sensitive[String]]   $api_password = undef,
```
jeffmccune: Similar to the error with rabbitmq_password using `vagrant up sensu-server`

```
==> sensu-server: Error: Evaluation Error: Error while evaluating a Resource Statement, Class[Sensu]:
...
==> sensu-server:   parameter 'api_password' expects a value of type Undef or Sensitive, got String at /vagrant/tests/sensu-server.pp:2:3 on node sensu-server.example.com
```

manifests/init.pp
```diff
@@ -445,153 +445,130 @@
...
-  $windows_package_name           = 'Sensu',
-  $windows_package_title          = 'sensu',
+  Pattern[/^absent$/, /^installed$/, /^latest$/, /^present$/, /^\d[\d\.\-\w]+$/] $sensu_plugin_version = 'installed',
+  Boolean            $install_repo = true,
+  Boolean            $enterprise = false,
+  Pattern[/^absent$/,/^installed$/,/^latest$/,/^present$/,/^[\d\.\-]+$/] $enterprise_version = 'latest',
+  Optional[String]   $enterprise_user = undef,
+  Optional[String]   $enterprise_pass = undef,
+  Boolean            $enterprise_dashboard = false,
+  String             $enterprise_dashboard_version = 'latest',
```
jeffmccune: Should this be a Pattern[..] as well, since it's a version parameter?

manifests/init.pp
```diff
@@ -445,153 +445,130 @@
...
+  Boolean            $manage_repo = true,
+  Enum['main','unstable'] $repo = 'main',
```
jeffmccune: This should be a generic string with an example for "main" and "unstable" in the documentation.  With an Enum, the module will require a new release if the repo identifiers are changed or new ones added.

manifests/init.pp
jeffmccune: My suggestion is to accept Variant[Sensitive, String].  If we get a String, immediately cast it to a sensitive.  This will maintain backwards compatibility, but will unfortunately require some refactoring to make sure only the sensitive variables are used internally.

Another approach is to just leave it as is and make this a breaking change.  @ghoneycutt Thoughts on if the sensitive parameters should be backwards compatible or not?  I think yes, because of Hiera.

ghoneycutt: We can drop support for Puppet v4.5 if that solves this. We want to support the last five releases and we kept 4.5 because it uses ruby 2.1.9 (older versions used an older version of ruby).

jeffmccune: @ghoneycutt Sensitive was introduced in 4.6.  Dropping 4.5 doesn't resolve the issue mentioned here.  The issue here is that in 4.6 and above the catalog will fail unless a value of type Sensitive is passed in.  This would be a breaking change as most everyone is likely passing in a String.

manifests/init.pp
alvagante: Yes indeed. There probably was no existing validation for version style. So I kept it generic.

manifests/init.pp
alvagante: This reproduces the current behaviour, where a check on main/unstable actually exists.

jeffmccune:
OK, I don’t have a strong preference either way since the manifest code is using a selector over this variable anyway.

manifests/init.pp
alvagante: Yeah, will allow parameters as strings. These were my first experiments with sensitive. It's not possible to make it reguired for every password.

manifests/init.pp
jeffmccune: OK, I don’t have a strong preference either way since the manifest code is using a selector over this variable anyway.

alvagante:
So, the PR with squashed commit for the whole type stuff is this: https://github.com/sensu/sensu-puppet/pull/746

here there are the added Sensitive stuff, but can merge on top of the other PR, which is more important. For the Sensitive type here I wonder that we might have to unwrap it if used for passwords used in templates, but then the secret remains in clear text in the generated template, the catalog and the diff reports, afaik.  Honestly it's the first time I use the Senstivie type, not really sure of the usage patterns in real world. Any suggestion?

ghoneycutt:
@alvagante could you please rebase this, now that we are using data types.

alvagante:
Going to make a new PR based on current master with just the sensitive stuff

PR #755, opened 2017-07-17
(#463) Ensure sensu::plugins are managed before checks

### LINKED ISSUE #463: plugins should install before checks
In the current puppet module the checks are set up before the plugins.

This causes some failing checks when a node is being configured, because some of my checks depend on the plugins. (which are shell scripts defined with a template path via the plugins hash)

To fix, the puppet module should install the plugins before it creates any checks or even starts the client up.

jeffmccune:
Without this patch plugins may be managed after the checks which use them. This
patch addresses the problem by adding a well known inert resource named
`Anchor[plugins_before_checks]`.  The sensu::check and sensu_check_config type
are managed after this anchor.  The sensu::plugin defined type is managed
before this anchor.  The anchor approach has been implemented over collections
to avoid realizing resources which the end user may be using.

Resolves #463

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -563,9 +563,7 @@ class sensu (
   Hash               $filter_defaults = {},
   Hash               $mutators = {},
   ### END Hiera Lookups ###
-
 ) {
-
   if $dashboard { fail('Sensu-dashboard is deprecated, use a dashboard module. See https://github.com/sensu/sensu-puppet#dashboards')}
   if $purge_config { fail('purge_config is deprecated, set the purge parameter to a hash containing `config => true` instead') }
   if $purge_plugins_dir { fail('purge_plugins_dir is deprecated, set the purge parameter to a hash containing `plugins => true` instead') }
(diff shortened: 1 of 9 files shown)


manifests/init.pp
```diff
@@ -666,6 +664,10 @@
     $_purge_mutators   = $full_purge_hash['mutators']
   }
 
+  # (#463) This well-known anchor serves as a reference point so all checks are
+  # managed after all plugins.  It must always exist in the catalog.
+  anchor { 'plugins_before_checks': }
```
ghoneycutt: We want to kill off the anchor pattern in the next big refactor after the data types. Is there another way we can do this, such as using the spaceship operator?

jeffmccune: I thought about using the spaceship operator but didn't because it's very likely some end users will have virtual or exported check resources.  Collections realize resources, which would be a PITA for the end users.

This isn't the full blown anchor pattern, which is more about class containment.  This is more of a lightweight, "Here's a single, well-known reference point to establish before and after."

I chatted with some people in the Puppet slack, and the best alternative to the anchor idea was https://github.com/binford2k/binford2k-manifold, but I did get some validating feedback a single anchor is reasonable in this situation and I like how simple it is to explain.  All plugins are managed before the anchor, all checks are managed after the anchor, the anchor always exists in the catalog.

jeffmccune:
Rebased.

ghoneycutt:
That was fast!

jeffmccune:
> That was fast!

Thanks!  Was a good place to switch context, so good timing on the merge.

ghoneycutt:
Released in v2.27.0

PR #756, opened 2017-07-18
(#748) Add puppet-strings gem and dependencies

### LINKED ISSUE #748: Add `rake doc` task to generate documentation from inline comments
Working with the types and providers, there is little documentation for the class and instance methods.   This isn't a huge problem, much of it is self-explanatory, however it would make maintenance of the module much more accessible to have YARD documentation for the types and providers.

As an example, I tend to reach for [this cheatsheet](https://gist.github.com/chetan/1827484) when adding code comments.

jeffmccune:
Build the documentation with `bundle exec rake puppet:strings`

The documentation is generated in every matrix cell for Travis.

RedCarpet is used to support Github Flavored Markdown.

The change to sensu_redis_config.rb is to avoid a bug in YARD, throwing
ArgumentError trying to parse the affected node in the syntax tree.  This patch
changes the code to be compatible with YARD.

Resolves #748

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
 
 sudo: false
 
-script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec'
+script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec strings:generate'
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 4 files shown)


jeffmccune:
Clean.  Doesn't need a rebase.

ghoneycutt:
Released in v2.27.0

PR #762, opened 2017-07-20
WIP: Added Sensitive type to password params

alvagante:
This replaces https://github.com/sensu/sensu-puppet/pull/754


spec/classes/sensu_rabbitmq_spec.rb
```diff
@@ -93,6 +93,24 @@
       ) }
     end # when using local key
 
+    context 'when passing password as Sensitive' do
```
ghoneycutt: We need tests for the other three params.

alvagante: Sure, this was the first quick one, will add the others once defined the best approach

ghoneycutt:
@alvagante What's the status on this?  Is it viable?

alvagante:
@ghoneycutt I was stuck in testing a case when sensitive type is used.
And honestly i have no idea on how to do it.
The implementation is done, but note, this approach does not hide passwords in puppet runs diffs and in the catalog.

manifests/init.pp
```diff
@@ -497,7 +497,7 @@
   Variant[Undef,Integer,Pattern[/^(\d+)$/]] $rabbitmq_port = undef,
   Optional[String]   $rabbitmq_host = undef,
   Optional[String]   $rabbitmq_user = undef,
-  Optional[String]   $rabbitmq_password = undef,
+  Variant[Undef,Sensitive,String] $rabbitmq_password = undef,
```
ghoneycutt: I think that you are seeing the sensitive info in the diff's is because the type is incorrect. Try `Variant[Undef, Sensitive]`

https://docs.puppet.com/puppet/4.10/lang_data_sensitive.html

alvagante: Actually I think we should keep Variant[Undef,Sensitive,String] to preserve backwards compatibility, have tests run without changes, and allow usage of a normal string as parameter.

My problem here is with the tests, code works:

```
Info: Applying configuration version '858c641 - Alessandro Franceschi, Thu Aug 17 16:49:10 2017 +0200 : Merge branch 'development' of github.com:example42/psick into development'
Notice: /Stage[main]/Sensu::Rabbitmq::Config/Sensu_rabbitmq_config[sensu.lab.psick.io]/password: changed [redacted] to [redacted]
Info: Class[Sensu::Rabbitmq::Config]: Scheduling refresh of Service[sensu-client]
Info: Class[Sensu::Rabbitmq::Config]: Scheduling refresh of Service[sensu-api]
Info: Class[Sensu::Rabbitmq::Config]: Scheduling refresh of Service[sensu-server]
Notice: /Stage[main]/Sensu::Client::Service/Service[sensu-client]: Triggered 'refresh' from 1 events
Notice: /Stage[main]/Sensu::Api::Service/Service[sensu-api]: Triggered 'refresh' from 1 events
Notice: /Stage[main]/Sensu::Server::Service/Service[sensu-server]: Triggered 'refresh' from 1 events
Notice: Applied catalog in 7.76 seconds
```
but I don't know how to write tests for the Sensitive type.

alvagante: @ghoneycutt about the diff, I actually talked based on the wrong assumption that we were managing the file content via the file type and not via custom types.

alvagante: @ghoneycutt So, a working but not optimal solution would be to merge this pr without the tests, and leave them for later or to more experts hands.

ghoneycutt:
@Phil-Friderici could you please help with the tests here

Phil-Friderici:
I'll have a look at it and come back.

Phil-Friderici:
Seems we have to wait before we can successfully test sensitive data types.
PR https://github.com/rodjek/rspec-puppet/pull/464 seems to bring what we need here, but isn't merged yet.
Issue https://github.com/rodjek/rspec-puppet/issues/529 reports mostly the same, just the implementation is different.

Haven't found a way to pass the data type sensitive (`Sensitive('password')`) to the tests. With that in place we could at least get some very basic tests running.

ghoneycutt:
Thanks for figuring this out! Let's hold off on adding this functionality until it is able to be tested.

PR #765, opened 2017-07-21
(#759) Add reference spec tests for sensu_check JSON provider

### LINKED ISSUE #759: Create a reference implementation for provider spec tests
The sensu module does a great job of exercising the behavior of Puppet classes, defined types, and the native types in the module.  The providers lack test coverage, however.

The goal of this issue is to create a reference implementation of spec tests covering the behavior of the `sensu_check` provider and the `custom` property specifically.  These spec tests will serve as a foundation to build spec tests for other providers.

The custom property has been chosen because it accepts an arbitrary map which is merged with the other properties.  This behavior is related to how all types in this module could become compatible with arbitrary attributes added in the future.

jeffmccune:
Without this patch there are no spec tests for the sensu_check JSON provider.
This is problem because a reference is needed to specify the expected behavior
of all providers.

This patch implements a pattern of stubbing out the filesystem.  All reads and
writes in the provider itself are routed through the `read_file` and
`write_json_object` methods.  The RSpec tests then use rspec-mocks to stub out
the reads and set expectations on the output.

This reference may be applied to any provider using the `flush` method.  The
setter methods in the provider for each property are expected to update state
in an instance variable, conventionally named @property_flush but named @conf
in the sensu_check provider.  The flush method is responsible for writing out
@property_flush (@conf), which we intercept and set expectations on the data
provided.

Resolves #759

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ end
 group :development, :unit_tests do
   gem 'rake',                                             '< 11.0.0'
   gem 'rspec-puppet', '~> 2.5.0',                         :require => false
+  gem 'rspec-mocks',                                      :require => false
   gem 'puppetlabs_spec_helper', '>= 2.0.0',               :require => false
   gem 'puppet-lint', "~> 2.0",                            :require => false
   gem 'json', "~> 1.8.3",                                 :require => false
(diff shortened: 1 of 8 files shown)


ghoneycutt:
Released in v2.28.0

PR #766, opened 2017-07-24
WIP 582  Don't do type convertion on keys of sensu_client_config custom param

alvagante:


ghoneycutt:
Thanks!!

ghoneycutt:
Released in v2.28.0

PR #768, opened 2017-07-25
Add validation of spec/fixtures/unit/**/*.json

ghoneycutt:
Validates that these files are JSON files.


ghoneycutt:
Released in v2.28.0

PR #780, opened 2017-07-27
[WIP] Add sensu::check config property and remove sensu_check_config

### LINKED ISSUE #764: Remove sensu_check_config
Garrett and I discussed this today.  There is confusion surrounding sensu_check_config, the JSON provider doesn't seem to be well-behaved, the type isn't used anywhere in the module, there isn't any documentation, and the sensu::check type appears to provide all the necessary functionality for checks.

One possible situation is sensu_check_config is able to set keys and values at the same level as "checks" in the target file, whereas sensu::check can only set custom configuration nested under checks['mycheck'].

If it is necessary to add configuration in the top-level map of the JSON files in `conf.d/checks/`, then sensu::check should take a `config` property which is merged together with the `checks` top level key.

jeffmccune:
= Overview of work

 - [x] Add `config` property to sensu::check
 - [x] Ensure JSON maps are written in sort-order from the top level rather
   than the nested level of checks.mycheck.properties
 - [ ] Test arbitrary config already exists, does the default of {} prune it?  It shouldn't be removed by default, it should be preserved if config is undef.
 - [ ] Test arbitrary config already exists, does specifying `config => { 'foo' => 'bar' }` remove it?
 - [ ] Test arbitrary config nested under `"checks"` at the same level as `"mycheck"`.  Is it removed when `config => { 'foo' => 'bar' }`?  It should be.
 - [ ] Test the behavior when the config file doesn't exist at all.
 - [ ] Remove sensu_check_config related files.

= Add config property to sensu::check and sensu_check

Without this patch there is no way to specify arbitrary configuration at the top
level of the sensu check configuration object.  This patch addresses the problem
by adding a new property to sensu::check and sensu_check named `config`.  This
property accepts a Hash map and merges all other configuration on top of this
base starting point.

This patch also adds a new mix-in method, `PuppetX::Sensu::SortHash#sort_hash`.
This method returns a hash sorted recursively, meaning all nested hash values
are sorted in the same manner.  This method is intended for use in the #flush
method of all Sensu providers.

= Remove sensu_check_config

TBD

Resolves #764

### DIFF
--- a/lib/puppet/provider/sensu_check/json.rb
+++ b/lib/puppet/provider/sensu_check/json.rb
@@ -1,5 +1,7 @@
 require 'json' if Puppet.features.json?
 require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
+                                   'puppet_x', 'sensu', 'sort_hash.rb'))
+require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
                                    'puppet_x', 'sensu', 'provider_create.rb'))
 require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..',
                                    'puppet_x', 'sensu', 'to_type.rb'))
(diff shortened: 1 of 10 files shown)


jeffmccune:
Closing to pursue sensu_write_json() instead of adding a config propety as per comment in #764.

Issue #782, opened 2017-07-27
When removing a check, sensu service not refresh


jothoma1:
## Description of problem

- What did you do?
I'm trying to remove a check with

> ensure => absent
- What happened?
It works but it did not refresh the service sensu-client
- What did you expect to happen?
Refresh the service in order to have the good configuration.
- How can someone reproduce the problem?
1) Create a check
2) try to remove it with ensure => absent
3) sensu has not refresh so the check is still present

## Command used and debugging output

- What does your puppet config look like (including any hiera config)
- Is this a masterless or master based puppet setup?

## Platform and version information

- Your OS:  CentOS 7
- Your Ruby version: 2.4
- Your version of Puppet:  3.8.7
- Your version of Sensu:  1.0
- Your version of this module:

alvagante:
@jothoma1 https://github.com/sensu/sensu-puppet/pull/785 should address this, can you confirm that?

alvagante:
@ghoneycutt I can confirm that this issue is now fixed in current master, so I suppose it can be closed
```
==> ubuntu1604-client: Notice: /Stage[main]/Main/Sensu::Check[check_cpu]/Sensu::Write_json[/etc/sensu/conf.d/checks/check_cpu.json]/File[/etc/sensu/conf.d/checks/check_cpu.json]/ensure: removed
==> ubuntu1604-client: Notice: /Stage[main]/Sensu::Client/Service[sensu-client]: Triggered 'refresh' from 1 events

==> el7-client: Notice: /Stage[main]/Main/Sensu::Check[check_cpu]/Sensu::Write_json[/etc/sensu/conf.d/checks/check_cpu.json]/File[/etc/sensu/conf.d/checks/check_cpu.json]/ensure: removed
==> el7-client: Notice: /Stage[main]/Sensu::Client/Service[sensu-client]: Triggered 'refresh' from 1 events
```

ghoneycutt:
Released in v2.50.0

PR #785, opened 2017-07-27
(#783) Add sensu::check content parameter, use sensu::write_json

### LINKED ISSUE #783: Modify sensu::check to use defined type sensu::write_json instead of native type sensu_check
In an effort to support arbitrary configuration scopes for plugins and extensions as mentioned in #764, the goal of this card is to implement a new parameter to the sensu::check defined type.  This parameter needs a reasonable name (`config` is too generic and over-loaded).  The desired behavior is  to have the new parameter merged with all other parameters to produce a single, top-level JSON object written to the filesystem using sensu::write_json.

For the purposes of the change, two scopes are considered.  All configuration in-scope is nested under `checks.mycheck` for a check named `mycheck`.  All configuration out-of-scope is outside the nested `checks.mycheck` map.  All out-of-scope configuration is managed by the new parameter, all in-scope configuration is managed by the existing parameters.

The new parameter is akin to the existing `custom` parameter, just at the top level instead of nested inside the check configuration scope.

jeffmccune:
sensu::write_json
===

Without this patch arbitrary to-level configuration scope items cannot be
managed in a sensu check.  This patch addresses the problem by adding a
`content` parameter to the sensu::check defined type.

This patch changes the defined type to pass a JSON object to sensu::write_json
instead of passing each named property to the sensu_check custom type.
sensu::write_json is used to ensure all attributes in a configuration file are
manageable and arbitrary attributes need only be placed in the `content` hash
map, or have a new parameter defined in the `sensu::check` defined type.

When adding a new parameter, as long as the parameter and value are passed as a
key/value pare to the `sensu_check_content()` function, the attribute and value
will be placed in the check configuration scope under checks.mycheck.  The same
behavior may be achieved by passing the key and value via the `custom`
parameter.

The ultimate goal of this work is to eliminate the `custom` parameter and
eventually refactor all defined types in the Puppet module to use
sensu::write_json and this approach.

Remaining work
===

 - [x] Fix failing tests which are now marked pending.
 - [x] Add new expectations and fixture data for the expected output of
       sensu::check via the content property of sensu::write_json.
 - [x] Vagrant testing of example checks, ensure backwards compatibility.

Resolves #783

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -70,6 +70,11 @@
 #   Note: The validity of the other checks is not enforced by puppet
 #   Set this to 'absent' to remove it completely.
 #
+# @param content Mapping of arbitrary attributes from the top-level of the target
+#   configuration JSON map.  This parameter is intended to configure plugins and
+#   extensions which look up values outside of the check configuration scope.
+#   Example: { "mailer" => { "mail_from" => "sensu@example.com", "mail_to" => "monitor@example.com" } }
+#
 # @param custom List of custom attributes to include in the check.
 #   You can use it to pass any attribute that is not listed here explicitly.
 #   Example: { 'remediation' => { 'low_remediation' => { 'occurrences' => [1,2], 'severities' => [1], 'command' => "/bin/command", 'publish' => false, } } }
@@ -107,6 +112,7 @@ define sensu::check (
   Variant[Undef,Enum['absent'],Boolean] $publish = undef,
   Variant[Undef,String,Array]           $dependencies = undef,
   Optional[Hash]                        $custom = undef,
+  Hash                                  $content = {},
   Variant[Undef,Enum['absent'],Integer] $ttl = undef,
   Variant[Undef,Enum['absent'],Hash]    $subdue = undef,
   Variant[Undef,Enum['absent'],Hash]    $proxy_requests = undef,
@@ -161,17 +167,9 @@ define sensu::check (
   Anchor['plugins_before_checks']
   ~> Sensu::Check[$name]
 
-  file { "${conf_dir}/checks/${check_name}.json":
-    ensure => $ensure,
-    owner  => $user,
-    group  => $group,
-    mode   => $file_mode,
-    before => Sensu_check[$check_name],
-  }
-
-  sensu_check { $check_name:
-    ensure              => $ensure,
-    base_path           => "${conf_dir}/checks",
+  # This Hash map will ultimately exist at `{"checks" => {"$check_name" =>
+  # $check_config}}`
+  $check_config_start = {
     type                => $type,
     standalone          => $standalone,
     command             => $command,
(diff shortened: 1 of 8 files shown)


manifests/check.pp
```diff
@@ -236,10 +238,53 @@
...
     proxy_requests      => $proxy_requests,
-    require             => File["${conf_dir}/checks"],
-    notify              => $::sensu::check_notify,
     ttl                 => $ttl,
   }
+
+  # Produce a new Hash map suitable for use with sensu::write_json.  The mapping
+  # behavior of Puppet parameters to scoped Sensu configuration attributes is
+  # implemented as a function.
+  $content_hash = sensu_check_content($resource_hash)
```
jeffmccune: @ghoneycutt I'd love to get your early feedback on this approach of using a parser function to create the output hash for sensu::write_json using the existing parameters of the defined type as input.

manifests/check.pp
```diff
@@ -107,11 +107,18 @@
 #   Set this to 'absent' to remove it completely.
 #   Default: undef
 #
+# [*content*]
```
ghoneycutt: This seems to be the same thing as custom, so I don't think we need it on this iteration.

Next iteration would be to remove all parameters and only have `content`. At that point we can use the Struct data type for validation.

manifests/check.pp
```diff
@@ -236,10 +238,53 @@
     publish             => $publish,
     dependencies        => $dependencies,
     custom              => $custom,
```
ghoneycutt: I think this would create a key named `custom` with the data in it and so it should be removed. We can join the hashes like

```puppet
$content_hash = $resource_hash + $custom
```

jeffmccune: That's effectively what sensu_check_content() does, with content and custom.  It could definitely be a little cleaner.  I used a function because my first pass at manipulating immutable Hashes the Puppet DSL turned out awkward.  Is there a better way to handle the behavior of content and custom to get the final JSON?

Something like:

```
# as before, with name, custom and content stripped out.
$resource_hash = { ... }
# Final structure of the output.
$content_hash = $content + {'checks' => { "$check_name" => $resource_hash + $custom } }
```

The only other thing is to deal with all of the undefined and `"absent"` values, so this should work:

```puppet

$resource_hash_real = $resource_hash.reduce({}) |$memo, $kv| {
  if ($kv[1] == "absent" or $kv[1] == undef) {
    $memo
  } else {
    $memo + Hash.new($kv)
  } 
}
```

I'll replace the function with this Puppet DSL and some comments describing the structure.

manifests/check.pp
jeffmccune: They're slightly different.  Custom is a hash which gets merged into the scope of the check configuration at checks.mycheck.  Content is a hash merged onto the top level JSON object.  Without content, there isn't a way to specify configuration for the [sensu-mailer-plugin](https://github.com/sensu-plugins/sensu-plugins-mailer) example.

I kept this in the define, please take a look at the Puppet DSL behavior replacing the parser function in the latest patch.

manifests/check.pp
jeffmccune: Behavior has been moved to the Puppet DSL in the latest patch.  I'll proceed with it, let me know if you'd still like `content` removed after taking a look at the structure being built up.

jeffmccune:
@ghoneycutt This is ready for another review.  All of the tests have been updated and are passing and `sensu::check` behavior is implemented entirely in the Puppet DSL now, no custom Ruby code.  This should make it much easier to adapt for the other defined types.

manifests/check.pp
```diff
@@ -235,11 +239,61 @@
...
+    $checks_scope = { 'checks' => $content['checks'] + $checks_scope_start }
+  }
+
+  # The final structure from the top level.  Check configuration scope is merged
+  # on top of any arbitrary plugin and extension configuration in $content.
+  $content_real = $content + $checks_scope
+
+  # Compute the services to notify
+  $notification_map = {
+    'Class[Sensu::Client::Service]' => $::sensu::client,
```
jeffmccune: This should also address #782 but need to test it out.

jeffmccune: I think it is:

```patch
diff --git a/tests/sensu-server.pp b/tests/sensu-server.pp
index c0a6616..1106403 100644
--- a/tests/sensu-server.pp
+++ b/tests/sensu-server.pp
@@ -64,6 +64,7 @@ node 'sensu-server' {

   # Example check using the cron schedule.
   sensu::check { 'remote_check_ntp':
+    ensure         => 'absent',
     command        => 'PATH=$PATH:/usr/lib64/nagios/plugins check_ntp_time -H :::address::: -w 30 -c 60',
     standalone     => false,
     handlers       => 'default',
```

```
$ sudo puppet apply /vagrant/tests/sensu-server.pp
Notice: Compiled catalog for sensu-server.example.com in environment production in 0.78 seconds
Notice: /Stage[main]/Main/Node[sensu-server]/Sensu::Check[remote_check_ntp]/Sensu::Write_json[/etc/sensu/conf.d/checks/remote_check_ntp.json]/File[/etc/sensu/conf.d/checks/remote_check_ntp.json]/ensure: removed
Notice: /Stage[main]/Sensu::Client::Service/Service[sensu-client]: Triggered 'refresh' from 1 events
Notice: /Stage[main]/Sensu::Api::Service/Service[sensu-api]: Triggered 'refresh' from 1 events
Notice: /Stage[main]/Sensu::Server::Service/Service[sensu-server]: Triggered 'refresh' from 1 events
Notice: Applied catalog in 4.24 seconds
```

manifests/check.pp
ghoneycutt: Why can't sensu-mailer-plugin be specified using custom?

jeffmccune: They're different configuration scopes.  Mailer is up at top-level (the same as "checks").  Custom adds key/value pairs two levels deep within the checks.mycheck scope.

ghoneycutt: It was my understanding that custom adds key/values at top scope. So if you want data from custom to be in the checks scope you need the hash in custom specified as such.

Would that work? Does Puppet's $hash + $another_hash work for that?

jeffmccune: > It was my understanding that custom adds key/values at top scope. So if you want data from custom to be in the checks scope you need the hash in custom specified as such.

We could change the behavior to operate like that, but today the value of custom is assumed to be a hash merged with the value of checks.mycheck.  Check it out at https://github.com/sensu/sensu-puppet/blob/v2.30.0/lib/puppet/provider/sensu_check/json.rb#L92

> Would that work? Does Puppet's $hash + $another_hash work for that?

We'd probably want to use the deep_merge() function in stdlib.  The `+` operator isn't a deep merge, so duplicated keys result in the rightmost value clobbering the other values.

An example of the gymnastics I'm doing with the Puppet DSL to merge nested hash values is at https://github.com/sensu/sensu-puppet/pull/785/commits/cddda1a07d78c2699f10b07fbfb2c5376358311e#diff-7bfaf46cbbb09c017062dc6fbdfdd383R262

jeffmccune:
This has been rebased.

ghoneycutt:
Thank you!!

Released in v2.32.0

Issue #757, opened 2017-07-18
Refactor inline documentation to puppet strings (yard) format


jeffmccune:
After #756 and #682 are merged, the inline documentation needs to be updated to the new `puppet strings` format.  This format is described at https://github.com/puppetlabs/puppet-strings/blob/master/README.md

Potentially out of scope, but helper methods for types and providers should be moved into the PuppetX namespace and have documentation added.

alvagante:
@ghoneycutt  going to start with this, so that I learn better the module params ;-)
Can't self assign it, anyway.

alvagante:
#792 merged, ticket can be closed.

PR #792, opened 2017-08-09
Puppet strings 4 all #757

alvagante:


lib/puppet/type/sensu_api_config.rb
```diff
@@ -1,3 +1,14 @@
+# @!puppet.type.param name
```
ghoneycutt: you doing the rest of the types/providers in another pass?

alvagante: @ghoneycutt yes, but actually I was wondering if all the current types, and relevant providers, are going to remain as is or some of them are will be replaced by the works @jeffmccune is doing in https://github.com/sensu/sensu-puppet/pull/785 will affect some of the current ones, and their params. If so I wonder if it makes sense to prepare docs for types that are going to change soon.

ghoneycutt: Makes sense. Don't bother with the ones that are changing and we'll make another pass after the dust settles.

alvagante:
@ghoneycutt This is ready for merging, from my point of view. I've added Puppet 4 params validations in a pair of classes where it was missing, in order to have valid entries in puppet strings and to align those classes to the others.

ghoneycutt:
Thanks @alvagante 

Released in v2.31.0

Issue #791, opened 2017-08-09
Remove apt module from metadata


ghoneycutt:
The apt module is a soft dependency. It should be removed from the metadata, though noted in the README.

ghoneycutt:
Released in v2.31.0

PR #793, opened 2017-08-09
Remove soft dependencies on apt and powershell

ghoneycutt:


ghoneycutt:
Resolves #791

ghoneycutt:
Released in v2.31.0

Issue #760, opened 2017-07-19
Vagrant uses an older version of rabbitmq


ghoneycutt:
puppetlabs/rabbitmq is at 4.1.0 and has been transferred to voxpupuli which is at 5.6.0.

Using the old version throws warnings in our vagrant setup. We should change puppetlabs/rabbitmq to voxpupuli/puppet-rabbitmq and test that all vagrant systems still work as expected.

ghoneycutt:
@Phil-Friderici this is a great one to become a pro with the vagrant setup

ghoneycutt:
The module is still known as puppetlabs-rabbitmq

PR #794, opened 2017-08-14
(GH-760) Document rabbitmq's move to Voxpupuli

ghoneycutt:


ghoneycutt:
Resolves #760

ghoneycutt:
Released in v2.32.0

PR #802, opened 2017-08-24
Update the README to clarify support resources

obfuscurity:
Users should rely on Sensu, not Puppet Inc, for support of this module. This clarification was requested by our friends at Puppet.

@ghoneycutt @agoddard Any concerns wrt directing users to our support page rather than opening issues here instead?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 Installs and manages the open source monitoring framework [Sensu](http://sensuapp.org).
 [![Puppet Forge](http://img.shields.io/puppetforge/v/sensu/sensu.svg)](https://forge.puppetlabs.com/sensu/sensu)
 
+Please note, that this is a **Partner Supported** module, which means that technical customer support for this module is solely provided by Sensu. Puppet does not provide support for any **Partner Supported** modules. Technical support for this module is provided by Sensu at https://sensuapp.org/support.
+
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Some projects such as Puppet will disable issues and have the community use their ticketing system. If this is highly valuable for Sensu and is done across all the projects, then we should do it, but it does come at a cost. Users have to have an account there and it takes extra steps. The extra steps add friction to the process that will discourage some folks from participating.

obfuscurity:
I'm not suggesting that we take community participation out of GitHub. We're talking about end-user support (i.e. GitHub issues vs Sensu support) and making it clear to users that where they should expect to get support from (read: not Puppet Inc). There's certainly some grey area in there which is why I wanted to discuss and clarify it here before we commit any changes.

agoddard:
👍 I think we should make it clear where they should go for support when they have questions about the module not working as expected etc. We can then create issues against the module here. I don't think we should remove the ability of users to add issues here themselves though, as this might be what they're used to and we should let them continue if they choose. Obviously the SLAs won't apply with issues submitted directly to the repo though.

obfuscurity:
SGTM, thanks @agoddard @ghoneycutt!

ghoneycutt:
v2.33.1

PR #803, opened 2017-08-25
Force array for sensu::check subscribers, handlers and aggregates params #801

alvagante:
# Pull Request Checklist

This commit converts strings to arrays with a single element, when a String is used for the handlers, subscribers and aggregates parameters in sensu::check

## Description
Parameter type is evaluated, if i'ts s String is converted to an array and a deprecation notice is cast.

## Related Issue
This fixes #801 

## How Has This Been Tested?
Tested on sensu-server vagrant vm and spec tests.

## General

- [ ] Update `README.md` with any necessary configuration snippets

- [ ] New parameters are documented

- [x] New parameters have tests

- [x] Tests pass - `bundle exec rake validate lint spec`

PR #804, opened 2017-08-26
Force array for some sense::check params #801

### LINKED ISSUE #801: Checks not working as expected
From @alvagante 

@ghoneycutt I think the error is due to how checks files are created:
if a single check is added in a conf.d/checks/ file then the json file should begin with:
```
{
    "check": {
```
and not 
```
{
    "checks": {
```

manually changing the files in conf.d/checks/ on the Vagrant sensu-server vm allows the api service to start.

alvagante:
# Pull Request Checklist

Changed sensu::check define and relevant spec tests.

## Description
This commit converts strings to arrays with a single element, when a String is used for the handlers, subscribers and aggregates parameters in sensu::check

This PR replaces  #803 

## Related Issue

Fixes #801 

## Motivation and Context
Syntax errors on check files when using strings on the given params.
 
## How Has This Been Tested?
Tested on sensu-server vagrant vm and spec tests.

## General

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -95,7 +95,7 @@ define sensu::check (
   Enum['present','absent']              $ensure = 'present',
   Optional[String]                      $type = undef,
   Variant[Undef,String,Array]           $handlers = undef,
-  Optional[Array]                       $contacts = undef,
+  Variant[Undef,String,Array]           $contacts = undef,
   Variant[Boolean,Enum['absent']]       $standalone = true,
   String                                $cron = 'absent',
   Variant[Integer,Enum['absent']]       $interval = 60,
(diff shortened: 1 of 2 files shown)


alvagante:
@ghoneycutt this is finally a clean PR on #801 .
But I think we may reconsider the opportunity to throw a deprecation notice.
I see in the README (which would need update if we keep the current code) examples like this:
```
  sensu::check { 'check_ntp':
    command     => 'PATH=$PATH:/usr/lib/nagios/plugins check_ntp_time -H pool.ntp.org -w 30 -c 60',
    handlers    => 'default',
    subscribers => 'sensu-test'
  }
```
which means that a lot of existing code has strings where we expect an array. 
I would just remove the deprecation notice, and convert string to array when needed.
That's probably the original intention.

ghoneycutt:
@alvagante Good eye, I agree. Let's just silently convert to array for our implementation.

alvagante:
@ghoneycutt ready for merge

ghoneycutt:
Thanks @alvagante 

Released in v2.33.1

PR #805, opened 2017-08-30
Added http_socket param to client config #776

### LINKED ISSUE #776: Client config should support http_socket
https://sensuapp.org/docs/latest/reference/clients.html#http-socket-attributes

alvagante:
# Pull Request Checklist

Added parameter http_socket to client config.

## Description
Added parameter to main class, and then reference to it in sunse::client is passwd to the sensu_client_config type used the new json write approach.
 
## Related Issue

Fixes #776 

## How Has This Been Tested?
Tested os samples vms and spec.

## General

### DIFF
--- a/spec/unit/sensu_client_config_spec.rb
+++ b/spec/unit/sensu_client_config_spec.rb
@@ -62,4 +62,22 @@ describe Puppet::Type.type(:sensu_client_config) do
       end
     end
   end
+  describe 'http_socket' do
+    subject { described_class.new(resource_hash)[:http_socket] }
+    context 'in the default case' do
+      it { is_expected.to be_nil }
+    end
+    http_socket = {
+      'bind' => '127.0.0.1',
+      'port' => '3031',
+      'user' => 'sensu',
+      'password' => 'sensu'
+    }    
+    context '=> custom values' do
+      let(:resource_hash_override) { {http_socket: http_socket} }
+      it { is_expected.to eq(http_socket) }
+    end
+
+  end
+
 end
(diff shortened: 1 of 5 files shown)


spec/unit/sensu_client_config_spec.rb
```diff
@@ -62,4 +62,22 @@
       end
     end
   end
+  describe 'http_socket' do
```
ghoneycutt: Could you add a test that reads a fixture file and sees that the json created matches like what is done in `spec/defines/sensu_check_spec.rb`.

alvagante: @ghoneycutt sorry, I spent 2 hours in trying to make it and I didn't succeed, so I prefer to raise hand before engaging further efforts.
First, I'm not 100% sure on where you want the test, on the sensu_client_config type (here) on n the sensu::client class.
I've tried with both , but haven't any usable resource to compare directly  to the final generated json file, in spec/classes/sensu_client_spec.rb (where I could use an approach similar to the spec/defines/sensu_check_spec.rb) I can only tests parameters passed to the the sensu_client_config type, and is is already done.
In spec/unit/sensu_client_config_spec.rb I don't know what to compare to the content of the fixture file, I can test single params, but haven't found a way to do something based on the whole file.
IMHO current tests, and tested implementation on real VM, could be enough, considering that also for other params of sensu_client_config do don't tests agains fixture json files.
But I leave to you the final word here and, eventually, the support to finalise the tests.

ghoneycutt:
Thanks! Released in v2.34.0

Issue #809, opened 2017-09-05
transport.json not created when transport_type = rabbitmq


htallur:
## Description of problem
transport.json is not created 

incorrect logic in transport.pp
-----------
  if $::sensu::transport_type != 'redis' {
    $ensure = 'absent'
  } else {
    $ensure = 'present'
  }
-------------

ensure = 'absent',  when transport_type = 'redis' in transport.pp

alvagante:
@htallur what's the expected behaviour? I suppose that a transport.json has to be created in any case , both with redis and rabbitmq transport .

I can eventually add a parameter $ensure to the sensu::transport class, to allow users to eventually force the removal of transport.json

ghoneycutt:
Released in v2.37.0

PR #811, opened 2017-09-07
Addressing #809 Added  param to sensu::transport

alvagante:
# Pull Request Checklist

Solves issue #809 

## Description
This solves issues #809 by exposing an exnsure parameter to allow users to manage the presence of trnsport.json and by ensuring it's present if transport_type is either redis or rabbitmq.

Need double check if this is the expected and correct behaviour.

## Related Issue
Fixes # 809

## General

### DIFF
--- a/manifests/transport.pp
+++ b/manifests/transport.pp
@@ -1,14 +1,12 @@
 # @summary Configures Sensu transport
 #
-# Configure Sensu Transport
+# Configures Sensu transport
 #
-class sensu::transport {
-
-  if $::sensu::transport_type != 'redis' {
-    $ensure = 'absent'
-  } else {
-    $ensure = 'present'
-  }
+# @param ensure If to create or remove the transport.json file
+#
+class sensu::transport (
+  Enum['present','absent'] $ensure = 'present',
+) {
 
   $transport_type_hash = {
     'transport' => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,22 @@
 # Change Log
 
-## [v2.33.1](https://github.com/sensu/sensu-puppet/tree/HEAD)
+## [v2.34.0](https://github.com/sensu/sensu-puppet/tree/v2.34.0)
 
-[Full Changelog](https://github.com/sensu/sensu-puppet/compare/v2.33.0...HEAD)
+[Full Changelog](https://github.com/sensu/sensu-puppet/compare/v2.33.1...v2.34.0)
+
+**Closed issues:**
+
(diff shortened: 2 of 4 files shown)


manifests/transport.pp
```diff
@@ -1,14 +1,12 @@
...
-
-  if $::sensu::transport_type != 'redis' {
-    $ensure = 'absent'
-  } else {
-    $ensure = 'present'
-  }
+# @param ensure If to create or remove the transport.json file
+#
+class sensu::transport (
+  Enum['present','absent'] $ensure = 'present',
```
ghoneycutt: Since we're adding a param, this would need the param to be documented as well as tests, unless I'm missing something here.

alvagante: Actually I tried to make tests, but it's a bit tricky, if not impossible, to test directly the sensu::transport class, as all the other classes which are declared in the main class and use it as entry point to parameters.
To test it's behaviour we should pass via the main class (https://github.com/sensu/sensu-puppet/blob/master/spec/classes/sensu_transport_spec.rb#L4), and this would have required to add an additional parameters like sensu::transport_ensure, which seemed too much for me.

ghoneycutt: @Phil-Friderici Any thoughts?

ghoneycutt: ping @Phil-Friderici

Phil-Friderici:
Indeed the class sensu::transport doesn't work on it's own and therefore is a little bit tricky to test. You need to use (:pre_condition) to add the class sensu with the needed parameters.

I have created a test [1] and sent a PR to the 809 branch of @alvagante that does the trick.

[1] https://github.com/Phil-Friderici/sensu-puppet/blob/809/spec/classes/sensu_transport_spec.rb

alvagante:
That's a quite neat trick @Phil-Friderici , to define a bogus sensu class and declare it in the precondition, I mea.
I think that if the model works well, it could eventually be extended to other classes: most of them are declared from main sensu class, and the relevant tests are done on it. Our tests declarations would change as follows
```
-describe 'sensu' do
+describe 'sensu::transport' do
```

Phil-Friderici:
Yepp that should work.

In the longer run these _subclases_ should become either integrated into the main class or should be refactored to become real classes (that could live on their own).

alvagante:
Going to close this MR, please use #832 , which takes to most direct and simple approach without the addition of a mostly useless parameter for which testing would have been a bit excessive. Thanks for @Phil-Friderici for the hints on tests, they've somehow opened my mind :-D

Phil-Friderici:
@alvagante The pleasure was on my side, but thanks for lifting me up today ;)

PR #812, opened 2017-09-07
Added register and registration params to client config #749

### LINKED ISSUE #749: Add client register and registration client configs
Nearly identical to, #550 the client configuration should support registration attributes.

https://sensuapp.org/docs/latest/reference/clients#registration-attributes

alvagante:
# Pull Request Checklist

This implements client registration settings for client config.

## Description
I've made this completely similar to the de-register configurations, but I'm now sure we actually need the register parameter (in sensu docs only registration is mentioned)

The service starts anyway, as Sensu is tolerant of unknown settings in client.json. 

Fixes #749 

## General

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -455,6 +455,12 @@ class sensu (
     $client_service = undef
   }
 
+  if $enterprise and $manage_services {
+    $enterprise_service = Service['sensu-enterprise']
+  } else {
+    $enterprise_service = undef
+  }
+
   if $server {
     $server_service_class = Class['sensu::server::service']
   } else {
@@ -467,7 +473,7 @@ class sensu (
     $api_service = undef
   }
 
-  $check_notify = delete_undef_values([ $client_service, $server_service_class, $api_service ])
+  $check_notify = delete_undef_values([ $client_service, $server_service_class, $api_service, $enterprise_service ])
 
   # Because you can't reassign a variable in puppet and we need to set to
   # false if you specify a directory, we have to use another variable.
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -162,6 +162,15 @@
 #   status, and issued timestamp. The following attributes are provided as
 #   recommendations for controlling client deregistration behavior.
 #
+# @param client_register Enable the [registration
+#   event](https://sensuapp.org/docs/latest/reference/clients#registration-attributes)
+#   if true
+#
+# @param client_registration [Attributes](https://sensuapp.org/docs/latest/reference/clients#registration-attributes)
+#   used to generate check result data for the registration event. Client
+#   registration attributes are merged with some default check definition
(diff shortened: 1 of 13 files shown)


manifests/init.pp
```diff
@@ -363,6 +372,8 @@
   Hash               $client_custom = {},
   Variant[Undef,Boolean] $client_deregister = undef,
   Variant[Undef,Hash] $client_deregistration = undef,
+  Variant[Undef,Boolean] $client_register = undef,
```
ghoneycutt: What if you specify `client_register = true` and `client_deregister = true` ? 

Perhaps we need some logic so you can't do mutually exclusive things?

alvagante: Yes we can, but are they mutually exclusive?

ghoneycutt: I'm not positive, what do the docs say?

alvagante: Docs don't mention any mutuality.
I think that registration and deregistration can both be configured, 
my concern is about the boolean register parameter, which is not mentioned in the docs (but deregister is).
My understanding based on https://sensuapp.org/docs/1.0/reference/clients.html#client-attributes is that we should just have registration , and not register, and they can cohexist with the de-reg* ones. But maybe it's better to ask to Sensu guys.

ghoneycutt: @agoddard Could you please shed some light on this for us?

cwjohnston: @ghoneycutt @alvagante 

* `deregister` should be a boolean, it controls whether or not the client will send a "deregistration event" upon clean shutdown

* `register` should be a hash, the contents of which is passed along when the client first registers with the Sensu installation.

* `deregister` and `register` are not mutually exclusive

alvagante:
Uhm, I don't know if this message written as a new review a few days ago went through, so I repost it:

---
@cwjohnston @ghoneycutt sorry for late feedback loop here.
So, looking at https://sensuapp.org/docs/latest/reference/clients.html#client-attributes I suppose @cwjohnston has written client_register meaning client_registration.
So I would remove the not existing client_register parameter and keep the client_registration hash.

If a client_register boolean actually exists, it's not mentioned in the docs.
---

Going to send a PR based on these assumptions.

alvagante:
@ghoneycutt https://github.com/sensu/sensu-puppet/pull/831 replaces this one.
@cwjohnston PR in https://github.com/sensu/sensu-puppet/pull/831 assumes that when you wrote:
client_register should be a hash, 
you were meaning:
client_registration should be a hash, 
and that actually there's not a client_register parameter (only client_registration).

If these assumptions, based on my understanding of the docs, are correct, https://github.com/sensu/sensu-puppet/pull/831  should be ok to merge.

ghoneycutt:
Merged #831 which solves this.

Issue #815, opened 2017-09-19
Cyclical dependencies when using Sensu Enterprise and the Enterprise API


ghoneycutt:
## Description of problem

- What did you do?

Brought up `sensu-server-enterprise` vagrant VM and applied the code below.

- What happened?

Got an error about a cyclical dependency

```
Error: Failed to apply catalog: Found 1 dependency cycle:
(Sensu_enterprise_dashboard_api_config[sensu-server.example.com] => Service[sensu-enterprise-dashboard] => Class[Sensu::Enterprise::Dashboard] => Sensu::Enterprise::Dashboard::Api[sensu-server.example.com] => Sensu_enterprise_dashboard_api_config[sensu-server.example.com])
```

- What did you expect to happen?

No error.

- How can someone reproduce the problem?

Apply this puppet code. Note that `FACTER_SE_USER` and `FACTER_SE_PASS` environment variables need to be populated.

```puppet
class { '::sensu':
  enterprise                   => true,
#  enterprise_version           => $::sensu::enterprise_version,
  enterprise_dashboard         => true,
#  enterprise_dashboard_version => $::sensu::enterprise_dashboard_version,
  enterprise_dashboard_host    => $::fqdn,
  enterprise_dashboard_port    => 3000,
  enterprise_dashboard_refresh => '5',
  enterprise_dashboard_ldap    => {
    'basedn'                     => "cn=users,dc=domain,dc=tld",
    'groupbasedn'                => "cn=groups,dc=domain,dc=tld",
    'userbasedn'                 => "cn=admins,dc=domain,dc=tld",
    'binduser'                   => "cn=binder,cn=users,dc=domain,dc=tld",
    'bindpass'                   => 'secret',
    'server'                     => 'localhost',
    'port'                       => 389,
    'roles'                      => [
      {
        'name'    => 'operators',
        'members' => ['group1','group2'],
      },
      {
        'name'     => 'guests',
        'members'  => ['employees','contractors'],
        'readonly' => true,
      }
    ],
    'security'            => 'starttls',
    'userattribute'       => 'sAMAccountName',
    'disablenestedgroups' => true,
  },
  enterprise_user  => $::se_user,
  enterprise_pass  => $::se_pass,
  install_repo     => true,
  log_level        => 'info',
  purge => {
    config => true,
  },
#  rabbitmq_cluster => $rabbitmq_cluster,
#  redis_db         => $redis_db,
#  redis_host       => $redis_host,
#  version          => $client_version,
[...]

PR #816, opened 2017-09-20
[815] Resolve circular dependency when using sensu::enterprise::dashboard::api

glarizza:
See issue #815 on Github for more details. Previously, when declaring the sensu
class along with an instance of the sensu::enterprise::dashboard::api defined
type, a circular dependency was raised by Puppet. This was caused because
first, the sensu::enterprise::dashboard::api defined type declares an instance
of the sensu_enterprise_dashboard_api_config type, and the
sensu_enterprise_dashboard_api_config type contains, within the initialize
method, contains an automatic "notify" on the "sensu-enterprise-dashboard" service (i.e.
Service['sensu-enterprise-dashboard']). The sensu::enterprise::dashboard::api
defined type also uses the require function on the sensu::enterprise::dashboard
class, which sets a 'require' dependency on every resource within the class.
Because Service['sensu-enterprise-dashboard'] is located within the
sensu::enterprise::dashboard class, that means that
Service['sensu-enterprise-dashboard'] contains a circular dependency on any
instance of the sensu::enterprise::dashboard::api defined type.

To resolve the circular dependency, this commit changes the require function
within the sensu::enterprise::dashboard::api defined type to
the include function which eliminates the 'require' dependency.

I tested this change out with vagrant and was able to get JSON back from the
API, but I'd prefer someone with greater Sensu Enterprise knowledge to validate
and ensure that converting 'require' to 'include' didn't upset any existing
dependencies.

### DIFF
--- a/manifests/enterprise/dashboard/api.pp
+++ b/manifests/enterprise/dashboard/api.pp
@@ -39,7 +39,7 @@ define sensu::enterprise::dashboard::api (
   Optional[String]  $pass          = undef,
 ) {
 
-  require ::sensu::enterprise::dashboard
+  include ::sensu::enterprise::dashboard
 
   sensu_enterprise_dashboard_api_config { $title:
     ensure     => $ensure,


ghoneycutt:
@agoddard Is getting JSON back from the API test enough or should we be testing something else?

ghoneycutt:
@agoddard ping

cwjohnston:
@ghoneycutt getting a HTTP 200 and JSON response body back from the API is sufficient to indicate that the API is up and available. Are you looking for a way to test that the API is online or that the Dashboard is online?

alvagante:
FYI, this is addressed also in https://github.com/sensu/sensu-puppet/pull/828

ghoneycutt:
Released in v2.36.0

Issue #542, opened 2016-08-23
upgrading to newer version of sensu with newer embeded ruby doesn't reinstall plugins


bchaja:
We are using embedded ruby when installing newer version of sensu with newer embedded ruby the plugins not being reinstalled making them unavailable under upgraded version (also we are using provider sensu_gem). 2 puppet run fixes the issue. All plugins are defined in hiera as:

sensu_gems:
   sensu-plugins-cpu-checks: {}

dmizelle:
We have also run into this issue, and have to solve it by running puppet again. Is there some way to perhaps create a subscribe parameter on each plugin to 'refresh'/reinstall after the version of sensu changes?

If we update sensu-client right now in our environment, all checks go critical/unknown, which makes us get paged.

jaxxstorm:
I don't use this myself, can you provide me some examples of how you're using this?

rowleyaj:
```
  # When updating sensu we found that the sensu_gem installed plugins would take
  # a second puppet run to install. This ensures that when the Sensu package
  # changes, all sensu_gem installed packages will be notified, and then notifies
  # the sensu services (if installed)
  Package['sensu'] ~>
    Package<| provider == 'sensu_gem' |> ~>
      Service['sensu-client', 'sensu-server', 'sensu-api']
```

We added the above to our profile to work around this. There is still a brief period of time where a check may run and error, but it is better than waiting for 30 minutes for the next puppet run.

These kind of dependencies could be captured in the module using the above, there may be others that can be resolved this way too, for example #463

glarizza:
Hopping on this ticket to replicate and see what I can update - is this still a problem for all involved with the recent module changes (presuming so but asking just to be sure).  Does anyone have any code they're using to replicate this issue (I'm presuming install a specific version with all plugins, specify a new version, and then observe)?

ghoneycutt:
@agoddard Any thoughts on how we can reproduce? Should the sensu package itself handle this so gems with the old version are attempted to be reinstalled?

glarizza:
As I understand it, plugins to Sensu's embedded ruby installation are installed via package resources with the `sensu_gem` provider included in this module:

```puppet
package { 'redphone':
  ensure   => 'installed',
  provider => sensu_gem,
}

package { 'sensu-plugins-nginx':
  ensure   => 'installed',
  provider => sensu_gem,
}

package { 'sensu-plugins-disk-checks':
  ensure   => 'installed',
[...]

PR #817, opened 2017-09-21
[542] Add dependencies for sensu_gem plugins

glarizza:
Previously, on upgrades to Sensu that installed a new embedded Ruby
installation, there was a chance that plugins installed with the `sensu_gem`
package provider would be evaluated BEFORE the Sensu package was upgraded
(which would mean that a second run of Puppet was necessary to get the plugins
installed into the new embedded Ruby installation). This commit adds a resource
chain ensuring that a change to the Sensu package will come before any package
resources using the `sensu_gem` provider (which would in turn notify the Sensu
services of their change).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -598,7 +598,7 @@ class sensu (
   # necessary to get the plugins installed into the new embedded Ruby folder.
   Package['sensu']
   ~> Package<| provider == 'sensu_gem' |>
-  ~> Service['sensu-client', 'sensu-server', 'sensu-api']
+  ~> Service<| title == 'sensu-client' or title == 'sensu-server' or title == 'sensu-api'|>
 
   if $plugins_dir {
     sensu::plugin { $plugins_dir: type => 'directory' }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -591,6 +591,15 @@ class sensu (
   -> Class['::sensu::server::service']
   -> Class['::sensu::enterprise::dashboard']
 
+  # Dependencies for Plugins #
+  # Without this line, plugins installed with the sensu_gem provider may get
+  # evaluated before the sensu package, which, on sensu package upgrades that
+  # install a new version of Ruby, would mean that two runs of Puppet would be
+  # necessary to get the plugins installed into the new embedded Ruby folder.
+  Package['sensu']
+  ~> Package<| provider == 'sensu_gem' |>
+  ~> Service['sensu-client', 'sensu-server', 'sensu-api']
+
   if $plugins_dir {
     sensu::plugin { $plugins_dir: type => 'directory' }
   } else {
--- a/tests/verify_plugin_installation_order.sh
+++ b/tests/verify_plugin_installation_order.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+
+# This script will verify that plugins installed with the "sensu_gem" package
+# provider are installed AFTER any sensu package changes.
+#
+# USAGE: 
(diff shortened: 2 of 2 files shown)


glarizza:
Tied to #542

manifests/init.pp
```diff
@@ -591,6 +591,15 @@
...
   -> Class['::sensu::enterprise::dashboard']
 
+  # Dependencies for Plugins #
+  # Without this line, plugins installed with the sensu_gem provider may get
+  # evaluated before the sensu package, which, on sensu package upgrades that
+  # install a new version of Ruby, would mean that two runs of Puppet would be
+  # necessary to get the plugins installed into the new embedded Ruby folder.
+  Package['sensu']
+  ~> Package<| provider == 'sensu_gem' |>
+  ~> Service['sensu-client', 'sensu-server', 'sensu-api']
```
ghoneycutt: Please test with the enterprise version as it uses different service names, we would need to take that into account.

glarizza: Good call, I'll test and most likely will ultimately add sensu-enterprise to the service list.  I think we're fine with only needing Package['sensu'] because I think the Sensu core package is the one that installs the embedded Ruby (not sure what Package['sensu-enterprise'] adds, but as long as it's something other than the Ruby implementation I don't care about it).

glarizza: Added in ebeeeef.....also lol at ebeeeef

glarizza:
Tests are failing on some rspec-puppet tests because of the last line:

Package['sensu'] ~>
Package<| provider == 'sensu_gem' |> ~>
Service['sensu-client', 'sensu-server', 'sensu-api']
On cases where the api or client hasn't been installed there won't be any resources for the sensu-api or sensu-client service. In its place I can do something like this:

~> Service<| title == 'sensu-client' or title == 'sensu-server' or title == 'sensu-api'|>
That collector will discover the resources if they're there and setup the dependencies, or exit silently if they're not in the catalog. It's not the prettiest, but it will do.

manifests/init.pp
```diff
@@ -591,6 +591,16 @@
...
 
+  # Dependencies for Plugins #
+  # Without this line, plugins installed with the sensu_gem provider may get
+  # evaluated before the sensu package, which, on sensu package upgrades that
+  # install a new version of Ruby, would mean that two runs of Puppet would be
+  # necessary to get the plugins installed into the new embedded Ruby folder.
+  Package['sensu']
+  ~> Package<| provider == 'sensu_gem' |>
+  ~> Service<| title == 'sensu-client' or title == 'sensu-server' or title == 'sensu-api'|>
+  #~> Service['sensu-client', 'sensu-server', 'sensu-api']
```
ghoneycutt: Still working through the enterprise side?

These will need the corresponding spec tests.

glarizza: I need to pull out that comment for sure.  As for the enterprise side, I'm waiting for @agoddard to reply to what services (if any) on the enterprise we will need to bump (see: https://github.com/sensu/sensu-puppet/issues/542#issuecomment-331249697).  I don't think we need to bounce that service, but I'll add it when I get a reply

ghoneycutt:
ping @agoddard

cwjohnston:
@ghoneycutt I think I'm missing the outstanding question here. Are we trying to determine how to handle the case where no Sensu service resources are defined?

ghoneycutt:
Fixes #542

ghoneycutt:
Released in v2.46.0

PR #818, opened 2017-09-25
add sensu version facter

bovy89:

### DIFF
--- a/lib/facter/sensu_version.rb
+++ b/lib/facter/sensu_version.rb
@@ -1,0 +1,16 @@
+Facter.add(:sensu_version) do
+  case Facter.value(:kernel)
+  when "windows" || "Windows"
+    setcode do
+      if File.exists? 'C:\opt\sensu\embedded\bin\sensu-client.bat'
+        sensuversion = Facter::Util::Resolution.exec('C:\opt\sensu\embedded\bin\sensu-client.bat --version 2>&1')
+      end
+    end
+  else
+    setcode do
+      if File.exists? '/opt/sensu/embedded/bin/sensu-client'
+        sensuversion = Facter::Util::Resolution.exec('/opt/sensu/embedded/bin/sensu-client --version 2>&1')
+      end
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


lib/facter/sensu_version.rb
```diff
@@ -0,0 +1,10 @@
+Facter.add(:sensu_version) do
+  confine :kernel => 'Linux'
+  setcode do
+    if File.exists? '/opt/sensu/embedded/bin/gem'
+      sensuversion = Facter::Util::Resolution.exec('/opt/sensu/embedded/bin/gem list "^sensu$"')
```
ghoneycutt: Suggest we use `/opt/sensu/bin/sensu-client --version`.

ghoneycutt:
Thank you for the contribution!

We need spec tests for this, as well as ensure it does not error when used on a system that does not have sensu installed.

spec/unit/facter/sensu_version_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    describe 'returns nil when sensu is not present' do
+        it do
+          File.stubs(:exists?).with('/opt/sensu/embedded/bin/sensu-client').returns(false)
+          expect(Facter.value(:sensu_version)).to be_nil
+        end
+    end
+  end
+
+  context "on non linux systems" do
```
ghoneycutt: I believe this should read non-posix systems as the test should work on Solaris, OSX and FreeBSD.

lib/facter/sensu_version.rb
```diff
@@ -0,0 +1,8 @@
+Facter.add(:sensu_version) do
+  confine :kernel => 'Linux'
```
ghoneycutt: This should not confine for Linux as it should work on the other posix platforms.

Perhaps a conditional such as
```
if Facter.value('kernel') == 'windows'
  <GH will figure out windows code>
else
  <posix code here>
end
```

ghoneycutt:
Thanks for adding the tests! Unless you're keen to add windows support, I'll handle that and then merge. Want to ensure that all of our platforms can use this great new fact.

ghoneycutt:
We should add `facter -p sensuversion` to the vagrant scripts. This will test that this works while also giving us useful information.

treydock:
@ghoneycutt Changes made in #894

ghoneycutt:
Released functionality in v2.52.0

PR #823, opened 2017-10-03
[WIP] Refactor to use DSC for installing the service on Windows

### LINKED ISSUE #819: Running Sensu on Windows as a specific account
I have a need to run Sensu on Windows with a different account to the default LocalService/LocalSystem that it uses by default. This would require registering the service with some different Powershell that looks something like this:
```ps1
$securePassword = convertto-securestring -String "password" -AsPlainText -Force  
$credentialsObject = new-object -typename System.Management.Automation.PSCredential -argumentlist "username", $securePassword
New-Service -Name sensu-client -Credential $credentialsObject -BinaryPathName c:\\opt\\sensu\\bin\\sensu-client.exe -DisplayName 'Sensu Client' -StartupType Automatic
```
This doesn't strike me as something very idempotent if one were to change the account details after the service has already been registered.

Would it make sense to instead use one of the `dsc_service` or `dsc_xservice` types from the DSC module which should a) hide the scary Powershell and b) hopefully be a bit more idempotent w.r.t. the account credentials?

bodgit:
## Description
* Add new parameter `$windows_service_user` which can be used to run the
  service as a different user on Windows.
* Use `dsc_service` resource instead of `exec` to install the service.
  If `$windows_service_user` isn't set then this will just use the Local
  System user as normal.
* If `$windows_service_user` is set then also use the
  `dsc_userrightsassignment` resource to grant the "Log on as a service"
  right to the user and also use the `acl` module to grant permission
  for the user to access the Sensu install directory.

## Related Issue
Fixes #819 .

## Motivation and Context
This allows you to run Sensu as a specific user account instead of the default Local System account, which in this case means it can use an AD account that has permission to monitor/read files on restricted SMB/CIFS shares.

## How Has This Been Tested?
So far I've tested this by running Puppet with and without setting `$windows_service_user` and confirming that the Sensu Client service is running.

In accordance with the DSC module, you do need to have the following resource set somewhere:
```puppet
reboot { 'dsc_reboot' :
  message => 'DSC has requested a reboot',
  when    => 'pending'
}
```
However this doesn't change doesn't trigger any reboots.

## General

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -62,6 +62,14 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">=4.16.0 <5.0.0"
+    },
+    {
+      "name": "puppetlabs/dsc",
+      "version_requirement": ">=1.4.0 <2.0.0"
+    },
+    {
+      "name": "puppetlabs/acl",
+      "version_requirement": ">=2.0.1 <3.0.0"
     }
   ]
 }
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -40,12 +40,35 @@ class sensu::client (
         content => template("${module_name}/sensu-client.erb"),
       }
 
-      exec { 'install-sensu-client':
-        provider => 'powershell',
-        command  => "New-Service -Name sensu-client -BinaryPathName c:\\opt\\sensu\\bin\\sensu-client.exe -DisplayName 'Sensu Client' -StartupType Automatic",
-        unless   => 'if (Get-Service sensu-client -ErrorAction SilentlyContinue) { exit 0 } else { exit 1 }',
-        before   => Service['sensu-client'],
-        require  => File['C:/opt/sensu/bin/sensu-client.xml'],
+      if $::sensu::windows_service_user {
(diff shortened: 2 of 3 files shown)


metadata.json
```diff
@@ -62,6 +62,14 @@
     {
       "name": "puppetlabs/stdlib",
       "version_requirement": ">=4.16.0 <5.0.0"
+    },
+    {
+      "name": "puppetlabs/dsc",
+      "version_requirement": ">=1.4.0 <2.0.0"
+    },
+    {
+      "name": "puppetlabs/acl",
```
alvagante: Even if required by this commit, I'd suggest to not add this as an hard dependency in metadata.json.
The (great) dsc module provides an hell of plugins that are synced to every client (both linux, where they are not used, and windows). I'd suggest to document the requirement for Windows nodes when the windows_service_user param is set. my2c

ghoneycutt: +1

alvagante:
I added tests in #873 and other corrections, this can be closed.

PR #824, opened 2017-10-03
fix

csolei2:
# Pull Request Checklist

Changing user and group from undef to use the user paramater

## Description
Changing user and group from undef to use the user paramater

## Related Issue
I was having an issue where the module was setting the user to a domain account because we had set it to that using a very old fork of a module. For some reason, puppet kept changing it back to that user if I'd change the ownership, so I'm defining the user from init.pp rather than leaving it as undef

Fixes # .

## Motivation and Context

I was having an issue where the module was setting the user to a domain account because we had set it to that using a very old fork of a module. For some reason, puppet kept changing it back to that user if I'd change the ownership, so I'm defining the user from init.pp rather than leaving it as undef. Also adding in the cyclical dependency fix that @glarizza put in.

## General

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -149,8 +149,8 @@ define sensu::check (
     'windows': {
       $etc_dir = 'C:/opt/sensu'
       $conf_dir = "${etc_dir}/conf.d"
-      $user = undef
-      $group = undef
+      $user = $::sensu::user
+      $group = $::sensu::group
       $file_mode = undef
     }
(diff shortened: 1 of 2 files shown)


ghoneycutt:
@csolei2 Apologies this one slipped through the cracks. Is it still an issue for you?

csoleimani:
I think this might have gotten merged when we added Mac support. I think we can close this

Issue #834, opened 2017-10-26
redact parameter causes errors


ghoneycutt:
## Description of problem

- What did you do?

Attempted to use redact by setting the redact parameter to the sensu class with the value `["password", "passwd", "pass", "api_key", "api_token", "access_key", "secret_key", "private_k
ey", "secret", "ec2_access_key", "ec2_secret_key"], `

- What happened?

Catalog compilation fails.

- What did you expect to happen?

For the catalog to compile and redact those fields.

- How can someone reproduce the problem?

Include the sensu class with the `redact` parameter set to the values above.

## Command used and debugging output

- What does your puppet config look like (including any hiera config)
- Is this a masterless or master based puppet setup?

alvagante:
So @ghoneycutt I've explored a bit the issue.
It looks like it's not related only to redact parameter but also to subscribers, and I suspect, any parameter on the sensu class which is used in sensu::client where an array (but not an hash) is expected as argument, and there's no default, even empty, set.

The referenced commit sets an empty array as default value for redact, and this creates in /etc/sensu/conf.d/client.json a line like:
```
{
[...]
    "redact": [

    ]
  }
}
```
This allows this entry to be recognised and changeable by sensu_client_config type.

So the commit solves the problem when a new configuration is created from scratch, but there's a caveat: if there's an existing /etc/sensu/conf.d/client.json but without any 'redact' (or 'subscriptions') entry we bumb again in the error which traces back to:
```
Error: undefined method `sort' for :absent:Symbol
lib/puppet/type/sensu_client_config.rb:47:in `insync?'
``` .

Probably @jeffmccune has better understanding on how the json provider for sensu_client_config works.
I've the feeling that parameters that are arrays are not correctly managed on existing files, if they are not already present in the json file.

alvagante:
Ok, after some digging,  https://github.com/alvagante/sensu-puppet/commit/9206f0f62705eed732f14c4c52acb753a66e4f3b seems to fix the problem. But since is mostly a copy and paste with limited introspection, I'd ask for @jeffmccune opinion.

ghoneycutt:
Released in v2.38.1

PR #835, opened 2017-10-30
Set a defauly empty array for redact #834

alvagante:
# Pull Request Checklist

## Related Issue

Fixes # .

## General


alvagante:
So, this commit is incomplete.
Besides the spec tests, due to the change of the default value, there'll would be errors on existing configurations.
Reason is deeper, to be looked in how json files are parsed and resource parameters added.

alvagante:
@ghoneycutt this should be fixed, can rebase if no other changes are needed, or maybe before remove the test file

ghoneycutt:
Thanks @alvagante 

Released in v2.38.1

PR #846, opened 2017-11-21
Fix #820

### LINKED ISSUE #820: Roundrobin subscriptions on Windows aren't configured
## Description of problem

- What did you do?

Applied the following on a Windows host:
```
::sensu::subscription { 'roundrobin:foo':
  ensure => present,
}
```
- What happened?

A zero-length file `C:\opt\sensu\conf.d\subscription_roundrobin` is created

The subscription should be created, with a suitable filename.

It looks like the subscription name should be sanitised as while the `:` character is fine in filenames on !Windows, it has special meaning as a drive letter separator which likely explains why the filename is truncated.

- Your OS: Windows 2012r2
- Your version of Puppet: 4.10.0
- Your version of Sensu: 1.0.2
- Your version of this module: 2.35.0

alvagante:
Should fix #820 
Have had to add a new parameter to client_config_register.

## Related Issue

Fixes #820  .

## General

### DIFF
--- a/lib/puppet/type/sensu_client_subscription.rb
+++ b/lib/puppet/type/sensu_client_subscription.rb
@@ -34,6 +34,11 @@ Puppet::Type.newtype(:sensu_client_subscription) do
     defaultto '/etc/sensu/conf.d/'
   end
 
+  newparam(:file_name) do
+    desc "The name of the client config file"
+    defaultto { "subscription_" + resource.name + ".json" }
+  end
+
   newproperty(:subscriptions, :array_matching => :all) do
     desc "Subscriptions included, defaults to resource name"
 
--- a/manifests/subscription.pp
+++ b/manifests/subscription.pp
@@ -13,7 +13,11 @@ define sensu::subscription (
 
   include ::sensu
 
-  file { "${::sensu::conf_dir}/subscription_${name}.json":
+  # Remove any from title any char which is not a letter, a number
+  # or the . and - chars. Needed for safe path names.
+  $sanitized_name=regsubst($name, '[^0-9A-Za-z.-]', '_', 'G')
+
+  file { "${::sensu::conf_dir}/subscription_${sanitized_name}.json":
     ensure => $ensure,
     owner  => $::sensu::user,
     group  => $::sensu::group,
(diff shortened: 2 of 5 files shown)


manifests/subscription.pp
```diff
@@ -13,7 +13,9 @@
 
   include ::sensu
 
-  file { "${::sensu::conf_dir}/subscription_${name}.json":
+  $sanitized_name=regsubst($name, '[^0-9A-Za-z.-]', '_', 'G')
```
ghoneycutt: Can you please comment on what this function is doing?

spec/defines/sensu_subscription_spec.rb
```diff
@@ -50,4 +50,23 @@ class {'sensu':
...
   end
 
+  context 'with char : in title' do
+    let(:title) { 'roundrobin:foo' }
+
+    it { should contain_sensu_client_subscription('roundrobin:foo').with(:ensure => 'present', :file_name => 'subscription_roundrobin_foo.json') }
+    it { should contain_file('/etc/sensu/conf.d/subscription_roundrobin_foo.json').with(:ensure => 'present' ) }
+  end  
+
+  context 'with char : in title in windows' do
```
ghoneycutt: Awesome! Love seeing those tests :)

lib/puppet/type/sensu_client_subscription.rb
```diff
@@ -34,6 +34,11 @@ def initialize(*args)
     defaultto '/etc/sensu/conf.d/'
   end
 
+  newparam(:file_name) do
```
ghoneycutt: This should be documented in the README

alvagante: Actually in README we document only sensu::subscription and from here we call sensu_client_subscription using the new file_name param (https://github.com/sensu/sensu-puppet/pull/846/commits/f8b25fe82584ebebab61461cd43e9803409ec531#diff-dbd03c1a865315be6893da851b471baeR29)  but not exposing it as param for sensu::subscription.
So either we expose file_name parameter also in sensu::subscription and we document it, or we just leave it a an internal setting that users don't have to mess it.

ghoneycutt: ah ok, it can stay internal

manifests/subscription.pp
ghoneycutt: regex explained in words can be really useful so you dont have to dissect what it is doing

ghoneycutt:
Released in v2.43.0

PR #845, opened 2017-11-18
Renamed references to puppetlabs to voxpupuli rabbitmq #844

alvagante:
# Pull Request Checklist
Renamed references from puppetlabs/rabbitmq to puppet/rabbitmq
The module is basically the same, just changed ownership.
Going to make some tests around it before removing the WIP:

## Related Issue

Fixes # .

## General

- [ ] Update `README.md` with any necessary configuration snippets

- [ ] New parameters are documented

- [ ] New parameters have tests

- [ ] Tests pass - `bundle exec rake validate lint spec`

PR #847, opened 2017-11-21
Revert "Merge pull request #845 from alvagante/844"

ghoneycutt:
This reverts commit c57c26823fb469af12ed9a2bbcb6083d4d21def7, reversing
changes made to de934d5a72f5b7105e12cc9ae860e2a1f2e5e2c0.


ghoneycutt:
The change to puppet/rabbitmq looks fine from a puppet run but leaves a non working system. This can be checked by looking at the API.

I use the following bash function to assist in this.

```shell
function sensuclients {
}
```

ghoneycutt:
@alvagante could you please work with the voxpupuli module to get it working for us so we can use it.

Issue #849, opened 2017-11-23
MAX_OPEN_FILES should be configurable for Sensu Enterprise


cwjohnston:
## Description of problem

Sensu Enterprise uses the value of MAX_OPEN_FILES environment variable to configure ulimit settings prior to starting the JVM. The value of this environment variable can be configured by editing /etc/default/sensu-enterprise, but is not present in [the template](https://github.com/sensu/sensu-puppet/blob/v2.40.1/templates/sensu.erb) used by this module to render that file. As a result, users cannot configure the number of file handles Sensu Enterprise can open when using this module.

## Platform and version information

- Your OS: Centos 7
- Your version of Sensu: Sensu Enterprise 2.7.0
- Your version of this module: 2.40.1

ghoneycutt:
Thanks @cwjohnston 

Could you let us know if there are any other options that should be configurable there? That way we can make this feature complete in one pass.

alvagante:
For what I see here:
https://sensuapp.org/docs/latest/enterprise/configuration.html
There are this Sensu Enterprise only environment variables:
HEAP_SIZE
HEAP_DUMP_PATH
JAVA_OPTS
MAX_OPEN_FILES
We currently support HEAP_SIZE and with the PR for this ticket, MAX_OPEN_FILES.

alvagante:
The variables common both to OSS and EE are, if I correctly understood:
SENSU_CLIENT_NAME
SENSU_CLIENT_ADDRESS
SENSU_CLIENT_SUBSCRIPTIONS
SENSU_TRANSPORT_NAME
RABBITMQ_URL
REDIS_URL
REDIS_SENTINEL_URLS
SENSU_API_PORT
In our /etc/default/sensu file, we we set other environment variables,  we do not manage any of them (they are manaed via the normal  configuration json files)

cwjohnston:
@alvagante You're right, both `HEAP_DUMP_PATH` and `JAVA_OPTS` are honored by Sensu Enterprise only, and are not currently managed by this module. It would be ideal to support configuring these via Puppet. In conjunction with #850, I believe we'd have all the Enterprise-specific environment variables covered at that point.

FWIW, Sensu Enterprise will source both /etc/default/sensu and /etc/default/sensu-enterprise on startup, so it's probably reasonable to isolate Enterprise-only variables into their own file.

cc @ghoneycutt

alvagante:
[...]

PR #850, opened 2017-11-25
Added support for MAX_OPEN_FILES environment variable #849

alvagante:
# Pull Request Checklist

## Related Issue

Fixes # .

## General

### DIFF
--- a/manifests/enterprise.pp
+++ b/manifests/enterprise.pp
@@ -30,6 +30,10 @@
 #
 # @param max_open_files Value of the MAX_OPEN_FILES environment variable.
 #
+# @param heap_dump_path Value of the HEAP_DUMP_PATH environment variable.
+#
+# @param java_opts Value of the JAVA_OPTS environment variable.
+#
 class sensu::enterprise (
   Optional[String]  $deregister_handler = $::sensu::deregister_handler,
   Optional[Boolean] $deregister_on_stop = $::sensu::deregister_on_stop,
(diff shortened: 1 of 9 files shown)


manifests/enterprise.pp
```diff
@@ -28,6 +28,8 @@
 #
 # @param heap_size Value of the HEAP_SIZE environment variable.
 #
+# @param max_open_files Value of the MAX_OPEN_FILES environment variable.
```
ghoneycutt: Why is the parameter added to three different classes?

alvagante: I followed the current approach for similar cases, but actually you're right, max_open_files in this case can and should not stay in package.pp. Actually if we split the sensu.erb template used both by /etc/default/sensu and /etc/default/sensu-enterprise we will be able to set ee-only parameters only in enterprise.pp

ghoneycutt: Do we need it anywhere except the enterprise class then?

alvagante: Actually, we need these params also in init.pp, besides enterprise.pp for how the module is done and tested (all spec tests for the various subclasses, in reality, test on the main class)

ghoneycutt: Do we need it in the package class then?

alvagante: Parameters used only in puppet enterprise are not needed in packages class.
FYI just pushed with the code to support other sense ee settings like JAVA_OPTS.
Have to properly test on vagrant yet. 
These changes anyway, might be obsoleted by what we decide in #851 .

alvagante:
@ghoneycutt this should be ready, and even if it might be refactored on the lines on what's under discussion in https://github.com/sensu/sensu-puppet/pull/852 I'd not wait for it.

Issue #851, opened 2017-11-28
CONFIG_FILE environment variable should be configurable


cwjohnston:
## Description of problem

The value of `CONFIG_FILE` environment variable is not currently configurable via this module.

Both Sensu Core and Sensu Enterprise use the value of the `CONFIG_FILE` environment variable to set the `--config` flag when starting Sensu processes. This flag specifies a singular JSON config file, the content of which is deep merged with the contents of the directory specified by the `--config_dir` flag. The `CONFD_DIR` environment variable used to set the value of `--config_dir` flag is [already configurable via this module](https://github.com/sensu/sensu-puppet/blob/master/templates/sensu.erb#L16-L18).

The default `CONFIG_FILE` value is `/etc/sensu/config.json` (e.g. https://github.com/sensu/sensu-omnibus/blob/master/files/sensu-gem/bin/sensu-service#L46-L55)

ghoneycutt:
I do not see `CONFIG_FILE` listed in #849 
Seems like that should be configurable in `/etc/default/sensu`, correct?

cwjohnston:
@ghoneycutt yes, `CONFIG_FILE` is honored by both Sensu Core and Sensu Enterprise, so I believe `/etc/default/sensu` is the right place.

alvagante:
@cwjohnston @ghoneycutt If you agree I'd add it with eventually all the others listed in #849.
Note that given the current code base this would require adding relevant parameters to the main class in init.pp, and at least packages.pp. There parameters would match environment variables like SENSU_CLIENT_SUBSCRIPTIONS, SENSU_TRANSPORT_NAME, RABBITMQ_URL, REDIS_URL which are already defined in other parameters.
In order to manage  CONFIG_FILE, the above environment variables, and anything that is used in /etc/default/sensu (and /etc/default/sensu-enterprise), I would introduce a single parameter, an hash of configuration settings, so that there will ne need in the future to introduce new variables to support new  settings for those files.
In order to preserve compatibilty with current parameters, I'd merge the user provided hash, in this new parameter (something like env_vars) with a default one, with settings based on the current parameters of the class.
I probably take more time describing than doing, anyway.
If you want I can prepare the code that implements this to give you and idea and then decide if to follow it. Scope is the configuration of /etc/default/sensu and /etc/default/sensu-enterprise and the relevant current and future parameters to manage them.

ghoneycutt:
I like this approach though it will require a major version change if we change parameters.

[...]

PR #852, opened 2017-11-29
WIP: Untested POC for #851

alvagante:
@cwjohnston  @ghoneycutt this is what I have in mind, I think that seeing the sample (not tested) code give you the idea of what I mean.
A similar thing could be done for the sensu::enterprise class and /etc/default/sensu-enterprise.

# Pull Request Checklist

## Related Issue

Fixes # .

## General


templates/sensu.erb
```diff
@@ -1,21 +1,4 @@
...
-CLIENT_DEREGISTER_HANDLER="<%= @deregister_handler %>"
-<% end -%>
-SERVICE_MAX_WAIT="<%= @init_stop_max_wait %>"
-PATH=<%= @path %>
-<% if @confd_dir -%>
-CONFD_DIR="<%= @conf_dir %>,<%= Array(@confd_dir).join(",") %>"
-<% end -%>
-<% if @heap_size -%>
-HEAP_SIZE="<%= @heap_size %>"
+# File managed by Puppet
```
alvagante: We may also mantain all the if conditions in the template, it would just require more maintenance, as contante would need to be changed whenever a new parameter is added

types/env_vars.pp
```diff
@@ -0,0 +1,14 @@
...
+  Optional[LOG_DIR] => Stdlib::Absolutepath,
+  Optional[RUBYOPT] => Stdlib::Absolutepath,
+  Optional[GEM_PATH] => String,
+  Optional[CLIENT_DEREGISTER_ON_STOP] => Boolean,
+  Optional[SERVICE_MAX_WAIT] => Stdlib::Absolutepath,
+  Optional[CLIENT_DEREGISTER_HANDLER] => Stdlib::Absolutepath,
+  Optional[PATH] => String,
+  Optional[CONFD_DIR] => String,
+
+}]
```
alvagante: Each time a new parameter has to be supported this would be the only place where to add something.

types/env_vars.pp
```diff
@@ -0,0 +1,14 @@
+type Sensu::Env_Vars = Struct[{
+
+  Optional[EMBEDDED_RUBY] => Boolean,
```
ghoneycutt: Wow, love the type validation here of the struct!

manifests/package.pp
```diff
@@ -144,6 +145,23 @@
...
+    'LOG_LEVEL'                 => $log_level,
+    'LOG_DIR'                   => $log_dir,
+    'RUBYOPT'                   => $rubyopt,
+    'GEM_PATH'                  => $gem_path,
+    'CLIENT_DEREGISTER_ON_STOP' => true,
+    'CLIENT_DEREGISTER_HANDLER' => $deregister_handler,
+    'SERVICE_MAX_WAIT'          => $init_stop_max_wait,
+    'PATH'                      => $path,
+    'CONFD_DIR'                 => "${conf_dir},${confd_dir}",
+  }
```
ghoneycutt: shouldn't heap size be here?

if we remove it, we would need need to bump the major version, which I'm not excited about doing since we'll be doing that soon enough for the next version of Sensu.

alvagante: heap_size afaik is used only on sensu enterprise, so it would be used in sensu::enterprise

types/env_vars.pp
alvagante: yeah, struct is the missing link to allow usage of single hashes to manage AND validate multiple configuration settings without multiplying classes' parameters.
Only limit is that currently, afaik, it' not possible to allow any possible key value, the ones accepted, even if optional, must be defined in the struct.

alvagante:
Closing this PR, too many fixes to do on tests which are not worth, now, the advantages of this refactoring. @ghoneycutt  Use https://github.com/sensu/sensu-puppet/pull/861 instead.

ghoneycutt:
@alvagante I like this approach to managing config files. Doing a similar thing with the facter module - https://github.com/ghoneycutt/puppet-module-facter/pull/59

PR #864, opened 2018-01-04
Support only the latest releases of Puppet versions 4 and 5

ghoneycutt:
# Pull Request Checklist

No longer supporting minor releases of Puppet 5. If you are having
issues, suggest updating Puppet to the newest release within your major
version.

## Description
Not testing against each minor release of puppet.

## Related Issue
None.

## Motivation and Context
Speed up testing and minimize the number of platforms that are supported.

## How Has This Been Tested?
TravisCI

## General


ghoneycutt:
Released in v2.43.0

PR #868, opened 2018-01-16
Adding options to configure dir and file modes + tests

### LINKED ISSUE #825: Allow for management of file and directory permissions
## Description of problem

- `$dir_mode` and `$file_mode` are both initialized, and set to their respective defaults depending on the OS [here](https://github.com/sensu/sensu-puppet/blob/master/manifests/init.pp#L535-L568).  This could be improved upon by having a configurable option within the class which defaults to the same values if not explicitly defined so it's backwards compatible.   This would allow users to explicitly manage their own dir and file permissions to line up with other types of management systems where applicable (i.e. rpm packages).

- N/A, but as an example of our workflow or use case, we have both team and enterprise packages that are installed on top of of the Sensu client package.  These additional packages install plugins and checks that run on each system.  Any time there is an update to these packages and Puppet runs, it overwrites the owner, group and mode on the files which results in a system change during every run which can be noisy at times.

- Your OS: CentOS 7
- Your Ruby version: Embedded, Ruby 2.4
- Your version of Puppet: 5
- Your version of Sensu: sensu-0.26.5-2.x86_64
- Your version of this module: 33e0ff373b4a29642412ac4761d5a540f14bb4ee, but this appears to be the same in the latest version as well, which we're trying to move to.

I'd be happy to submit a PR if this would be something that would be worthwhile.

jplindquist:
# Pull Request Checklist

Adds 2 new options to manage config directory and file permissions

## Description
Adds 2 new options (`config_dir_mode` and `config_file_mode`) to enable management of directory and file permissions.  Both options are _optional_ and default to their existing settings depending on OS (Windows / other).  If the new settings are undefined, there should be net zero changes to existing configs.

## Related Issue
Closes #825

## Motivation and Context
Allow users to explicitly manage their own directory and file permissions to line up with other types of management systems where applicable (i.e. rpm packages)

## How Has This Been Tested?
Ran spec tests before and after code changes, net zero changes unless config settings are present.  Added new tests to verify settings when options are set.

## General

- [X] Update `README.md` with any necessary configuration snippets

- [X] New parameters are documented

- [X] New parameters have tests

- [X] Tests pass - `bundle exec rake validate lint spec`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,6 +69,12 @@
 # @param sensu_group Name of the group Sensu is running as. Default is calculated
 #   according to the underlying OS
 #
+# @param config_dir_mode Directory mode for Sensu conf directory. Default is calculated
+#   according to the underlying OS
+#
+# @param config_file_mode File mode for config files under Sensu conf directory . Default is calculated
+#   according to the underlying OS
+#
 # @param rabbitmq_port Rabbitmq port to be used by sensu
 #
 # @param rabbitmq_host Host running rabbitmq for sensu
@@ -347,6 +353,14 @@ class sensu (
   Boolean            $manage_mutators_dir = true,
   Optional[String]   $sensu_user = undef,
   Optional[String]   $sensu_group = undef,
+  Optional[String]   $config_dir_mode = $::osfamily ? {
+    'windows' => undef,
+    default   => '0555',
+  },
+  Optional[String]   $config_file_mode = $::osfamily ? {
+    'windows' => undef,
+    default   => '0440',
+  },
   Variant[Undef,Integer,Pattern[/^(\d+)$/]] $rabbitmq_port = undef,
   Optional[String]   $rabbitmq_host = undef,
   Optional[String]   $rabbitmq_user = undef,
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -347,6 +353,14 @@
   Boolean            $manage_mutators_dir = true,
   Optional[String]   $sensu_user = undef,
   Optional[String]   $sensu_group = undef,
+  Optional[String]   $config_dir_mode = $::osfamily ? {
```
ghoneycutt: These should be `Optional[Stdlib::Filemode]`.

ghoneycutt:
PR looks great! Thank you!

I added Stdlib::Filemode not long ago and transitioned the module to use it, so these should use that new type.

https://github.com/puppetlabs/puppetlabs-stdlib/#stdlibfilemode

jplindquist:
Awesome, thanks for the quick feedback @ghoneycutt. Updated types to `Stdlib::Filemode`

ghoneycutt:
Thanks @jplindquist 

I'm closing this in favor of https://github.com/sensu/sensu-puppet/pull/869 where I just rebased your commits.

ghoneycutt:
Released in v2.49.0

PR #869, opened 2018-01-16
Add parameters to configure dir and file modes

### LINKED ISSUE #825: Allow for management of file and directory permissions

ghoneycutt:
This patch adds config_dir_mode and config_file_mode parameters and uses
the default values already present in the sensu class.

# Pull Request Checklist

Just a rebase of PR #868.

## Description

Fixes #825 
Fixes #868

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -69,6 +69,12 @@
 # @param sensu_group Name of the group Sensu is running as. Default is calculated
 #   according to the underlying OS
 #
+# @param config_dir_mode Directory mode for Sensu conf directory. Default is calculated
+#   according to the underlying OS
+#
+# @param config_file_mode File mode for config files under Sensu conf directory . Default is calculated
+#   according to the underlying OS
+#
 # @param rabbitmq_port Rabbitmq port to be used by sensu
 #
 # @param rabbitmq_host Host running rabbitmq for sensu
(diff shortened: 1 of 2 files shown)


ghoneycutt:
Released in v2.49.0  Thanks @jplindquist

PR #880, opened 2018-02-09
 enable specifying enable/ensure when managing services

dutsmiller:
## Description
This allows for changing the default setting for service enable/ensure when manage_service is set to true.

## Motivation and Context
We are using packer to build AMIs and have a need for the sensu client service to be set to:
    ensure => stopped
    enable => true
The reason for this is we do not want the packer instance to kick off the monitoring client while our AMI is being built, however we do want the service to start automatically whenever the AMI is eventually launched.


ghoneycutt:
Thanks for the PR!!

Could you please add spec tests for the new parameters?

dutsmiller:
Hmm, sadly, I can't say I've ever used rspec.  I'll have to take a look.

dutsmiller:
@ghoneycutt before I go any farther, is that what you're looking for?  (sorry, don't really know how to test rspec code).

spec/classes/sensu_client_spec.rb
```diff
@@ -232,14 +232,68 @@
...
+      end # with enable=false ensure=stopped hasrestart=true
+
+      context 'with enable=false ensure=stopped hasrestart=false' do
+        let(:params) { { :client => true, :client_service_enable => false, :client_service_ensure => 'stopped', :hasrestart => false } }
+        it { should contain_service('sensu-client').with(
+          :ensure     => 'stopped',
+          :enable     => false,
+          :hasrestart => false
+        ) }
+      end # with enable=false ensure=stopped hasrestart=false
```
ghoneycutt: this is great!

dutsmiller: I don't see any server service spec code to update, so is my work here done?

dutsmiller:
Just checking back in, any chance of cutting a new release?

ghoneycutt:
Solves #888

ghoneycutt:
Rebased in PR #889

ghoneycutt:
Released in v2.51.0

dutsmiller:
awesome, thanks for the merge!

PR #885, opened 2018-03-31
(GH-883) Remove example sensu_enterprise_dashboard_api_config resources by default

### LINKED ISSUE #883: sensu_enterprise_dashboard_api_config purge
v0.50.0

So following the README the following should work but still yields the 2 api1,api2 datacenters that come with install and get purged after a "second" puppet run:

```
  resources { 'sensu_enterprise_dashboard_api_config':
    purge   => true,
  }

  sensu::enterprise::dashboard::api { '127.0.0.1':
    ensure     => present,
    datacenter => 'My Datacenter',
  }
```

I even tried to make sure the purge happened after the Package install

```
  resources { 'sensu_enterprise_dashboard_api_config':
    purge   => true,
    require => Package['sensu'],
  }

  sensu::enterprise::dashboard::api { '127.0.0.1':
    ensure     => present,
    datacenter => 'My Datacenter',
  }
```

If I do another puppet apply run it works fine... but the first run doesn't delete those... 
I'm going to guess it has something to do with the resource purging has to happen after service is started and on first run it's happening before?
If that's the case then resource purge should have to subscribe to `Service['sensu']`? But even this won't work because the resources to purge aren't available on first run so they remove on second run only... any ideas?

```
  resources { 'sensu_enterprise_dashboard_api_config':
    purge     => true,
    subscribe => Service['sensu'],
  }

  sensu::enterprise::dashboard::api { '127.0.0.1':
    ensure     => present,
    datacenter => 'My Datacenter',
  }
```

I know there was some work previously done by @alvagante @glarizza @ghoneycutt so wasn't sure if this was related to that?

treydock:
# Pull Request Checklist

## Description
Remove `sensu_enterprise_dashboard_api_config` resources for `api1.example.com` and `api2.example.com` by default.

## Related Issue

Fixes #883 

## Motivation and Context

Better support resource purging for `sensu_enterprise_dashboard_api_config`

## How Has This Been Tested?

Unit tested

## General

### DIFF
--- a/manifests/enterprise/dashboard.pp
+++ b/manifests/enterprise/dashboard.pp
@@ -70,6 +70,16 @@ class sensu::enterprise::dashboard (
       oidc      => $::sensu::enterprise_dashboard_oidc,
       notify    => $file_notify,
     }
+
+    sensu_enterprise_dashboard_api_config { 'api1.example.com':
+      ensure => absent,
+      notify => $file_notify,
+    }
+
+    sensu_enterprise_dashboard_api_config { 'api2.example.com':
+      ensure => absent,
+      notify => $file_notify,
+    }
   }
 
   # Service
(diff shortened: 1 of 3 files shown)


treydock:
@ghoneycutt Worth adding the purge of `sensu_enterprise_dashboard_api_config` as optional boolean that is `false` by default?

ghoneycutt:
Released in v2.50.1

PR #887, opened 2018-04-02
(GH-884) Update acceptance testing

### LINKED ISSUE #884: Functional testing with serverspec
We need functional testing of the code with serverspec. The first iteration of this should support the linux platforms from https://github.com/sensu/sensu-puppet/blob/master/metadata.json and run under docker.

treydock:
Add travis-ci tests for all supported versions of Linux that use docker nodesets
Default PUPPET_GEM_VERSION to be Puppet 5 to avoid errors when not defined in environment

# Pull Request Checklist

## Description
Add automated acceptance tests.

## Related Issue

Fixes #884 

## Motivation and Context
Improved module testing by adding automated system acceptance tests.

## How Has This Been Tested?
Tested all supported Linux systems via travis-ci acceptance tests

## General


spec/acceptance/nodesets/amazonlinux-201703-docker.yml
```diff
@@ -0,0 +1,16 @@
...
+    roles:
+      - agent
+    platform: el-6-x86_64
+    hypervisor: docker
+    image: amazonlinux:2017.03
+    docker_preserve_image: true
+    docker_cmd:
+      - '/sbin/init'
+    docker_image_commands:
+      - 'yum install -y tar wget cronie git upstart'
```
ghoneycutt: How are these dependencies determined?

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
```diff
@@ -0,0 +1,16 @@
+HOSTS:
+  amazonlinux-201703:
+    roles:
+      - agent
+    platform: el-6-x86_64
+    hypervisor: docker
+    image: amazonlinux:2017.03
+    docker_preserve_image: true
```
ghoneycutt: what does this do?

spec/acceptance/nodesets/centos-6.yml
```diff
@@ -0,0 +1,10 @@
+HOSTS:
```
ghoneycutt: Why a vagrant node if we have docker?

spec/acceptance/nodesets/default.yml
```diff
@@ -1,10 +1,9 @@
 HOSTS:
-  centos-70-x64:
+  centos-7:
     roles:
-      - master
+      - agent
     platform: el-7-x86_64
-    box : puppetlabs/centos-7.0-64-nocm
-    box_url : https://vagrantcloud.com/puppetlabs/boxes/centos-7.0-64-nocm
+    box : centos/7
```
ghoneycutt: How is the default invoked? Why is it vagrant?

spec/acceptance/nodesets/ubuntu-1404-docker.yml
```diff
@@ -0,0 +1,17 @@
...
+    image: ubuntu:14.04
+    docker_preserve_image: true
+    docker_cmd: '["/sbin/init"]'
+    docker_image_commands:
+      - 'rm /usr/sbin/policy-rc.d'
+      - 'rm /sbin/initctl; dpkg-divert --rename --remove /sbin/initctl'
+      - 'apt-get install -y -q net-tools wget'
+      - 'locale-gen en_US.UTF-8'
+CONFIG:
+  type: aio
```
ghoneycutt: Why are some types `aio` and some are `foss`? What is the difference?

spec/acceptance/nodesets/ubuntu-server-1204.yml
```diff
@@ -1,7 +1,7 @@
 HOSTS:
-  ubuntu-server-1204-x64:
```
ghoneycutt: We don't support ubuntu 12. I don't think this should be here.

spec/acceptance/nodesets/ubuntu-server-1404.yml
```diff
@@ -1,7 +1,7 @@
 HOSTS:
-  ubuntu-server-1404-x64:
```
ghoneycutt: Why do we have Ubuntu and Ubuntu Server? I don't get the distinction.

.travis.yml
```diff
@@ -24,6 +24,54 @@ matrix:
     env: PUPPET_GEM_VERSION="~> 4"
   - rvm: 2.4.1
     env: PUPPET_GEM_VERSION="~> 5"
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="centos-6-docker"
+    bundler_args:
+    script: sudo service docker restart ; sleep 10 ; bundle exec rake beaker
```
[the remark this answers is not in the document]
treydock: Honestly never thought about it, just something I have been doing since while ago when I started doing acceptance tests.  May be unnecessary.

Gemfile
```diff
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-if puppetversion = ENV['PUPPET_GEM_VERSION']
+if puppetversion = ENV['PUPPET_GEM_VERSION'] || "~> 5.x"
```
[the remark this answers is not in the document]
treydock: The Gemfile errors out if `PUPPET_GEM_VERSION` is not set because `puppetversion` ends up being nil.  Without this change `PUPPET_GEM_VERSION` would have to be defined to the acceptance tests even though it has not bearing on the version of Puppet used by the tests.

```
$ bundle install
The latest bundler is 1.16.1, but you are currently running 1.15.4.
To update, run `gem install bundler`

[!] There was an error parsing `Gemfile`: undefined method `<' for nil:NilClass. Bundler cannot continue.

 #  from /Users/tdockendorf/git/sensu-puppet/Gemfile:37
 #  -------------------------------------------
 #    gem 'puppet-lint-version_comparison-check',             :require => false
 >    if puppetversion < '5.0'
 #      gem 'semantic_puppet', :require => false
 #  -------------------------------------------
```

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
treydock: They are things you may need to bootstrap test environment.  I always do git incase git cloning modules for testing.  For something like `upstart` is needed to support `docker_cmd` of `/sbin/init` which is necessary to have functioning container for tests.  Others may be unnecessary but always added as sometimes useful.  `cronie` likely not needed.

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
treydock: Allows container to be re-used between tests, useful for testing changes on local systems.

https://github.com/puppetlabs/beaker-docker/blob/master/docker.md#preserve-docker-image

These YAML may need to be updated to also allow `BEAKER_provision=no` which has options likely needed with beaker 3.x to fully support re-using containers.

https://github.com/puppetlabs/beaker-docker/blob/master/docker.md#reuse-docker-image

I'll update all images as really handy when doing local development to be able to re-run tests.  There is a bug though where executions after first will put the module in /etc/puppet/modules for some reason which is not used by AIO Puppet.  Never got around to opening an issue with Puppetlabs.

spec/acceptance/nodesets/ubuntu-1404-docker.yml
treydock: Should probably be `foss`.  I pulled the ubuntu files from one of my other puppet modules and the files were actually a contribution.  I know the determination of what is installed is actually not in beaker and not from nodeset file, not sure if `type` even used but it is here: https://github.com/puppetlabs/beaker/blob/master/acceptance/fixtures/module/spec/acceptance/nodesets/default.yml

Real place Puppet install type is defined: https://github.com/puppetlabs/beaker-puppet_install_helper#run_puppet_install_helper

Probably just need to make them all `foss`.

spec/acceptance/nodesets/ubuntu-server-1404.yml
treydock: I think the `server` was left overs from previous vagrant boxes.  I don't know of a difference, the containers are all just `ubuntu`, no server distinction.

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
treydock: FYI - reusable docker images works as documented but annoying bug of module being installed to wrong place is still present:

```
centos-7 20:07:02$ echo /etc/puppet/modules
  /etc/puppet/modules

centos-7 executed in 0.12 seconds
Using scp to transfer /home/treydock/git/sensu-puppet to /etc/puppet/modules/sensu
```

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
treydock: Further investigation, I believe cronie may have been added to support potentially adding cron jobs via Puppet in other modules and being able to test cron with serverspec: http://serverspec.org/resource_types.html#cron.  Not 100% certain if serverspec uses crontab or looks in /var/spool/cron type situation.

spec/acceptance/nodesets/amazonlinux-201703-docker.yml
ghoneycutt: Could you please report this bug to Puppet.

spec/acceptance/nodesets/centos-6.yml
ghoneycutt: If we are not testing it with TravisCI, this seems like cruft that could be removed. What do you think?

spec/acceptance/nodesets/default.yml
ghoneycutt: Is a default really needed? If not, let's remove it, else let's make it docker.

spec/acceptance/nodesets/ubuntu-1404-docker.yml
ghoneycutt: Please make them all `foss` if that's what they should be.

spec/acceptance/nodesets/ubuntu-server-1204.yml
ghoneycutt: Please remove any platforms we do not support.

spec/acceptance/nodesets/ubuntu-server-1404.yml
ghoneycutt: Please remove all the cruft we don't need.

.travis.yml
```diff
@@ -24,6 +24,54 @@ matrix:
     env: PUPPET_GEM_VERSION="~> 4"
   - rvm: 2.4.1
```
ghoneycutt: Seems we should be using ruby 2.4.3

https://puppet.com/docs/puppet/5.3/about_agent.html

spec/acceptance/nodesets/centos-6.yml
treydock: This is done, all nodesets are docker now.

.travis.yml
treydock: I've updated PR to use 2.4.3.

.travis.yml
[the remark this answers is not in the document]
ghoneycutt: Please test and see if we need this.

.travis.yml
```diff
@@ -24,6 +24,54 @@ matrix:
...
     env: PUPPET_GEM_VERSION="~> 5"
+  - rvm: 2.4.2
```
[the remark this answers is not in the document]
ghoneycutt: Will 2.4.3 work? That's what should also be used for Puppet 5. Could you please change those.

.travis.yml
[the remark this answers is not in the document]
treydock: I removed the service restarts and tests still work so was unnecessary.

.travis.yml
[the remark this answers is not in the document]
treydock: We could, but the version used in travis for acceptance tests isn't actually what runs inside the container so no real impact on puppet compatibility.  The only ruby version restriction for the acceptance tests is being supported by beaker and the various beaker support libraries.

I think the default Puppet version used in the tests inside the various containers is puppet-agent 1.10.0 which is Puppet 4.x.  Should we force latest Puppet or stick with default from beaker-puppet_install_helper?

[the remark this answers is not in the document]
treydock: Further analysis seems the various helpers to install Puppet with beaker don't yet support Puppet 5, at least not obvious.  https://github.com/puppetlabs/beaker-puppet_install_helper/issues/33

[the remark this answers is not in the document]
treydock: I'll see what's involved getting Puppet 5 used.  May be we just can't use the helper methods and have to explicitly bootstrap puppet.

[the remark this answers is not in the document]
treydock: Got Puppet 5 as the version installed for acceptance tests.  `PUPPET_INSTALL_VERSION` sets the default, leaving undefined results in latest `puppet-agent` present in `puppet5` repo.

spec/spec_helper_acceptance.rb
```diff
@@ -1,44 +1,29 @@
...
-        # RubyGems missing on some Ubuntu 12 boxes
-        # Otherwise you'lll get a load of 'Provider gem is not functional on this host'
-        shell('apt-get install rubygems -y')
-      end
       if fact('osfamily') == 'RedHat'
         # CentOS has epel-release package in Extras, enabled by default
         shell('yum -y install epel-release')
-        shell('yum -y install erlang')
       end
-      on host, puppet('module', 'install', 'puppetlabs-stdlib'), { :acceptable_exit_codes => [0,1] }
```
ghoneycutt: why does stdlib go away here?

treydock: Pulled in from `metadata.json` by `install_module_dependencies_on(hosts)` at the top of file.  Removed to avoid having to update some dependencies in two places.

ghoneycutt:
Released in v2.53.0

PR #897, opened 2018-04-26
(GH-896) Support beaker testing on windows server 2012r2

ghoneycutt:
# Pull Request Checklist

This allows beaker to connect, though the tests fail since they are attempting to run the linux command `cat`.

Fixes #896

## General

- [ ] Tests pass

Issue #896, opened 2018-04-26
beaker testing should support windows


ghoneycutt:
## Description of problem

Beaker supports linux platforms and should also support Windows. I think this will only work with Vagrant and not docker.

ghoneycutt:
Started in https://github.com/sensu/sensu-puppet/pull/897 though that just allows beaker to connect. Still getting errors.

ghoneycutt:
Finished in #898

ghoneycutt:
Released in v2.53.0

PR #898, opened 2018-04-28
(GH-896) Support beaker testing on windows server 2012r2 (continued)

treydock:
This is a continuation of #897 

```
$ bundle exec rake beaker:windows-server-2012r2
<SNIP>
Finished in 2 minutes 42 seconds (files took 5 minutes 44 seconds to load)
13 examples, 0 failures, 7 pending
```

Could probably use `unless` with contexts that should skip for windows but wasn't sure if better to have output explaining why some tests were skipped or just skip silently.


ghoneycutt:
Released in v2.53.0

Issue #901, opened 2018-05-03
v3 - Support sensu-go


ghoneycutt:
What's needed to support sensu-go

- [x] document development procedures for maintaining support for sensu-go and sensu v1.
- [x] document config approach - backend vs local config
- [x] base module to manage files, packages, services, etc.
- [x] documentation with puppet-strings
- [x] README
- [x] extension provider code is cleaned up after https://github.com/sensu/sensu-go/issues/1767 is fixed
- [x] support resource purging - https://github.com/sensu/sensu-puppet/pull/954
- [ ] types and providers
  - [x] asset
    - [ ] `ensure => absent` works - https://github.com/sensu/sensu-go/issues/988
  - [x] check
  - [x] cluster
  - [x] config
  - [x] entity
  - [x] environment
  - [x] event
  - [x] extension
  - [x] filter
  - [x] handler
  - [x] hook
  - [x] mutator
  - [x] organization
  - [x] role
  - [x] silenced
  - [ ] user
    - [ ] password support works - https://github.com/sensu/sensu-go/issues/1437
  - [x] api_validator
- [x] major platforms
  - [x] EL7
  - [x] Debian 9
  - [x] Ubuntu 16.04
- [ ] minor platforms
  - [x] EL 6
  - [x] Debian 7 is removed - https://github.com/sensu/sensu-puppet/issues/904
  - [x] Debian 8
  - [x] Debian 9
  - [x] Ubuntu 16.04
  - [x] Ubuntu 18.04
  - [x] Amazon 2017.3
  - [x] Amazon 2018.3 (travis only not vagrant)
  - [x] Amazon 2

For an item to be checked, it needs the following if applicable.
* spec tests
* acceptance tests (beaker/serverspec)
* Vagrant instance for each platform

treydock:
@ghoneycutt Where is environment documented?  Also may not need to do anything for `event`, looks like entirely informational and nothing that can be changed with `sensuctl`.

treydock:
@ghoneycutt Nevermind about environment, found under RBAC section.

treydock:
@ghoneycutt Any references for `extension`?  Also do we want to add a type for `event`?  `event` seems like something sent by checks or on a more ad-hoc basis.

ghoneycutt:
I think extension is just for v1 of sensu as it is not listed in the v2 docs

ghoneycutt:
@treydock Event's can be deleted and resolved, so we should have a type/provider for them.

[...]

PR #920, opened 2018-06-24
Add sensu_event type

treydock:
# Pull Request Checklist

## Related Issue

Part of #901  .

## Motivation and Context

Allow sensu-go events to be resolved and deleted.

## How Has This Been Tested?

Beaker acceptance tests and vagrant instance.  Travis-ci tests will fail until #916 is merged.

## General

### DIFF
--- a/lib/puppet/provider/sensu_event/sensuctl.rb
+++ b/lib/puppet/provider/sensu_event/sensuctl.rb
@@ -1,0 +1,83 @@
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'sensuctl'))
+
+Puppet::Type.type(:sensu_event).provide(:sensuctl, :parent => Puppet::Provider::Sensuctl) do
+  desc "Provider sensu_event using sensuctl"
+
+  mk_resource_methods
+
+  def self.instances
+    events = []
+
+    output = sensuctl_list('event')
+    Puppet.debug("sensu event: #{output}")
+    begin
+      data = JSON.parse(output)
+    rescue JSON::ParserError => e
+      Puppet.debug('Unable to parse output from sensuctl event list')
+      data = []
+    end
+
+    data.each do |d|
+      event = {}
+      event[:ensure] = :present
+      event[:entity] = d['entity']['id']
+      event[:check] = d['check']['name']
+      event[:name] = "#{event[:check]} for #{event[:entity]}"
+      if d['check']['status'] == 0
+        event[:ensure] = :resolve
+      end
+      events << new(event)
+    end
+    events
+  end
+
+  def self.prefetch(resources)
+    events = instances
+    resources.keys.each do |name|
+      if provider = events.find { |e|
+          e.check == resources[name][:check] &&
+          e.entity == resources[name][:entity]
+          }
+        resources[name].provider = provider
+      end
+    end
+  end
+
+  def exists?
+    @property_hash[:ensure] == :present || @property_hash[:ensure] == :resolve
+  end
+
+  def state
+    return @property_hash[:ensure]
+  end
+
+  def resolve
+    cmd = ['event', 'resolve']
+    cmd << resource[:entity]
+    cmd << resource[:check]
+    if resource[:organization]
+      cmd << '--organization'
+      cmd << resource[:organization]
+    end
+    if resource[:environment]
+      cmd << '--environment'
+      cmd << resource[:environment]
+    end
+    begin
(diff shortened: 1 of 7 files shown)


lib/puppet/provider/sensu_event/sensuctl.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  def exists?
+    @property_hash[:ensure] == :present || @property_hash[:ensure] == :resolve
+  end
+
+  def state
+    return @property_hash[:ensure]
+=begin
```
ghoneycutt: what's this begin/end bit about?

spec/unit/provider/sensu_event/sensuctl_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it 'should create a event' do
+      expect(@resource.provider).to receive(:sensuctl).with(['event', 'resolve', 'test', 'keepalive', '--organization', 'default', '--environment', 'default'])
+      @resource.provider.resolve
+      property_hash = @resource.provider.instance_variable_get("@property_hash")
+      expect(property_hash[:ensure]).to eq(:resolve)
+    end
+  end
+
+  describe 'destroy' do
+#    before(:each) do
```
ghoneycutt: why the commented out code in the commit?

spec/unit/provider/sensu_event/sensuctl_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    it 'should return the resource for a event' do
+      allow(@provider).to receive(:sensuctl_list).with('event').and_return(my_fixture_read('event_list.json'))
+      property_hash = @provider.instances[0].instance_variable_get("@property_hash")
+      expect(property_hash[:name]).to eq('keepalive for sensu-backend.example.com')
+      expect(property_hash[:entity]).to eq('sensu-backend.example.com')
+      expect(property_hash[:check]).to eq('keepalive')
+    end
+  end
+
+=begin
```
ghoneycutt: What's this?

spec/unit/sensu_event_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+      end
+    end
+  end
+
+  # String regex validated properties
+  [
+  ].each do |property|
+    it "should not accept invalid #{property}" do
+      config[property] = 'foo bar'
+      expect { event }.to raise_error(Puppet::Error)
```
ghoneycutt: This should probably have an error message that is expected

ghoneycutt:
@treydock still getting errors after #916 was merged

lib/puppet/provider/sensu_event/sensuctl.rb
treydock: These are block comments I forgot to remove after testing.

spec/unit/provider/sensu_event/sensuctl_spec.rb
treydock: Commented out from file this was copied from, forgot to cleanup, will remove.

spec/unit/provider/sensu_event/sensuctl_spec.rb
treydock: Block comment, I'll remove.

spec/unit/sensu_event_spec.rb
treydock: Will update.

treydock: Actually this block is empty, so the test never actually gets executed.  I left these empty test blocks around in case properties got added that fit the block's test criteria.  There are several places in code these blocks are actually tested, I'll update those in separate pull request.

ghoneycutt:
Having issues with acceptance tests

treydock:
I thought that would be fixed by latest commit but it wasn't, will investigate.  Thus far I haven't been able to reproduce locally using beaker and centos-7 set.  The testing of events is rather tricky since events can't be created for the purpose of testing, at least not without first creating some checks and having them run and then forcing one to fail, but I fear all that effort would add a lot of time to the acceptance tests which some are already timing out.

ghoneycutt:
Agreed that we should not do that extra effort which will lead to the tests timing out.

treydock:
Issue resolved by forcing agent node to have specific entity ID and not relying on local hostname lookup as seems some containers kept getting `localhost.localdomain` and this collided with backend container.

PR #927, opened 2018-06-27
add field ttl_status to sensu_check

jasiek191:
# Pull Request Checklist

## Related Issue

Fixes # .

## General


ghoneycutt:
Thanks @jasiek191 for the contribution! Current the spec tests are not passing, could you have a look.

jasiek191:
Hi
I didn't saw it, but now is fixed

spec/defines/sensu_check_spec.rb
```diff
@@ -105,7 +105,8 @@
         :handle              => true,
         :publish             => true,
         :auto_resolve        => true,
-        :ttl                 => 30
+        :ttl                 => 30,
+        :ttl_status          => 1
```
ghoneycutt: what happens if ttl_status is a non integer and not `'absent'` ?

jasiek191: if is set to absent then have default value 1 (which mean warning)
I don't know what happens if will be set to string, probably sensu client die.
Probably we should check also ttl, interval refresh etc

ghoneycutt: I meant, what will Puppet do? We want some checks to ensure that Puppet errors appropriately.

ghoneycutt:
@treydock could you please look into this. It seems it just needs tests before it can be merged.

ghoneycutt:
merged in PR #1000

ghoneycutt:
Thanks @jasiek191 and @treydock !! This has been released in v2.57.0

PR #929, opened 2018-06-29
Bundle RBAC related acceptance tests to improve runtime

treydock:
# Pull Request Checklist

## Description

Put all RBAC related acceptance tests together.

## Related Issue

Relates to #901 

## Motivation and Context

Some acceptance tests are timing out and appears to be large number of executions of `puppet apply`.  This change bundles all RBAC acceptance tests into same test blocks to hopefully resolve acceptance test timeouts in travis-ci.

## How Has This Been Tested?

`time BEAKER_set=debian-8 bundle exec rake beaker` which completed in 29 minutes.

## General


spec/acceptance/sensu_rbac_spec.rb
```diff
@@ -1,11 +1,17 @@
 require 'spec_helper_acceptance'
```
ghoneycutt: Could you please add a comment block here that explains why we have sensu_rbac_spec.rb?

spec/acceptance/sensu_environment_spec.rb
```diff
@@ -1,72 +0,0 @@
-require 'spec_helper_acceptance'
```
ghoneycutt: Since the filenames follow a pattern following the types, perhaps this should be a file that is just a comment that lists the issue number and lets you know the tests are in `spec/acceptance/sensu_rbac_spec.rb`. What do you think?

ghoneycutt: I'm worried it will look like tests are missing and it will not be clear if these types are modified where the tests are.

treydock: I think empty files should be fine, with a comment pointing back to sensu_rbac_spec.rb.

treydock:
@ghoneycutt Added comments to previously removed spec files and comment to new rbac spec file. Let me know if they are good enough or need a bit more information.

ghoneycutt:
looks good!

treydock:
@ghoneycutt This should not be merged.  I'll open a new PR that rules the type/provider tests only on EL7 and run agent/backend tests on all platforms.

treydock:
Closing in favor of #935

PR #936, opened 2018-07-07
Update README to mention soft dependency on puppetlabs/apt

treydock:
Need to mention module has soft dependency on `puppetlabs/apt` since not in `metadata.json`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ This module will install packages, create configuration and start services neces
 
 Plugin sync is required if the custom sensu types and providers are used.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,8 @@ This module will install packages, create configuration and start services neces
 
 Plugin sync is required if the custom sensu types and providers are used.
 
+This module has a soft dependency on the [puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) module for systems using `apt`.
```
ghoneycutt: Can you specify the version string like you would with `metadata.json`. I think it is `>= 4.0.0 > 5.0.0`

PR #937, opened 2018-07-07
sensu2 - Support Amazon Linux 2.0, 2017.03 and 2018.03

treydock:
# Pull Request Checklist

## Description

Add support for several Amazon Linux versions

## Related Issue

Relates to #901 

## Motivation and Context

Should support Amazon's latest versions.

## How Has This Been Tested?

In both Vagrant and Docker.

## General

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,24 @@ matrix:
     env: BEAKER_set="ubuntu-1804"
     bundler_args:
     script: bundle exec rake beaker
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="amazonlinux-2"
+    bundler_args:
+    script: bundle exec rake beaker
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="amazonlinux-201803"
+    bundler_args:
+    script: bundle exec rake beaker
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="amazonlinux-201703"
+    bundler_args:
+    script: bundle exec rake beaker
 
 notifications:
   email: false
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -3,9 +3,18 @@ class sensu::repo {
 
   case $facts['os']['family'] {
     'RedHat': {
+      if $facts['os']['name'] == 'Amazon' {
+        if $facts['os']['release']['major'] in ['2017','2018'] {
+          $repo_release = '6'
+        } else {
+          $repo_release = '7'
+        }
+      } else {
+        $repo_release = $facts['os']['release']['major']
+      }
       # TODO: change from nightly to stable once there are stable releases
       yumrepo { 'sensu_nightly':
-        baseurl         => "https://packagecloud.io/sensu/nightly/el/${facts['os']['release']['major']}/\$basearch",
+        baseurl         => "https://packagecloud.io/sensu/nightly/el/${repo_release}/\$basearch",
         repo_gpgcheck   => 1,
         gpgcheck        => 0,
         enabled         => 1,
--- a/spec/classes/repo_spec.rb
+++ b/spec/classes/repo_spec.rb
@@ -4,10 +4,18 @@ describe 'sensu::repo', :type => :class do
   on_supported_os({facterversion: '3.8.0'}).each do |os, facts|
     context "on #{os}" do
       let(:facts) { facts }
+      case os
+      when /(redhat-6|centos-6|amazon-2017|amazon-2018)-x86_64/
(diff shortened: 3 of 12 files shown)


treydock:
There is something wrong with Facter on Amazon 2.0 that causes the OS version to be kernel version and not 2.0.  This makes it difficult to use EL7 repo for just that version and EL6 repo for all others.  I found this https://tickets.puppetlabs.com/browse/FACT-1485, but says already fixed.  I think I can work around this issue with some reversed logic on fact checking for versions that do work and else falling back for version that doesn't work.

treydock:
This PR will need to be squashed before merge.  I discovered this in Vagrantbox:

```
[root@amazon2-agent ~]# cat /etc/system-release
Amazon Linux release 2.0 (2017.12) LTS Release Candidate
```

This reports as Amazon 2 via Facter and appears to be accessible via `amazonlinux:2017.12` docker image.  This version supports systemd and only mention I have found of Amazon Linux with systemd is Amazon Linux 2.

treydock:
Opened https://tickets.puppetlabs.com/browse/FACT-1867 for fact `amazonlinux:2` docker image is unable to report OS version using Facter.

ghoneycutt:
That facter ticket has been fixed.

spec/classes/repo_spec.rb
```diff
@@ -6,10 +6,10 @@
       let(:facts) { facts }
       if facts[:osfamily] == 'RedHat'
```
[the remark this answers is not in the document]
treydock: @ghoneycutt Amazon Linux 2 is systemd based so needs the EL7 repos.  The fact sensu-go bootstrap only does EL6 is a bug in my opinion and opened here: https://github.com/sensu/sensu-go/issues/1791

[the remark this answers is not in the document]
treydock: @ghoneycutt Uses big blocks that define facts could conflict and possibly break rspec-puppet-facts which looks up fact values based on what's supported in metadata.json.  Also mimics what values would be output by facter based on version so we're testing against Puppet 5 facts.  We could use the `os` value to build a mapping but that would get very verbose:

```
case os
when 'redhat-7-x86_64'
  something = something
when 'amazonlinux-2018-x86_64'
  something = something
else
  something = something
```

This would still rely on data from rspec-puppet-facts.

manifests/repo.pp
```diff
@@ -3,9 +3,18 @@
 
   case $facts['os']['family'] {
     'RedHat': {
+      if $facts['os']['name'] == 'Amazon' {
+        if $facts['os']['release']['major'] in ['2017','2018'] {
+          $repo_release = '6'
+        } else {
+          $repo_release = '7'
```
ghoneycutt: does `repo_release = 7` ever get picked then if we match against 2017 and 2018?

ghoneycutt: Or did amazon linux move from 2017, 2018 to 2.0 ?

tests/provision_amazon.sh
```diff
@@ -8,7 +8,7 @@ function rpm_install() {
     rm -f $package
 }
 
-release=$(awk -F \: '{print $5}' /etc/system-release-cpe)
+release=6
```
ghoneycutt: Wouldnt this use EL6 repos on Amazon Linux 2 which should use EL7 repos?

.travis.yml
```diff
@@ -64,6 +64,24 @@ matrix:
...
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="amazonlinux-2"
+    bundler_args:
+    script: bundle exec rake beaker
+  - rvm: 2.4.2
+    sudo: required
+    services: docker
+    env: BEAKER_set="amazonlinux-201803"
```
ghoneycutt: amazon linux 2018 is missing from vagrant

tests/provision_amazon.sh
treydock: Yep, I need to fix that.

.travis.yml
treydock: I couldn't find a vagrant box for 2018.

treydock: It seems Amazon focuses most their effort on official docker containers.

ghoneycutt: ok. we don't need to roll our own

treydock:
@ghoneycutt Updated to fix vagrant provisioning.  Sometimes `vagrant up` of `amazon-201703` box will fail but a `vagrant reload` will fix the VM.  I think some race condition or timing issue with vbguest additions getting installed.

I updated acceptance tests to test amazonlinux 2 against `amazonlinux:2` docker image now that puppet-agent-5.5.4 has patched Facter to properly identify Amazon Linux 2.

treydock:
@ghoneycutt rebased and squashed.

PR #940, opened 2018-07-10
Use facts hash and reference facts per Puppet docs

treydock:
# Pull Request Checklist

## Description

Consistent use of facts hash.

## Related Issue

Relates to #901 

## Motivation and Context

Make use of facts hash consistent and conform to Puppet docs: https://puppet.com/docs/puppet/5.5/lang_facts_and_builtin_vars.html

## General

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,7 +33,7 @@ class sensu (
     recurse => $etc_dir_purge,
   }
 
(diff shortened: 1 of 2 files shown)


treydock:
Also avoid using `osfamily` fact which is listed with `facter` only with `--show-legacy`.

PR #947, opened 2018-08-05
Add sensu_cluster_member type

treydock:
# Pull Request Checklist

## Description

Add a custom type that can add cluster members.

## Related Issue

Part of #901 

## Motivation and Context

sensuctl supports adding members.  The members won't actually be fully added until the new member is started with the proper configuration.  The necessary configuration is printed by this custom type in the info output which comes from output of the `sensuctl` command to add the member.

## How Has This Been Tested?

Added new way to run acceptance tests that spawn 3 backend servers.  Using only 2 caused issues (see https://github.com/sensu/sensu-go/issues/1890).  When `BEAKER_sensu_cluster` is `yes` or `true` the cluster acceptance tests are run.

## General


lib/puppet/provider/sensu_cluster_member/sensuctl.rb
```diff
@@ -0,0 +1,98 @@
+require File.expand_path(File.join(File.dirname(__FILE__), '..', 'sensuctl'))
```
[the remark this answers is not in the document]
treydock: There is an example in the type description that will get picked up with puppet-strings, do we also want additional examples in README?  If yes, should we also document how to setup a cluster?  It's currently not documented by Sensu but managed to figure it out based on their docker compose configs.  This new type is tightly coupled to configs in `backend.yml` as they won't work without both being implemented.

ghoneycutt:
@treydock please squash and rebase and this should pass tests once #948 is merged.

treydock:
@ghoneycutt Squashed and rebased.  Updated README with instructions on how to setup a cluster and also how to add a member to an existing cluster.

Vagrantfile
```diff
@@ -46,6 +46,32 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     backend.vm.provision :shell, :inline => "facter --custom-dir=/vagrant/lib/facter sensu_version"
   end
 
+  config.vm.define "sensu-backend-peer", primary: true, autostart: true do |backend|
```
ghoneycutt: call this one `sensu-backend-peer1` to be inline with `sensu-backend-peer2`

README.md
```diff
@@ -118,6 +118,58 @@ The backend system would collect all `sensu_check` resources.
...
+  listen-peer-urls: 'http://0.0.0.0:2380'
+  initial-cluster: 'backend1=http://192.168.0.1:2380,backend2=http://192.168.0.2:2380'
+  initial-advertise-peer-urls: "http://%{facts.ipaddress}:2380"
+  initial-cluster-state: 'new'
+  name: 'backend2'
+```
+
+Adding new members to a cluster requires two steps. First add the member using `sensu_cluster_member` type then second configure and start `sensu-backend` on the new node.
+
+First, add the member:
```
ghoneycutt: Does this mean it takes multiple puppet runs in a certain sequence? If so, could you elaborate?

ghoneycutt:
This strangely had an error with amazon linux 2018.3. Perhaps the service just took too long to start?

README.md
treydock: I think so, I'll try to come up with some better wording on order.  It's a really specific order of operations to avoid errors.  Basically add member on existing backend node with sensuctl then start new member with config that differs slightly from existing backend nodes.

treydock: Updated README, hopefully a bit clearer.  Let me know if needs improvements.

ghoneycutt:
The docs assume that two backends are running, which I thought is not good since you need one or three for quorum. If you have only two does is put the cluster is a not-good state? Could you give me a step by step in the README of why you have to setup one node, then can add more.

treydock:
Added a few bits of text to hopefully make it clear that adding members to an existing cluster differs from setting up a cluster from scratch.

ghoneycutt:
@treydock Could you please rebase. Is this ready to merge?

treydock:
@ghoneycutt rebased and squashed.

treydock:
I'd say ready for merge.  Open issue upstream but not sure would impact this PR since issue I opened was going from one to two node cluster and sensuctl hanging.  That's why acceptance tests go from two to three node cluster, seems keeps quorum.

PR #948, opened 2018-08-08
Fix acceptance tests

treydock:
# Pull Request Checklist

## Description

Pin beaker-rspec to avoid travis-ci pulling in really ancient versions that don't work

## Related Issue

Will fix tests for #947 

## Motivation and Context

Something with Travis CI recently changed where bundler installs a really old versions of `beaker-rspec` that doesn't work with the version of `serverspec` that gets installed.  This change will force the version of `beaker-rspec` that is known to work with this module's acceptance tests.

## How Has This Been Tested?

I only see this problem in Travis CI, can't reproduce locally.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ group :documentation do
 end
 
 group :system_tests do
-  gem 'beaker',                       :require => false
+  gem 'beaker', '~> 3.x',             :require => false
   gem 'beaker-rspec',                 :require => false
   gem 'serverspec',                   :require => false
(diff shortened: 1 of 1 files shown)


treydock:
I stand corrected.  The blame likely lies with beaker: https://github.com/puppetlabs/beaker/blob/master/docs/how_to/upgrade_from_3_to_4.md

I noticed a fresh `bundle install` on my local machine pulls in 4.0.0 and for some reason a really old version of `beaker-rspec` gets pulled in.

PR #954, opened 2018-08-16
Document and test resource purging

treydock:
# Pull Request Checklist

## Description

Document that all types (except `sensu_config`) support being purged.  Add some tests to verify what was documented in README.  Worth noting that `sensu_asset` doesn't yet support purging because deleting assets with sensuctl is not yet supported but the code in this module supports the operation.

## Related Issue

Relates to #901 

## Motivation and Context

Documenting the already supported ability to purge resources seems useful especially in the context of checks.  I only added tests for the documented example of purging checks since that seemed most likely what people would deploy.

## How Has This Been Tested?

`BEAKER_set=centos-7 BEAKER_sensu_full=yes  bundle exec rake beaker`

## General

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@
 3. [Usage - Configuration options and additional functionality](#usage)
     * [Basic Sensu backend](#basic-sensu-backend)
     * [Basic Sensu agent](#basic-sensu-agent)
+    * [Exported resources](#exported-resources)
+    * [Resource purging](#resource-purging)
 4. [Reference](#reference)
     * [Facts](#facts)
 5. [Limitations - OS compatibility, etc.](#limitations)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -118,6 +120,17 @@ The backend system would collect all `sensu_check` resources.
   Sensu_check <<||>>
 ```
 
+### Resource purging
+
+All the types provided by this module support purging except `sensu_config`.
+This example will remove all unmanaged Sensu checks:
+
+```puppet
+resources { 'sensu_check':
```
ghoneycutt: `sensu_asset` should be called out in the docs. We should also link to the ticket in sensu-go to get this to work or where they say they are not going to support it.

treydock:
Let me know if my latest commit works.  I added issue with `sensu_asset` to limitations section and referenced the limitation in purging section.

PR #956, opened 2018-08-18
Refactor sensu_check check_hooks property

treydock:
# Pull Request Checklist

## Description

Make sensu_check `check_hooks` property match what is actually output by sensuctl.

## Related Issue

Relates to #901.  Upstream documentation bug here: https://github.com/sensu/sensu-docs/issues/678

## Motivation and Context

The type `sensu_check` needs to treat `check_hooks` property the way it's handled by sensuctl.  Previously it was just expecting an array of hooks which isn't correct as hooks associated to a check also have a type like critical or warning.  This change ensures valid `check_hooks` are defined.

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -15,6 +15,17 @@ Manages Sensu checks
     interval      => 60,
   }
 
+@example Create a check that has a hook
+  sensu_check { 'test':
+    ensure        => 'present',
+    command       => 'check-cpu.sh -w 75 -c 90',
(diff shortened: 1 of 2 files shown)


treydock:
I will update acceptance tests to ensure that valid `check_hooks` are handled correctly.

treydock:
Acceptance tests updated and passing.

treydock:
I have local changes that are more complex but valid use of `check_hooks` and they are failing, this PR is not complete.

ghoneycutt:
@treydock could you please rebase so this will merge

Issue #959, opened 2018-08-18
sensu2 updates for types are poor for array and hash properties


treydock:
## Description of problem

The Puppet logged output of changes to array or hash properties take up multiple lines which would make log collection rather difficult.

This affects sensu2 types that use array or hash properties.

Example:

```
Notice: /Stage[main]/Main/Sensu_check[test]/extended_attributes: extended_attributes changed {
    'foo' => 'baz'
  } to {
    'foo' => 'bar'
  }
Notice: Applied catalog in 0.36 seconds
```

In my opinion the before/after values should all land on the same line which would make logging more complete and not require parsing multiple lines of text when ingested by tools like Splunk.

PR #961, opened 2018-08-18
Sensu2 - improve handling of array and hash properties

### LINKED ISSUE #960: Several sensu2 types shouldn't use Array property helper
## Description of problem

It seems that sorting an array of hashes is invalid and this means that properties that expect array of hashes can not use `PuppetX::Sensu::ArrayProperty` as their parent.

This is basically what happens when a property that is array of hashes has more than 1 element either on system or in Puppet catalog:

```
irb(main):007:0> [{'foo'=>'bar'},{'foo'=>'baz'}].sort
ArgumentError: comparison of Hash with Hash failed
        from (irb):7:in `sort'
        from (irb):7
        from /opt/puppetlabs/puppet/bin/irb:11:in `<main>'
```

treydock:
# Pull Request Checklist

## Description

Add a new type property that is ArrayOfHashesProperty.  This could be used with #956 so may be good to merge #956 first.  Also improve how logs from Puppet print changes of Array and Hash properties.

## Related Issue

Fixes #960 and #959 

## Motivation and Context

The logging of changes should all reside on one line for sake of log collection.  Also the use of ArrayProperty was not valid for Array of Hash properties so introduce a new property type.

### DIFF
--- a/lib/puppet/type/sensu_role.rb
+++ b/lib/puppet/type/sensu_role.rb
@@ -1,5 +1,6 @@
 require_relative '../../puppet_x/sensu/type'
 require_relative '../../puppet_x/sensu/array_property'
+require_relative '../../puppet_x/sensu/array_of_hashes_property'
 require_relative '../../puppet_x/sensu/hash_property'
 require_relative '../../puppet_x/sensu/integer_property'
 
@@ -27,7 +28,7 @@ DESC
     end
   end
 
-  newproperty(:rules, :array_matching => :all) do
+  newproperty(:rules, :array_matching => :all, :parent => PuppetX::Sensu::ArrayOfHashesProperty) do
     desc "The rulesets that a role applies."
     validate do |rule|
       if ! rule.is_a?(Hash)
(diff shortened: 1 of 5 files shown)


treydock:
Some examples of improved logging:

```
Notice: /Stage[main]/Main/Sensu_role[test]/rules: rules changed [{"type"=>"*", "environment"=>"*", "organization"=>"*", "permissions"=>["read"]}] to [{"type"=>"*", "environment"=>"*", "organization"=>"*", "permissions"=>["read", "create"]}, {"type"=>"*", "environment"=>"test", "organization"=>"*", "permissions"=>["create"]}]
```

```
Notice: /Stage[main]/Main/Sensu_check[test]/subscriptions: subscriptions changed ["demo", "linux"] to ["demo"]
Notice: /Stage[main]/Main/Sensu_check[test]/extended_attributes: extended_attributes changed {} to {"foo"=>"bar"}
```

ghoneycutt:
rebased in #963

PR #972, opened 2018-09-05
Do not start sensu-backend cluster in Vagrant by default

### LINKED ISSUE #971: Default `vagrant up` brings up too many systems
## Description of problem

Ran `vagrant up` and it brings up an entire sensu backend cluster. For basic usage and testing, we do not need that many systems to be running. This can be a detraction for users with limited resources.

`sensu-backend-peer1` and `sensu-backend-peer2` should not start by default. There should be documentation that explains that these should be started if a cluster is wanted.

treydock:
# Pull Request Checklist

## Description

Do not autostart the sensu-backend cluster nodes.  Should also document or provision an actual cluster (WIP).

## Related Issue

Fixes #971 

## Motivation and Context

Should not spawn multiple cluster backend vagrant boxes with simple `vagrant up`.  Also need to document and provision an actual cluster with vagrant.

## How Has This Been Tested?

WIP testing still the second half of this PR which is provision or document setting up the cluster.

## General

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ vagrant ssh sensu-backend
 Multiple Vagrant boxes are available for testing a sensu-backend cluster
 
 ```bash
-vagrant up sensu-backend sensu-backend-peer1 sensu-backend-peer2
+vagrant up sensu-backend-peer1 sensu-backend-peer2
+vagrant provision sensu-backend-peer1 sensu-backend-peer2
 ```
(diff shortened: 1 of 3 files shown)


treydock:
Documented how to provision cluster.  Have to run Puppet twice which is due to backend being in weird state when only 1 of 2 cluster nodes is actually online.

PR #975, opened 2018-09-13
Fixing invalid multibyte char error

pmjensen:
# Pull Request Checklist

## Description

This fixes and error that occurs during puppet agent run. There is a non-standard quote that puppet fails on.  Replacing the character allows puppet agent run to continue.  

Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Could not autoload puppet/type/sensu_check: /etc/puppetlabs/code/environments/pete_testing/modules/sensu/lib/puppet/type/sensu_check.rb:73: invalid multibyte char (US-ASCII)

## Related Issue

Fixes # .

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -98,7 +98,7 @@ DESC
   end
 
   newproperty(:stdin, :boolean => true) do
-    desc "If the Sensu agent writes JSON serialized Sensu entity and check data to the command process’ STDIN"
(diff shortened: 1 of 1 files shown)


treydock:
@ghoneycutt This wasn't part of v2.55.0 since targetted at sensu2 branch?  Just making sure.

ghoneycutt:
Correct, not released in v2.55.0. It showed up in the CHANGELOG because I forgot to tag this. It should get removed next time the CHANGELOG is generated for a release.

PR #977, opened 2018-09-18
Add description to resources that don't use default ensurable

treydock:
# Pull Request Checklist

## Description

Ensure all properties have a description when generating strings documentation.

## Motivation and Context

Resolves strings:generate warnings

## How Has This Been Tested?

`bundle exec rake strings:generate`


treydock:
The text used is what Puppet sets when you just do `ensurable` without a block.

PR #978, opened 2018-09-18
Do not manage the version of facter used for testing

ghoneycutt:
# Pull Request Checklist

This allows testing to use the ruby based facter so it can run. Without
this patch, tests cannot use facter 3 and will fail.

## Description
Remove facter bits from `Gemfile`

## Related Issue

PR #977 is failing and this PR is meant to fix that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,12 +6,6 @@ else
   gem 'puppet', :require => false
 end
 
-if facterversion = ENV['FACTER_GEM_VERSION'] || "~> 3.x"
-  gem 'facter', facterversion, :require => false
-else
-  gem 'facter', :require => false
-end
-
 group :development, :unit_tests do
   gem 'rake',                                             '< 11.0.0'
   gem 'rspec-puppet', '~> 2.5.0',                         :require => false


treydock:
The facts we use come from facterdb and the version of facter facts used is hardcoded at `3.8.0`:

https://github.com/sensu/sensu-puppet/blob/sensu2/spec/classes/backend_spec.rb#L4

The version of facter loaded should have little to no impact on class unit tests.  There may be some behavior changes with unit testing the `sensu_version` fact.

treydock:
Looks like RPM based acceptance tests are having issues, something unrelated to this PR.

treydock:
Puppet RPMs are broken.  They now install GPG key to `/etc/pki/rpm-gpg/RPM-GPG-KEY-puppet5-release` but the repo file still references `/etc/pki/rpm-gpg/RPM-GPG-KEY-puppet5`.  I've brought this up on Puppet Slack.

treydock:
Puppet fixed RPMs and tests now pass.

PR #982, opened 2018-09-18
Improvements to acceptance tests

### LINKED ISSUE #981: Re-evaluate parallel execution of acceptance tests
## Description of problem

It seems that if the parallel bits of acceptance tests fail the test will just hang and eventually time out rather than fail fast.  Initially parallel execution was used speed up tests but now that's mostly unnecessary because the long running tests only run for EL7 docker container.

Screen eventually gets flooded with this:

```
D, [2018-09-18T18:18:29.102433 #8877] DEBUG -- : Waiting for child processes.
```

treydock:
# Pull Request Checklist

## Description

Do not run parts of acceptance test setup in parallel
Rely on helpers and let travis tests set puppet collection version

## Related Issue

Fixes #981  .

## Motivation and Context

The parallel tests cause problems and no longer really needed.  Also use helpers which would allow for tests to be targeted at puppet6 collection when the time comes just by updating travis test matrix.

## How Has This Been Tested?

`BEAKER_set=centos-7 BEAKER_PUPPET_COLLECTION=puppet5 bundle exec rake beaker`

This fails fast as puppet-agent RPM is broken which is expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,67 +25,67 @@ matrix:
   - rvm: 2.4.4
     sudo: required
     services: docker
-    env: BEAKER_set="centos-6"
+    env: BEAKER_set="centos-6" BEAKER_PUPPET_COLLECTION=puppet5
     bundler_args:
     script: bundle exec rake beaker
   - rvm: 2.4.4
     sudo: required
     services: docker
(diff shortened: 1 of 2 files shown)


ghoneycutt:
@treydock could you please rebase this and resubmit

treydock:
Rebased

PR #984, opened 2018-09-19
removing agent from init.pp

pmjensen:
# Pull Request Checklist

## Description

The include statement in init.pp for sensu::agent does not allow for configuration of the agent on the backend server.  Removing it from init.pp allows for "resource-like class declarations" in the user's pp file for the backend server.  This allows for configuration customizations like adding subscriptions .  

## Related Issue

Fixes # .

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -28,7 +28,6 @@ class sensu (
   if $manage_repo {
     include ::sensu::repo
   }
-  include ::sensu::agent
 
   file { 'sensu_etc_dir':
     ensure  => 'directory',


ghoneycutt:
I don't understand the motivation behind this, could you please elaborate.

treydock:
I think I've faced what I think is a similar issue.  This is likely the problem:

```puppet
class { '::sensu::backend': }
class { '::sensu::agent': }
```

```
[root@sensu-backend ~]# puppet apply test1.pp
Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Sensu::Agent] is already declared; cannot redeclare (file: /root/test1.pp, line: 2) (file: /root/test1.pp, line: 2, column: 1) on node sensu-backend.example.com
```

Reversing the order of that puppet code does not suffer from errors because the class resource declaration comes before the include.

pmjensen:
That appears to have fixed my issue.  Thanks for the tip.

Regards

Peter Jensen
503 701 1829
Peter Jensen @ LinkedIn
<http://www.linkedin.com/pub/peter-jensen/31/869/2b8>
veni. vidi. velo.

ghoneycutt:
This is related to mixing resource like class statements and include. If you are not specifying any attributes for the class resource, you should always use include.

for example **always** use `include ::foo` instead of `class { '::foo'}`

PR #987, opened 2018-09-30
This makes sensu_gem work as advertised (cont)

treydock:
This is a continuation of #877 but with a script added that can be used with Vagrant to verify interaction between sensu_gem provider and RVM.


ghoneycutt:
@treydock script ends with the following by running it before applying the sensu_gem patch here.

```
GEM sensu-plugins-memory-checks installed
GEM sensu-plugins-memory-checks found in RVM
```

what is the expected before and after behavior?

treydock:
My comments on original PR weren't very clear.  I thought the behavior with the sensu_gem patch would be that gems don't end up in RVM.  I don't know what the behavior should be without the sensu_gem patch, I thought it was to have gems installed to /opt/sensu and not /usr/local/rvm.

ghoneycutt:
It seems that it was not finding checks installed by RVM, though it is now, so I'm closing.

PR #988, opened 2018-09-30
Support Puppet 6

### LINKED ISSUE #983: Support Puppet 6
## Description of problem

Looks like Puppet 6.0.0 was released today which is why suddenly facter 3.12.0 broke all tests.  Should be easy to support Puppet 6 tests once #982 is merged.

treydock:
# Pull Request Checklist

## Description

Add support for Puppet 6.

Use `filter` function for `check_notify` because unit tests would fail as it seems `delete_undef_values` was returning `[nil,nil,nil,nil]` and not actually filtering `Undef`.

## Related Issue

Fixes #983 

## Motivation and Context

Add support for Puppet 6.  Updated logic in acceptance tests setup to be simpler and more like sensu2 so that easy to use different Puppet collection in tests based on environment variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,46 +24,90 @@ matrix:
     env: PUPPET_GEM_VERSION="~> 4"
   - rvm: 2.4.4
     env: PUPPET_GEM_VERSION="~> 5"
-  - rvm: 2.4.2
+  - rvm: 2.5.1
+    env: PUPPET_GEM_VERSION="~> 6" FIXTURES_YML=".fixtures-puppet6.yml"
+  - rvm: 2.4.4
     sudo: required
     services: docker
-    env: BEAKER_set="centos-6"
+    env: BEAKER_set="centos-6" BEAKER_PUPPET_COLLECTION=puppet5
     bundler_args:
     script: bundle exec rake beaker
-  - rvm: 2.4.2
+  - rvm: 2.5.1
     sudo: required
     services: docker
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -20,50 +20,92 @@ script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec
...
-  - rvm: 2.4.2
+  - rvm: 2.5.1
+    env: PUPPET_GEM_VERSION="~> 6" FIXTURES_YML=".fixtures-puppet6.yml"
+  - rvm: 2.4.4
+    sudo: required
+    services: docker
+    env: BEAKER_set="centos-6" BEAKER_PUPPET_COLLECTION=puppet5
+    bundler_args:
+    script: bundle exec rake beaker
+  - rvm: 2.4.4
```
ghoneycutt: puppet6 should use 2.5.1

treydock: That version of RVM only affects beaker outside the container environment so no bearing on Puppet Ruby version since Ruby is installed into container using packages.  Still want it to be 2.5.1?  I kept consistent with other beaker tests since only interacts with beaker.

ghoneycutt: Yeah, because we want the testing environment to match the version we actually use. This is useful so any other ruby tooling used around your Puppet setup can be on the same version.

treydock:
Updated to use ruby 2.5.1 for all Puppet 6 related tests.

PR #991, opened 2018-10-12
Add support for Ubuntu 18.04

treydock:
# Pull Request Checklist

## Description

Add support for Ubuntu 18.04

## Related Issue

Fixes #990 .

## Motivation and Context

Match support that already exists in sensu2 branch.

## How Has This Been Tested?

Beaker acceptance tests.

PR #992, opened 2018-10-13
Update all Vagrant boxes to use Puppet5

treydock:
# Pull Request Checklist

## Description

Have all Vagrant boxes use Puppet5.

## Related Issue

This is needed for #991 Vagrantfile change to be functional.  Also should have likely been a part of #988 when Puppet 4 support was dropped.

## Motivation and Context

This module no longer tests against Puppet 4 so stop using Puppet 4 in Vagrant.


treydock:
My vagrant environment is not working at the moment so been difficult to test these changes locally.

Vagrantfile
```diff
@@ -65,20 +65,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     server.vm.provision :shell, :inline => "facter --custom-dir=/vagrant/lib/facter sensu_version"
   end
 
-  # This system is meant to be started without 'sensu-server' running.
-  config.vm.define "sensu-server-puppet5", autostart: false do |server|
```
ghoneycutt: This should still exist, but be puppet6

ghoneycutt:
`amazon201703-client` fails because it is not mounting `/vagrant`. We should probably only be supporting the latest version of Amazon Linux. They do not seem to put an EOL on their versions and just encourage using the latest.

ghoneycutt:
Both windows systems do not work. It seems like puppet is not in the PATH.

treydock:
I've not found official amazon linux vagrant boxes so it's always been hit or miss if the community ones work.  Amazon does release docker images.

treydock:
Windows box fixed.  Had to use x64 MSI instead of x86.  The chocolatey one won't fully work as still no chocolatey packages for sensu.

Also added commit to address #994

ghoneycutt:
@cwjohnston Are you looking to deploy to the chocolatey repo for windows users? It would really aid in adoption by making it easier to consume.

ghoneycutt:
@treydock let me know when the macos-client is fixed and I'll re-test everything again.

ghoneycutt:
Released in v2.60.0

Issue #994, opened 2018-10-14
Remove amazon linux from Vagrant setup


ghoneycutt:
We will still support Amazon Linux, though finding images that work reliably with VirtualBox is a challenge. Let's remove the Vagrant piece. The platform is still tested with the acceptance testing.

PR #995, opened 2018-10-21
Remove Amazon Linux from Vagrant

treydock:
# Pull Request Checklist

## Description

Remove Amazon boxes from Vagrant tests.

## Related Issue

sensu2 specific fix for #994.

## Motivation and Context

Amazon boxes are difficult to support so best to just remove and rely on docker based acceptance tests.


ghoneycutt:
note: this needs documentation as to why this is not in Vagrant. Should also add which versions of Puppet we support and the other supported platforms.

Issue #996, opened 2018-10-22
Document Amazon Linux support and only support the latest two versions


ghoneycutt:
We need to document which versions of Amazon Linux are supported (v2 and 2018.03) and the reasoning behind that, which is that we will support the latest release and the previous release.

ghoneycutt:
- [ ] sensu2 - remove 2017.03
- [ ] sensu1 - remove 2017.03
- [ ] sensu1 - add 2018.03
- [ ] sensu1 - add 2

ghoneycutt:
@treydock could you please add amazon linux 2 support for sensu1

PR #998, opened 2018-10-22
(GH-996) Use Amazon Linux 2018.03 instead of 2017.03 as it is EOL

ghoneycutt:
Though Amazon does not announce end of life (EOL) for its releases, it
does encourage you to use the latest releases which include versions 2
and 2018.03.

Remove Amazon 2017.03 and add 2018.03


ghoneycutt:
Released in v2.56.0

ghoneycutt:
Released in v2.56.0

PR #1000, opened 2018-10-24
add field ttl_status to sensu_check (cont)

treydock:
This is a continuation of #927 with unit tests added for new `ttl_status` property.


spec/unit/sensu_check_spec.rb
```diff
@@ -114,4 +114,27 @@
       end
     end
   end
+
+  describe 'ttl_status parameter' do
+    subject { described_class.new(resource_hash)[:ttl_status] }
+
+    valid = [1, 0, '1', '0']
```
ghoneycutt: `'absent'` is also valid and so is `undef`

ghoneycutt: Data type: 

```puppet
Variant[Undef,Enum['absent'],Integer]
```

ghoneycutt: '1' and '0' should not work since they are not integers

treydock: They work because the type property munges with `to_i` if not absent.  The validation in type takes place before munge.  The data type isn't actually tested yet, should the behavior of type property match data type or should data type be updated to allow numeric strings?

ghoneycutt: Seems the data type should be updated to allow for stringified integers by using a regex.

REFERENCE.md
```diff
@@ -1914,6 +1914,15 @@ considered stale. Set this to 'absent' to remove it completely.
 
 Default value: `undef`
 
+##### `ttl_status`
```
ghoneycutt: run `rake reference` to re-generate this as the data type in the docs does not match the code.

treydock:
Removed change to REFERENCE.md and squashed all commits into one.

ghoneycutt:
Released in v2.57.0

PR #1002, opened 2018-10-27
Remove sensu::backend username and password parameters

treydock:
# Pull Request Checklist

## Description

Remove `username` and `password` parameters from `sensu::backend`.  Document limitation.

## Related Issue

Part of #901 

## Motivation and Context

Changing either of these parameters would break the bootstrap of sensuctl.  These options will be added back once upstream supports better management of users which will allow this module to properly manage user resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,12 +84,10 @@ vagrant provision sensu-backend-peer1 sensu-backend-peer2
 
 ### Basic Sensu backend
 
-The following example will configure sensu-backend and add a check.  It's advisable to not rely on the default password.
+The following example will configure sensu-backend and add a check.
 
 ```puppet
-  class { 'sensu::backend':
-    password => 'P@ssw0rd!',
-  }
+  include sensu::backend
   sensu_check { 'check-cpu':
     ensure        => 'present',
     command       => 'check-cpu.sh -w 75 -c 90',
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -222,6 +220,8 @@ This module does not support adding `sensuctl` resources on a host other than th
 
 The type `sensu_asset` does not at this time support `ensure => absent` due to a limitation with sensuctl, see [sensu-go#988](https://github.com/sensu/sensu-go/issues/988).
 
+The sensuctl username and password can not be configured due to limitations with sensuctl. The checklist at [sensu-puppet#901](https://github.com/sensu/sensu-puppet/issues/901) will be updated as progress is made.
+
```
ghoneycutt: Please add that we use the default values `admin` and `P@ssw0rd!`

manifests/backend.pp
```diff
@@ -41,8 +37,6 @@
   Hash $config_hash = {},
   String $url_host = '127.0.0.1',
   Stdlib::Port $url_port = 8080,
-  String $username = 'admin',
-  String $password = 'P@ssw0rd!',
```
ghoneycutt: Could we keep these if we tell people how to set it manually and change the hiera data?

Could we use Tasks to help?

treydock: The way we use both username and password at this time is the initial bootstrap with `sensuctl configure` and that must be `admin` and `P@ssw0rd!`.  After initial bootstrap it can be changed to something else and the parameters wouldn't actually change anything since they are only used for bootstrap.  If we left the parameters it may mislead people to think they have some purpose when really they aren't yet used.  A task could be useful but would be just as easy for us to document the commands to use for sensuctl to change the password.

ghoneycutt: Makes sense. Please add more to the docs and this will be ready for merge.

treydock: Updated README with commands to perform the manual steps of changing password.  Let me know if more is needed.

PR #1011, opened 2018-11-10
Add sensu_user type

### LINKED ISSUE #928: sensu2 admin password handling is incorrect
## Description of problem

If this module is used for the first time with a non-default admin password the `sensuctl configure` step will fail.

I do not think a solution will be possible until the `sensu_user` type can be created at which point that type and provider will have to update the admin user and run `sensuctl configure`.  The `sensuctl configure` used by `backend.pp` needs to use the default password if first time being invoked.

Below output is a fresh install.

```
[root@sensu-backend ~]# sensuctl configure -n --url http://localhost:8080 --username admin --password foobar

Error: unable to authenticate with error: Request unauthorized

[root@sensu-backend ~]# sensuctl configure -n --url http://localhost:8080 --username admin --password 'P@ssw0rd!'
[root@sensu-backend ~]# sensuctl user list
  Username     Roles     Enabled
 ────────── ─────────── ─────────
  admin      admin       true
  agent      agent       true
  sensu      read-only   true
```

- Your OS: EL7
- Your Ruby version: 2.4
- Your version of Puppet: puppet-agent 5
- Your version of Sensu: sensu 2 nightlies
- Your version of this module: sensu2 branch

treydock:
# Pull Request Checklist

## Description

Add `sensu_user` type and fix support for changing sensuctl credentials.

## Related Issue

Fixes #928 , relates to #901.  Will obsolete #1002

## Motivation and Context

This module needs a way to use something other than the default `admin` password and also add additional users.

## How Has This Been Tested?

Unit tests and some ad-hoc tests using Vagrant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,9 +87,13 @@ vagrant provision sensu-backend-peer1 sensu-backend-peer2
 
 The following example will configure sensu-backend, sensu-agent on backend and add a check.
 By default this module will configure the backend to use Puppet's SSL certificate and CA.
+It's advisable to not rely on the default password. Changing the password requires providing the previous password via `old_password`.
 
 ```puppet
-  include sensu::backend
+  class { 'sensu::backend':
+    password     => 'supersecret',
+    old_password => 'P@ssw0rd!',
+  }
   include sensu::agent
   sensu_check { 'check-cpu':
     ensure        => 'present',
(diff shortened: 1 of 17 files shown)


treydock:
Tests will fail until https://github.com/sensu/sensu-go/pull/2278 is merged and new RPMs are released.

treydock:
Updated PR to account for RBAC changes, mostly replacing `roles` with `groups`.  Still need to fix auto requires so that role bindings auto require users though it appears possible to create a binding for non-existent user which not sure we should rely on.

treydock:
@ghoneycutt Rebased with latest sensu2 branch code.  I also figured out a way to test passwords using `/auth` API endpoint that's used by `sensuctl configure` when generating the access tokens.

I am waiting to hear back from someone at Sensu on https://github.com/sensu/sensu-go/issues/2578 so I know if the `sensu_user` type needs to be executing `sensuctl configure` upon password changes.  So far testing has shown that changes to passwords do NOT require re-running `sensuctl configure` but I'd like confirmation from upstream this is intended behavior.

treydock:
Need to remove `ensure => absent` because seems there is no delete subcommand for users: https://github.com/sensu/sensu-go/issues/2540

ghoneycutt:
@treydock could you add to the sequence of testing to show that you can change from the original password, to another password, to a third password. When you change something once you might not find bugs associated with changing it multiple times.

treydock:
Updated commit to not check password for disabled users as that appears to not be possible.  Also added commit that returns error if `ensure => absent` is used and added this limitation to README.

treydock:
Updated main commit to include tests so first creates user, second changes password and third changes password again.

treydock:
Got confirmation that my observation of not having to re-run `sensuctl configure` when changing passwords is likely a bug so I'm going to keep the `sensu_user` type running `sensuctl configure` when password changes and the appropriate properties are set since really only needed for `admin` user.

treydock:
I think this PR and #1034 will conflict and I think #1034 has implications for this PR so ideally #1034 would be merged first then this one rebased.

treydock:
@ghoneycutt rebased and tests passing.

PR #1010, opened 2018-11-10
Sensu2 namespaces and extended_attributes changes

treydock:
# Pull Request Checklist

## Description

Replace `organization` and `environment` properties with `namespace`.
Replace `sensu_organization` and `sensu_environment` types with `sensu_namespace`.
Remove `metadata` property from `sensu_asset`
Add `labels` and `annotations` properties to `sensu_check`, `sensu_asset` and `sensu_entity`.
Only autorequire this module's `sensu_api_validator` for all types.
Improve strings docs

## Related Issue

Fixes #1008, #1006, and #1005 

## Motivation and Context

Upstream sensu-go has made some changes that require updates to the Puppet code.

## How Has This Been Tested?

Only tested with unit tests.  Unable to test with acceptance tests until new beta RPMs are released.

Issue #1008, opened 2018-11-07
Remove or update extended_attributes for sensu2 types


treydock:
Upstream replaced extended attributes with labels.  This will impact `sensu_check` and `sensu_entity` at the very least.

Upstream change: https://github.com/sensu/sensu-go/pull/2283

treydock:
Looks like this is now what we get:

```
[
  {
    "command": "check-cpu.sh -w 75 -c 90",
    "handlers": [],
    "high_flap_threshold": 0,
    "interval": 60,
    "low_flap_threshold": 0,
    "publish": true,
    "runtime_assets": [],
    "subscriptions": [
      "linux"
    ],
    "proxy_entity_id": "",
    "check_hooks": null,
    "stdin": false,
    "subdue": null,
    "ttl": 0,
    "timeout": 0,
    "round_robin": false,
    "output_metric_format": "",
    "output_metric_handlers": [],
    "env_vars": null,
    "metadata": {
      "name": "check-cpu",
      "namespace": "default",
      "labels": null,
      "annotations": null
    }
  }
]
```
This will have to be bundled with #1006 since the `metadata` key for both `sensu_check` and `sensu_entity` is what holds the `namespace` value.

ghoneycutt:
@treydock can this be closed now that https://github.com/sensu/sensu-puppet/pull/1013 is merged?

treydock:
Yep, closing.

Issue #1006, opened 2018-11-05
Replace environment & organization with namespace for sensu2 types


treydock:
All types that use environment and/or organization will need to be updated to use namespaces.  Also flags passed to sensuctl will need to be updated.

Upstream change: https://github.com/sensu/sensu-go/pull/2254

ghoneycutt:
@treydock can this be closed now that https://github.com/sensu/sensu-puppet/pull/1013 is merged?

treydock:
Yep, closing.

Issue #1005, opened 2018-11-05
Remove sensu_asset metadata parameter (sensu2)


treydock:
Upstream removed metadata from assets (https://github.com/sensu/sensu-go/pull/2273).  Need to update sensu_asset type to account for this change.

treydock:
So this relates a little to #1006 but upon looking it seems the `metadata` property was replaced by `metadata` hash now used also by `check` and `entity` which means arbitrary key/values for `labels` and `annotations` will be possible with assets.

ghoneycutt:
@treydock can this be closed now that https://github.com/sensu/sensu-puppet/pull/1013 is merged?

treydock:
Yep, closing.

PR #1013, opened 2018-11-13
Replace id properties with name and use metadata

### LINKED ISSUE #1012: Update sensu2 to replace id properties with name
Upstream is changing references to `id` with `name` and this will require some changes to several types.  Also appears like the `metadata` key will be used for `name` and `namespace` combinations and this will also require some logic changes for most types.

Upstream PR: https://github.com/sensu/sensu-go/pull/2308

treydock:
# Pull Request Checklist

## Description

Several properties that had `id` in name are now using `name`.  Also all resources that belong to a namespace are now using the `metadata` key to hold both `name` and `namespace` attributes.

## Related Issue

Fixes #1012.  Contains fixes for #1005 , #1006 and #1008.  This pull request contains the changes in #1010 because this is an extension of the changes in that pull request.

## Motivation and Context

Upstream made some massive changes to the data structures for JSON in/out of sensuctl.  These changes should get the sensu2 branch full updated to work with current code for sensu-go.

## How Has This Been Tested?

Unit tests only.  The acceptance tests will not work until new RPMs are released.


ghoneycutt:
@portertech @cwjohnston  We need nightly releases of the beta software released so we can automatically test the changes that are happening with sensu-go.

cwjohnston:
@ghoneycutt as I understand it, an effort to revamp packaging has lead us to turn off nightly builds for the time being. Many of us are advocating for the return of nightly builds, but at this time there's currently no specific timeline for bringing them back.

treydock:
Added commit to resolve #1014 and one item in #901

treydock:
Two new commits.  First adds descriptions to properties that were missing now that Sensu docs have all these new things documented.

Second commit adds `labels` and `annotations` to all types that didn't already have them and support the new resource metadata.

PR #1017, opened 2018-11-23
Rework RBAC resources to work with sensu-go changes

treydock:
# Pull Request Checklist

## Description

Refactor sensu_role
Add sensu_role_binding
Add sensu_cluster_role
Add sensu_cluster_role_binding

## Related Issue

Fixes #1016

## Motivation and Context

Sensu-go rewrote the RBAC resources and this PR is necessary to work with upstream changes.  Upstream change: https://github.com/sensu/sensu-go/pull/2338.  This pull request will not work until that PR is merged and released as RPM.  Also currently testing in Vagrant is broken until I get confirmation on `subjects` data structure since docs do not match sensu-go code.  The docs for RBAC changes are already published: https://docs.sensu.io/sensu-go/5.0/reference/rbac/

## How Has This Been Tested?

In Vagrant `sensu-backend` box with binaries built from source with upstream pull request applied.

PR #1020, opened 2018-12-06
use stable package sensu-go

maxadamo:
the whole module clearly needs more work, but this change make the module usable with the new version of sensu-go

# Pull Request Checklist

## Related Issue

Fixes # .

## General

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -12,10 +12,9 @@ class sensu::repo {
       } else {
         $repo_release = $facts['os']['release']['major']
       }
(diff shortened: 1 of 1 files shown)


treydock:
@maxadamo This will have to be integrated into #1017 because a lot has changed between last beta and stable that breaks several Puppet resources and will cause tests to fail without changes in #1017 .  Also changing the repo path also requires updating the APT key and updating the unit tests.  I'll integrate this PR's change into #1017.

ghoneycutt:
Thank you for the pull request. We now have this functionality.

Issue #1025, opened 2018-12-26
Manage state of sensuctl configure


treydock:
Currently the Exec for `sensuctl configure` only runs once on first application of Puppet on the backend and doesn't get re-run if parameters are changed.  With #1024 it will be necessary to re-run if someone changes from non-SSL to SSL.  There should likely be a type/provider written that knows if `sensuctl configure` needs to be re-run.  If the provider is able to verify the access token it should also greatly simplify the implementation of future `sensu_user` in #1011 as that provider was going to have to execute `sensuctl configure` upon password changes.

PR #1024, opened 2018-12-25
Fix/backend ssl

rickatnight11:
# Pull Request Checklist

Fixes `sensu_api_validator` when Backend is using SSL

## Description
Adds a `sensu::backend::use_ssl` parameter (defaults to `false`), so the `sensu_api_validator` Exec can communicate with the Sensu Backend of SSL is enabled.

Right now the Backend has SSL implicitly enabled if `cert-file` and `key-file` are specified in `backend.yml`.  Perhaps the change in this PR could be extended to more explicitly set SSL.

## Related Issue

Fixes https://github.com/sensu/sensu-puppet/issues/1023

### Before

```
$ puppet agent -t
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for mgmt01.example.org
Info: Applying configuration version '1545709040'
Notice: /Stage[main]/Sensu::Backend/Service[sensu-backend]/ensure: ensure changed 'stopped' to 'running'
Info: /Stage[main]/Sensu::Backend/Service[sensu-backend]: Unscheduling refresh on Service[sensu-backend]
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): Failed to open TCP connection to 0.0.0.0:7001 (Connection refused - connect(2) for "0.0.0.0" port 7001)
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
...
```

### After

```
$ puppet agent -t
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for mgmt01.example.org
Info: Applying configuration version '1545709163'
[...]

PR #1026, opened 2018-12-27
Fix/backend ssl (cont.)

### LINKED ISSUE #1023: sensu_api_validator fails when Backend has SSL enabled
## Description of problem

Puppet run fails on `sensu_api_validator` Exec when Backend has SSL enabled (`cert-file` and `key-file` specified in `backend.yml`, as it always assumes HTTP.  The underlying code actually has a `use_ssl` parameter, but the incantation in `sensu::backend` doesn't currently set it.

```
$ puppet agent -t
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Caching catalog for mgmt01.example.org
Info: Applying configuration version '1545709040'
Notice: /Stage[main]/Sensu::Backend/Service[sensu-backend]/ensure: ensure changed 'stopped' to 'running'
Info: /Stage[main]/Sensu::Backend/Service[sensu-backend]: Unscheduling refresh on Service[sensu-backend]
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): Failed to open TCP connection to 0.0.0.0:7001 (Connection refused - connect(2) for "0.0.0.0" port 7001)
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
Notice: Unable to connect to sensu_api server (http://0.0.0.0:7001): wrong status line: "\x15\x03\x01\x00\x02\x02"
Notice: Failed to connect to sensu_api; sleeping 2 seconds before retry
...
```

- Your OS: CentOS 7.5
- Your version of Puppet: OS Puppet 6.1.0
- Your version of Sensu: Sensu-Go 5.1.0
- Your version of this module: `sensu2` (https://github.com/sensu/sensu-puppet/commit/fa0dfd176075925c4f9fc1f7d9c5e2c86d3b38de)

treydock:
# Pull Request Checklist

## Description

Continuation of #1024.
Expose $use_ssl flag for backend
Add force to /etc/sensu to ensure directories can be removed
Log output from 'sensuctl configure' Exec

## Related Issue

Fixes #1023 and replaces #1024.

## Motivation and Context

Better support for SSL backend.  The acceptance tests have to resort to removing configuration generated from `sensuctl configure` to ensure `sensuctl` functions switch to and from SSL.  This is something that will be eventually handled once #1025 is implemented.

## How Has This Been Tested?

Beaker.

### DIFF
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -28,6 +28,8 @@
 #   Sensu backend host used to configure sensuctl and verify API access.
 # @param url_port
 #   Sensu backend port used to configure sensuctl and verify API access.
+# @param use_ssl
+#   Sensu backend service uses SSL
 #
 class sensu::backend (
   Optional[String] $version = undef,
@@ -40,6 +42,7 @@ class sensu::backend (
   Hash $config_hash = {},
   String $url_host = '127.0.0.1',
   Stdlib::Port $url_port = 8080,
+  Boolean $use_ssl = false,
 ) {
 
   include ::sensu
@@ -51,7 +54,14 @@ class sensu::backend (
   }
   $config = $default_config + $config_hash
 
-  $url = "http://${url_host}:${url_port}"
+  if $use_ssl {
+    $url_protocol = 'https'
+  }
+  else {
+    $url_protocol = 'http'
+  }
+
+  $url = "${url_protocol}://${url_host}:${url_port}"
 
   if $version == undef {
     $_version = $::sensu::version
@@ -68,6 +78,7 @@ class sensu::backend (
   sensu_api_validator { 'sensu':
     sensu_api_server => $url_host,
     sensu_api_port   => $url_port,
+    use_ssl          => $use_ssl,
     require          => Service['sensu-backend'],
   }
   # Ensure sensu-backend is up before starting sensu-agent
@@ -76,10 +87,11 @@ class sensu::backend (
   $sensuctl_configure = "sensuctl configure -n --url '${url}' --username 'admin' --password 'P@ssw0rd!'"
   $sensuctl_configure_creates = '/root/.config/sensu/sensuctl/cluster'
   exec { 'sensuctl_configure':
-    command => "${sensuctl_configure} || rm -f ${sensuctl_configure_creates}",
-    creates => $sensuctl_configure_creates,
-    path    => '/bin:/sbin:/usr/bin:/usr/sbin',
-    require => Sensu_api_validator['sensu'],
+    command   => "${sensuctl_configure} || rm -f ${sensuctl_configure_creates}",
+    creates   => $sensuctl_configure_creates,
+    path      => '/bin:/sbin:/usr/bin:/usr/sbin',
+    logoutput => true,
+    require   => Sensu_api_validator['sensu'],
   }
 
   package { 'sensu-go-backend':
--- a/spec/acceptance/00_backend_spec.rb
+++ b/spec/acceptance/00_backend_spec.rb
@@ -13,14 +13,81 @@ describe 'sensu::backend class', unless: RSpec.configuration.sensu_cluster do
       apply_manifest_on(node, pp, :catch_changes  => true)
     end
 
-    it do
-      on node, "/opt/puppetlabs/bin/facter os" do
-        puts stdout
-      end
-    end
     describe service('sensu-backend'), :node => node do
       it { should be_enabled }
       it { should be_running }
+    end
+  end
+
+  context 'with SSL' do
+    it 'should bootstrap certs' do
+      on node, 'mkdir -p /root/ssl'
+      on node, 'openssl req -newkey rsa:2048 -nodes -keyout /root/ssl/ca_key.pem -x509 -days 365 -out /root/ssl/ca.pem -subj "/CN=localhost"'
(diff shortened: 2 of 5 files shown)


treydock:
CC @rickatnight11 

@ghoneycutt Do we want to document in the README how to do SSL with this module?  It currently requires some manual steps if going from non-SSL to SSL.

ghoneycutt:
What's the reasoning behind not using SSL by default? Is that just how the docs for sensu-go read?

spec/acceptance/00_backend_spec.rb
```diff
@@ -13,14 +13,75 @@
...
+    end
+  end
+
+  context 'with SSL' do
+    it 'should bootstrap certs' do
+      on node, 'mkdir -p /root/ssl'
+      on node, 'openssl req -newkey rsa:2048 -nodes -keyout /root/ssl/ca_key.pem -x509 -days 365 -out /root/ssl/ca.pem -subj "/CN=localhost"'
+      on node, 'openssl req -newkey rsa:2048 -nodes -keyout /root/ssl/key.pem -out /root/ssl/localhost.csr -subj "/CN=localhost"'
+      on node, 'openssl x509 -req -in /root/ssl/localhost.csr -CA /root/ssl/ca.pem -CAkey /root/ssl/ca_key.pem -CAcreateserial -out /root/ssl/cert.pem -days 365 -sha256'
+      on node, 'cat /root/ssl/ca.pem >> /etc/pki/tls/certs/ca-bundle.crt'
```
ghoneycutt: From the acceptance tests failures it would seem that this works differently with Ubuntu.

rickatnight11:
@ghoneycutt setting up SSL requires setting up valid trust between your Backend and Agents, something that will always be environment-specific and falls outside the scope of this module.

The sensu-go docs don't actually have much documentation on setting up SSL. There are very few SSL options exposed to the administrator, and most of them seem to just hand the settings to the underlying HTTP/TLS libraries without really any if the Sensu code really grokking it.

treydock:
@ghoneycutt To add to why we don't use SSL by default, it appears like telling the backend "trust CA at /path/to/ca.pem" works just fine but there is no corresponding sensu-specific way for sensuctl to trust the backend's cert so it has to be added to the system's trust store if it's not already trusted.  Also the default and example configs from sensu-go don't use SSL.

I contemplated documenting a method of using Puppet certs for sensu as I've done that for other services like LDAP.  The only issue there would be I'd want to implement the documented steps with beaker acceptance tests and the default mode of beaker acceptance is to run Puppet without a master so no SSL certs are generated by default.  Adding SSL certs for acceptance would be a bit involved as would have to have two approaches to support Puppet 5 and Puppet 6.

treydock:
Fixed acceptance tests on Ubuntu systems.

ghoneycutt:
@treydock we can check in certs to the repo for use with the acceptance testing.

treydock:
@ghoneycutt Added the certs to repo and using those during acceptance tests.

treydock:
@ghoneycutt Added same comment to #1027 which replaces this pull request.

spec/acceptance/00_backend_spec.rb
```diff
@@ -13,14 +13,81 @@
...
+    describe service('sensu-backend'), :node => node do
+      it { should be_enabled }
+      it { should be_running }
+    end
+  end
+
+  context 'with SSL' do
+    it 'should bootstrap certs' do
+      on node, 'mkdir -p /root/ssl'
+      on node, 'openssl req -newkey rsa:2048 -nodes -keyout /root/ssl/ca_key.pem -x509 -days 365 -out /root/ssl/ca.pem -subj "/CN=localhost"'
```
ghoneycutt: We could just copy files that are checked into this repo that correspond with puppet certs. How are we checking that an agent on the other non-backend node can talk SSL?

manifests/backend.pp
```diff
@@ -40,6 +42,7 @@
   Hash $config_hash = {},
   String $url_host = '127.0.0.1',
   Stdlib::Port $url_port = 8080,
+  Boolean $use_ssl = false,
```
ghoneycutt: Do the docs say anything about using a different port when using SSL?

treydock: The sensu-go docs don't say to use a different port, and it's not necessary.

treydock: https://docs.sensu.io/sensu-go/5.0/guides/securing-sensu/#securing-the-api-and-the-dashboard

spec/acceptance/00_backend_spec.rb
treydock: I think I'm going to move the SSL tests to a separate file that comes after the basic backend/agent tests so we can do end-to-end tests that SSL for both backend and agent works.

Can't copy the files because when beaker copies the module to nodes it excludes tests and spec directories.  Also the changes I'm making per these comments are in #1027.

treydock:
Closing this PR in favor of #1027.

PR #1027, opened 2018-12-29
SSL support and better handling of configure

### LINKED ISSUE #1025: Manage state of sensuctl configure

treydock:
# Pull Request Checklist

## Description

Support SSL by adding `use_ssl` parameter to `sensu::backend`.
Add `sensu_configure` type that handles running `sensuctl configure`.
Make type autorequire tests be a shared example.

## Related Issue

Fixes #1025.  This will replace #1024 and #1026.

## Motivation and Context

The SSL support makes this module properly support SSL by ensuring validator can be told to use SSL.

Replace `Exec` with a native type removes need for someone to run `sensuctl configure` when switching between SSL and non-SSL or vice versa.

## How Has This Been Tested?

Unit and beaker


treydock:
CC @rickatnight11 

Also should be noted that changes to `admin` user password do not appear to require re-running `sensuctl configure` since the API token and refresh token still work regardless of password it seems.  The only thing that actually can change with `sensu_configure` type after bootstrap is the URL of backend API.  The password is still needing to be accurate and provided in case the URL has to change as `sensuctl configure` requires both `--username` and `--password`.

treydock:
@ghoneycutt Added the certs to repo and using those during acceptance tests.

treydock:
@ghoneycutt Updated acceptance tests to ensure that both backend and agent work when SSL is used and that agent can communicate with backend with SSL used.  The SSL certs and CA are in repo.  I had to use cert with SAN to support it matching both `localhost` used by backend sensuctl and `sensu_backend` hostname used by agent.

PR #1028, opened 2018-12-29
Do not include sensu::agent class in main sensu

treydock:
# Pull Request Checklist

## Description

Do not include `sensu::agent` in `sensu` class.

## Motivation and Context

This allows the usage of resource-like declaration of classes like the following without being order dependent:
```puppet
class { 'sensu::agent':
...
}
class { 'sensu::backend':
...
}
```

The usage examples in README already describe directly calling `sensu::agent` and `sensu::backend` classes.

### DIFF
--- a/spec/classes/backend_spec.rb
+++ b/spec/classes/backend_spec.rb
@@ -7,7 +7,9 @@ describe 'sensu::backend', :type => :class do
       describe 'with default values for all parameters' do
         it { should compile.with_all_deps }
 
-        it { should contain_class('sensu::backend')}
+        it { should contain_class('sensu::backend') }
+        it { should contain_class('sensu') }
+        it { should_not contain_class('sensu::agent') }
 
         it {
           should contain_package('sensu-go-cli').with({
--- a/README.md
+++ b/README.md
@@ -84,10 +84,11 @@ vagrant provision sensu-backend-peer1 sensu-backend-peer2
 
 ### Basic Sensu backend
 
-The following example will configure sensu-backend and add a check.
+The following example will configure sensu-backend, sensu-agent on backend and add a check.
 
 ```puppet
   include sensu::backend
+  include sensu::agent
   sensu_check { 'check-cpu':
     ensure        => 'present',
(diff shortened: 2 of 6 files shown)


ghoneycutt:
Is there an issue associated with this? What's the motivation? The way the module works now you can just `include ::sensu` and have a working agent and if you `include ::sensu::backend` then it will run the backend along with an agent. `sensu::agent` is a private class not meant to be called directly.

treydock:
If a site setup a profile class that uses `class {}` style declaration it would be very order dependent to avoid duplicate declaration errors.  We document and test using resource-like declaration and also calling `sensu::agent` directly because it contains parameters like `config_hash` that are modified during tests.

We impose fewer restrictions on how someone uses this module with these changes.  If we keep the current code then we are partially forcing users to use only `include` style declaration unless they want to define with `class {}` and do so in the right order.

ghoneycutt:
Looks like the README already tells users to include `sensu::agent` instead of `sensu` and for the backend to include `sensu::backend` It should also be updated to show that the backend should also include `sensu::agent`.

There are not any acceptance tests around this. With this patch the backend would not necessarily run the agent. Could you add tests that would show this.

treydock:
Updated README to show backend also needing to include `sensu::agent` class.  Added unit tests to ensure backend isn't including `sensu::agent` and added acceptance tests that agent isn't present with just the backend and that the backend and agent can run together.

spec/classes/backend_spec.rb
```diff
@@ -7,7 +7,9 @@
       describe 'with default values for all parameters' do
         it { should compile.with_all_deps }
 
-        it { should contain_class('sensu::backend')}
+        it { should contain_class('sensu::backend') }
+        it { should contain_class('sensu') }
+        it { should_not contain_class('sensu::agent') }
```
ghoneycutt: Would not expect agent class to be here.

treydock: The test is using `should_not` so testing that `sensu::agent` isn't in catalog for just backend.

PR #1030, opened 2018-12-29
Improve facts by adding version information facts specific to each component

treydock:
# Pull Request Checklist

## Description

Add facts specific to each sensu-go component.

## Motivation and Context

It would be useful to have facts not just for the agent but also backends and sensuctl, especially if more than one backend exists.  Also using facts hash so that can get all the data out of `version` subcommands.

## How Has This Been Tested?

Unit and beaker


spec/acceptance/facts_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      # Run it twice and test for idempotency
+      apply_manifest_on(backend, pp, :catch_failures => true)
+      apply_manifest_on(backend, pp, :catch_changes  => true)
+      # Simulate plugin sync
+      fact_path = File.join(File.dirname(__FILE__), '../..', 'lib/facter')
+      scp_to(backend, fact_path, '/opt/puppetlabs/puppet/cache/lib/')
+    end
+
+    it "should have backend facts" do
```
ghoneycutt: perhaps we should test that before including the class, though after "plugin sync", that the facts do not exist.

treydock: The only way to do that would be to uninstall the various components first because tests in other files loaded first will install everything so when `facts_spec.rb` runs the backend and agent are already installed.  I only included the classes in this file to allow it to be run directly via `bundle exec rspec spec/acceptance/facts_spec.rb` which is sometimes handy to speed up testing.

ghoneycutt: ok, makes sense to me

PR #1032, opened 2018-12-30
Update JSON for sensuctl create to match docs and disallow asset removal

### LINKED ISSUE #1031: Make type usage of sensuctl create match docs
The current implementation of JSON passed to `sensuctl create -f` uses top level keys `type` and `spec`.  It appears like now there are several more top-level keys that need to be set such as moving `metadata` to top-level and specifying `api_version`.

Another look at all keys being managed by Puppet should be done to ensure we're matching the sensu-go docs.

Now:
```json
{
  "type": "Check",
  "spec": {
    "metadata": {
      "name": "test",
      "namespace": "default"
    },
    "command": "check-http.rb",
    "subscriptions": [
      "demo"
    ],
    "handlers": [
      "email"
    ],
    "interval": 60
  }
}
```

What it should look like based on docs:

```json
{
  "type": "Check",
  "api_version": "core/v2",
  "metadata": {
    "name": "test",
    "namespace": "default"
  },
  "spec": {
    "command": "check-http.rb",
    "subscriptions": [
      "demo"
    ],
    "handlers": [
      "email"
    ],
    "interval": 60
  }
}
```

treydock:
# Pull Request Checklist

## Description

Updated the JSON passed to `sensuctl create -f` to match the sensu-go docs for 5.0 and 5.1.
Also have Puppet error if `ensure => absent` is passed to `sensu_asset` as removal is not yet supported in sensu-go.

## Related Issue

Fixes #1031 

## Motivation and Context

The docs for 5.0 and 5.1 both reference a different format for `sensuctl create -f` than was documented before GA.  This PR updates this module's generated JSON to match the docs though the old format still works it seems but good to not rely on undocumented format.

Also produce compile error if `ensure => absent` is passed to `sensu_asset`.  Previously the acceptance tests were just skipping, now errors are produced for the unsupported behavior and the acceptance tests check this behavior.

## How Has This Been Tested?

Unit and beaker


treydock:
This PR is not complete.  Part of #1031 is also to update all types to match docs and it looks like there are several changes necessary.  I will update this PR if not merged by time finished.

treydock:
Added commit that should fully resolve #1031 as many updates to match sensu-go docs.  Updated a few doc strings as well as some defaults and a couple property changes.

sensu_check publish defaults to true to match behavior when creating checks using `sensuctl check create`
sensu_check stdin defaults to false
Add sensu_check silenced property
sensu_entity deregister defaults to false
sensu_handler timeout defaults to 60 when type is tcp
sensu_hook stdin defaults to false

treydock:
Added commit that updates unit tests to try and guard against unexpected default changes when possible.

PR #1034, opened 2019-01-01
Use Puppet SSL by default

treydock:
# Pull Request Checklist

## Description

By default the backend will be configured to use SSL and agents will communicate using SSL.  The certificates and CA used are from Puppet.

This adds several parameters.

**sensu**:
* ssl_dir
* ssl_dir_purge
* use_ssl
* ssl_ca_source
* ssl_add_ca_trust

**sensu::backend**:
* url_host - default changed to `certname`
* use_ssl - moved to `sensu` class
* ssl_cert_source
* ssl_key_source

**sensu::agent**:
* backends - used to generate `backend-url` config option

Added dependency on `puppet/trusted_ca` module.

## Motivation and Context

Puppet already has SSL certificates and would be good to default sensu-go to use a secure configuration.

## How Has This Been Tested?

Unit and beaker

## General


treydock:
Test failures I can't tell are issue with our dependencies or something broken at rubygems.org.  This PR did not modify `Gemfile` so thinking something is wrong with rubygems.org or travis-ci.

treydock:
I disabled SSL for vagrant environment.  I wasn't sure if we wanted to add puppetserver logic to Vagrant test environment.

treydock:
I wanted to use `camptocamp/openssl` for the `openssl` package but that module and `trusted_ca` module have conflicting resources.  I will try to fix `trusted_ca` module so we can use `openssl` module and not define the package in this module if that sounds like something we should do.

Also need to add the Puppet config facts in this PR to stdlib module.  Right now stdlib looks at Puppet config settings for a couple facts but I think their implementation doesn't work if `puppet.conf` doesn't have the value.  The method used in this PR will return puppet's default values if they are not defined in `puppet.conf`.

lib/facter/sensu_puppet_facts.rb
```diff
@@ -0,0 +1,33 @@
+require 'puppet'
+
+module PuppetFacts
```
ghoneycutt: does `PuppetFacts` come from puppet or from a gem?

treydock: It's defined in that file, I should have probably used `SensuPuppetFacts` to match file name.

ghoneycutt:
If we're going to SSL by default, then it should be that way in Vagrant, too.

lib/facter/sensu_puppet_facts.rb
ghoneycutt: Aha, let's do that to make it more apparent.

treydock:
The failures in tests appear to be specific to puppetserver which is odd given puppetserver is running on CentOS 7 for all OSes in the test matrix, so would expect it to fail for all and not just some.  I'm not having much luck reproducing the failures when running beaker locally.  Will investigate as just appears like puppetserver is getting timed out when starting.

Vagrantfile
```diff
@@ -30,7 +30,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.synced_folder ".", "/vagrant", type: "virtualbox"
 
   config.vm.provider :virtualbox do |vb|
-    vb.customize ["modifyvm", :id, "--memory", "512"]
+    vb.customize ["modifyvm", :id, "--memory", "1512"]
+    vb.cpus = 2
```
ghoneycutt: Why add CPU's?

manifests/ssl.pp
```diff
@@ -0,0 +1,37 @@
+# @summary Private class to manage sensu SSL resources
+#
```
ghoneycutt: I think it still needs `@api private`

manifests/repo.pp
```diff
@@ -1,4 +1,5 @@
-# private class
+# @summary Private class to manage sensu repository resources
+#
```
ghoneycutt: I think it still needs `@api private`

spec/acceptance/00_backend_spec.rb
```diff
@@ -5,7 +5,8 @@
   context 'default' do
     it 'should work without errors' do
       pp = <<-EOS
-      include ::sensu::backend
+      class { '::sensu': }
+      class { '::sensu::backend': }
```
ghoneycutt: curious why the change to resource syntax for the inclusion of the class.

types/backend_url.pp
```diff
@@ -0,0 +1,5 @@
+type Sensu::Backend_URL = Variant[
```
ghoneycutt: Check out this PR to stdlib as an example for writing spec tests for type aliases

https://github.com/puppetlabs/puppetlabs-stdlib/commit/6dad21e292436ac5b7d8113e81e572a0efc66782

ghoneycutt:
We should also have tests that show this works with puppet apply when setting use_ssl to false.

Vagrantfile
treydock: Had a few occurrences of startup timing out when testing and figured more CPUs would help given how resource intensive puppetserver can be.

spec/acceptance/00_backend_spec.rb
treydock: I wanted to verify the README style of examples would work.  If `class {}` works then we know `include` will work.

spec/acceptance/03_no_ssl_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'spec_helper_acceptance'
+
+describe 'sensu without SSL', unless: RSpec.configuration.sensu_cluster do
```
treydock: @ghoneycutt This is the tests where `use_ssl => false` is done using method documented in README.

treydock:
@ghoneycutt the part listed twice you mean the `sed` commands?  They are actually different.  One is `Xms` and the other is `Xmx`.

Vagrantfile
ghoneycutt: This can slow things down with virtualization as now there needs to be two free cpu's available at the same time. If your system only has one at that moment, then the hypervisor waits until both are free.

treydock: Removed cpus change.

manifests/repo.pp
treydock: Added `@api private` back

types/backend_url.pp
treydock: Added unit tests based on the stdlib example.

spec/type_aliases/backend_url_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    ].each do |value|
+      describe value.inspect do
+        it { is_expected.to allow_value(value) }
+      end
+    end
+  end
+
+  context 'disallow invalid values' do
+    [
+      'http://localhost:8081',
```
ghoneycutt: you can't have a HTTP backend?

tests/provision_basic_el.sh
```diff
@@ -31,4 +31,9 @@ puppet resource file /etc/puppetlabs/code/environments/production/modules/sensu
 
 # setup module dependencies
 puppet module install puppetlabs/stdlib --version 4.25.1
+puppet module install puppet/trusted_ca --version 2.0.0
+
+puppet resource host puppet ensure=present ip=192.168.52.10
+puppet resource host sensu-backend.example.com ensure=present ip=192.168.52.10
+puppet agent --test --tags no_such_tag
```
ghoneycutt: could you use `-t` instead of `--test`. Most people learn with `-t` as the name 'test' is vary misleading.

spec/type_aliases/backend_url_spec.rb
treydock: Not that I'm aware of, docs say use `ws://` or `wss://` (ssl).

tests/provision_basic_el.sh
treydock: Changed.

spec/type_aliases/backend_url_spec.rb
treydock: Confirmed with vagrant, have to use `ws://` or `wss://`, tried using `https://`:

```
Jan  3 01:02:51 el7-agent sensu-agent: {"component":"agent","error":"malformed ws or wss URL","level":"error","msg":"reconnection attempt failed","time":"2019-01-03T01:02:51Z"}
```

treydock:
I'll squash commits down once I get acceptance tests working.  I'm not liking how puppetserver is behaving and causing only some of the tests to fail.  It makes no sense because all OSes in matrix use exact same docker image (centos:7) for puppetserver and only the same 3 seem to fail.  A really ugly fallback solution could be to generate the puppet certs with Vagrant, store them in the repo and copy into test docker containers in same location that Puppet would use.

treydock:
Updated to store Puppet SSL certs in `tests/ssl` directory.  The backend cert has a very long Subject Alt Name that allows it to work with both Vagrant and docker.  I created copies of the private key and certs for backend so that the default host certname would pick up the correct certificate that has SAN.

I also changed how acceptance test roles are looked up because ran into issues where only one of the 3 backend cluster nodes for cluster tests had `sensu_backend` role and this was incorrect and previous role lookup only allowed role to exist on one node in a test which didn't seem right when cluster had 3 that needed same role.

treydock:
@ghoneycutt tests are now passing, this is good to merge if everything looks good.

spec/unit/facter/sensu_puppet_facts_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    Puppet[:codedir] = '/tmp'
+    subject.add_facts
+  end
+  after(:each) do
+    Facter.clear
+    Facter.clear_messages
+  end
+
+  it 'should have puppet_hostcert defined' do
+    Puppet[:hostcert] = '/dne/host.pem'
```
ghoneycutt: what's dne?

types/backend_url.pp
ghoneycutt: It can be useful to comment regex's with what they should match, such as  the following. This makes it easy to grok what's happening.

`Sensu Backend URL's use ws:// or wss:// or can be specified with the host only and no protocol. They must always have the port specified.`

spec/unit/facter/sensu_puppet_facts_spec.rb
treydock: Does Not Exist.  Just passing in something that doesn't exist for purposes of unit testing.

types/backend_url.pp
treydock: So `ws://` or `wss://` are actually always required but made optional because they will be automatically added by `sensu::agent` class based on value of `use_ssl`.  I am making note of that in comment.

PR #1036, opened 2019-01-05
Fix sensu_check check_hooks property

treydock:
# Pull Request Checklist

## Description

Fix check_hooks to ensure numeric range accepted includes 0
Munge integer keys to strings as sensuctl does not allow integer keys for check_hooks

## Motivation and Context

The docs for sensu-go show examples that `0` is valid type for `check_hooks` and confirmed this with vagrant.
Given Integer and not numeric strings also results in errors so munge all keys to strings:

```
[root@sensu-backend ~]# cat check.json
{
  "command": "check-http.rb",
  "handlers": [
    "email"
  ],
  "high_flap_threshold": 0,
  "interval": 60,
  "low_flap_threshold": 0,
  "publish": false,
  "runtime_assets": null,
  "subscriptions": [
    "demo"
  ],
  "proxy_entity_name": "",
  "check_hooks": [
    {
      '0': [
        "test"
      ]
    }
  ],
  "stdin": false,
  "subdue": null,
  "ttl": 0,
  "timeout": 0,
  "round_robin": false,
  "output_metric_format": "",
  "output_metric_handlers": null,
  "env_vars": null,
  "metadata": {
    "name": "test",
    "namespace": "default"
  }
}
[root@sensu-backend ~]# sensuctl create -f check.json
resource 0: invalid character '0' looking for beginning of object key string
resource 1: invalid character '0' looking for beginning of object key string
resource 2: invalid character '0' looking for beginning of object key string
resource 3: invalid character '0' looking for beginning of object key string
resource 4: invalid character '0' looking for beginning of object key string
resource 5: invalid character '0' looking for beginning of object key string
resource 6: invalid character '0' looking for beginning of object key string
resource 7: invalid character '0' looking for beginning of object key string
resource 8: invalid character '0' looking for beginning of object key string
resource 9: invalid character '0' looking for beginning of object key string
resource 10: invalid character '0' looking for beginning of object key string
Error: too many errors
```

### DIFF
--- a/lib/puppet/type/sensu_check.rb
+++ b/lib/puppet/type/sensu_check.rb
@@ -122,7 +122,7 @@ DESC
       type_valid = false
       if ['ok','warning','critical','unknown','non-zero'].include?(type)
         type_valid = true
-      elsif type.to_s =~ /^\d+$/ && type.to_i.between?(1,255)
+      elsif type.to_s =~ /^\d+$/ && type.to_i.between?(0,255)
         type_valid = true
       end
       if ! type_valid
@@ -131,6 +131,11 @@ DESC
       if ! hooks.is_a?(Array)
         raise ArgumentError, "check_hooks hooks must be an Array"
       end
+    end
+    munge do |value|
+      type = value.keys[0]
+      hooks = value[type]
+      { type.to_s => hooks }
     end
   end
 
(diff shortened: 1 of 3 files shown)


nikkiki:
The type should always be a string. In addition to “0”-“255”, hook types can also be expressed as “critical”, “non-zero” etc. so for that reason they are strings. I’ll file a docs issue to correct this!

ghoneycutt:
@nikkiki Shouldn't sensu-go support the following though? It seems like a bug that it's type handling would exclude integers and only allow strings.

Integers 0-255
Integers represented as strings "0" - "255"
Array of strings [ "critical", "non-zero", "etc" ]

PR #1040, opened 2019-01-08
Add default sensu resources

treydock:
# Pull Request Checklist

## Description

Add resources that are present on fresh install of sensu-go.

## Motivation and Context

If someone does aggressive resource purging with something like `resources { 'sensu_user': purge => true }` they would break their install unless the default agent user is defined.  The same goes for other resources defined, some would break sensu-go if missing others may just cause loss of functionality like missing roles.

### DIFF
--- a/manifests/backend/resources.pp
+++ b/manifests/backend/resources.pp
@@ -1,0 +1,149 @@
+# @summary Default sensu resources
+# @api private
+#
+class sensu::backend::resources {
+  sensu_namespace { 'default':
+    ensure => 'present',
+  }
+
+  sensu_user { 'agent':
+    ensure   => 'present',
+    disabled => false,
+    password => 'P@ssw0rd!',
+    groups   => ['system:agents'],
+  }
+
+  sensu_cluster_role { 'admin':
+    ensure => 'present',
+    rules  => [
+      {
+        'verbs'     => ['*'],
+        'resources' => [
+          'assets',
+          'checks',
+          'entities',
+          'extensions',
+          'events',
+          'filters',
+          'handlers',
+          'hooks',
+          'mutators',
+          'silenced',
+          'roles',
+          'rolebindings',
+        ],
+      },
+      {
+        'verbs'     => ['get', 'list'],
+        'resources' => ['namespaces'],
+      },
+    ],
+  }
+  sensu_cluster_role { 'cluster-admin':
+    ensure => 'present',
+    rules  => [
+      {
+        'verbs'     => ['*'],
+        'resources' => ['*'],
+      },
+    ],
+  }
+  sensu_cluster_role { 'edit':
+    ensure => 'present',
+    rules  => [
+      {
+        'verbs'     => ['*'],
+        'resources' => [
+          'assets',
+          'checks',
+          'entities',
+          'extensions',
+          'events',
+          'filters',
+          'handlers',
+          'hooks',
+          'mutators',
+          'silenced',
+        ],
+      },
+      {
+        'verbs'     => ['get', 'list'],
+        'resources' => ['namespaces'],
+      },
+    ],
+  }
+  sensu_cluster_role { 'system:agent':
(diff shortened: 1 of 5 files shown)


treydock:
I ran `puppet resources <resource>` and dumped contents to file in /vagrant using fresh install in Vagrant.  The indentation issues are a result of odd indentation from `puppet resources` output.  The only change I had to make to resource dumps was adding password for agent user because Puppet has no way to query the password but one is required.  Verified with `curl` that the password of default install is indeed `P@ssw0rd!`.

manifests/backend/resources.pp
```diff
@@ -0,0 +1,149 @@
...
+      ],
+    },
+    ],
+  }
+
+  sensu_cluster_role_binding { 'cluster-admin':
+    ensure   => 'present',
+    role_ref => 'cluster-admin',
+    subjects => [
+    {
```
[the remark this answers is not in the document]
treydock: Fixed.

spec/classes/backend_resources_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+
+describe 'sensu::backend::resources', :type => :class do
+  on_supported_os({facterversion: '3.8.0'}).each do |os, facts|
+    context "on #{os}" do
+      let(:facts) { facts }
+      context 'with default values for all parameters' do
+        it { should compile }
+        it { should have_sensu_namespace_resource_count(1) }
```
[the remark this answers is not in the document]
treydock: I left the count tests so that there is validation on how many defaults we're adding to avoid rogue additions.  Updated with more complete unit tests.

treydock:
Added commit that allows easily changing agent password.  The password of agent user has to match value given in `agent.yml` so seemed like something we should make easy to change.

spec/classes/agent_spec.rb
```diff
@@ -58,11 +58,12 @@
...
           should contain_file('sensu_agent_config').with({
-            'ensure'  => 'file',
-            'path'    => '/etc/sensu/agent.yml',
-            'content' => agent_content,
-            'require' => 'Package[sensu-go-agent]',
-            'notify'  => 'Service[sensu-agent]',
+            'ensure'    => 'file',
+            'path'      => '/etc/sensu/agent.yml',
+            'content'   => agent_content,
+            'show_diff' => 'false',
```
ghoneycutt: why the `show_diff` set to `false`. This is very useful to see what is changing.

spec/classes/backend_resources_spec.rb
```diff
@@ -8,13 +8,14 @@
         it { should compile }
         it { should have_sensu_namespace_resource_count(1) }
         it { should contain_sensu_namespace('default').with_ensure('present') }
-        it { should have_sensu_user_resource_count(1) }
+        it { should have_sensu_user_resource_count(2) }
```
ghoneycutt: What are the two?

treydock: One is `agent` and other is `admin` from included `sensu::backend` class.

spec/classes/agent_spec.rb
ghoneycutt: @treydock could you please comment on this? The `show_diff` question is the only one outstanding before merge.

treydock: Because there is a password config option in `agent.yml`.  A default install of sensu-go adds the agent user with same default password as admin user and thought if someone is going to change admin user they might be also be changing agent user password which is now possible with this pull request.  The password for agent user has to be updated in `agent.yml` for all agents if agent user password is changed.

ghoneycutt: Please remove from here and in another PR let's make it a param that can be set and apply it to all config files.

treydock: Removed `show_diff` change.

PR #1041, opened 2019-01-08
Document name handling for sensu_silenced

### LINKED ISSUE #1015: Rework sensu_silenced name handling
The name passed like this:
```puppet
      sensu_silenced { 'test':
        ensure => 'present',
        subscription => 'entity:sensu_agent',
      }
```

Will be replaced in `metadata` with `entity:sensu_agent:*`.  The name passed during creation isn't used, and based on docs name is a `#{subscription}:#{check}`.  To avoid confusion and inconsistent behavior the name of a `sensu_silenced` resource should match the convention that one would see when looking at things through sensuctl.

treydock:
# Pull Request Checklist

## Description

Update docs for `sensu_silenced` to describe when name is used and when it's not used.

## Related Issue

Fixes #1015 

## Motivation and Context

Try to document that the name in Puppet may not match sensuctl output.

### DIFF
--- a/lib/puppet/type/sensu_silenced.rb
+++ b/lib/puppet/type/sensu_silenced.rb
@@ -7,7 +7,12 @@ Puppet::Type.newtype(:sensu_silenced) do
   desc <<-DESC
 @summary Manages Sensu silencing
 
-The name of `sensu_silenced` can be used to define `check` and `subscription`.
+The name of a `sensu_silenced` resource may not match the name returned by sensuctl.
+The name from sensuctl will take the form of `subscription:check`.
+If you wish to have a `sensu_silenced` resource name match sensuctl then define the name
+using the `subscription:check` format and do not define `subscription` or `check` properties.
+
+The `subscription` and `check` properties take precedence over value in the name if name takes the form `subscription:check`.
 
 @example Create a silencing for all checks with subscription entity:sensu_agent
   sensu_silenced { 'test':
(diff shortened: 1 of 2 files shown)


lib/puppet/type/sensu_silenced.rb
```diff
@@ -8,6 +8,10 @@
 @summary Manages Sensu silencing
 
 The name of `sensu_silenced` can be used to define `check` and `subscription`.
+If the name used to define a `sensu_silenced` resource does not take the form
```
ghoneycutt: This sentence is very difficult to understand, especially for someone where english is not their first language. It starts with a negation which is also difficult. Could you please reword this documentation. Smaller sentences are easier to understand than big sentences with lots of logic in them.

treydock: What about now?

lib/puppet/type/sensu_silenced.rb
```diff
@@ -7,7 +7,10 @@
   desc <<-DESC
 @summary Manages Sensu silencing
 
-The name of `sensu_silenced` can be used to define `check` and `subscription`.
+The name of a `sensu_silenced` resource may not match the name returned by sensuctl.
+The name from sensuctl will take the form of `subscription:check`.
+If you wish to have a `sensu_silenced` resource name match sensuctl then define the name
+using the `subscription:check` format and do not define `subscription` or `check` properties.
```
ghoneycutt: what happens if you use subscription:check and specify subscription or check parameters?

treydock:
@ghoneycutt documented how `subscription` and `check` properties take precedence over `name` and added several tests to show this works.

PR #1042, opened 2019-01-10
Add show_diff parameter

treydock:
# Pull Request Checklist

## Description

Add `show_diff` that defaults to `false`. Diffs can be re-enabled globally via `sensu` class or for a specific role in either `sensu::agent` or `sensu::backend`.

## Motivation and Context

The `agent.yml` file can contain a password and thus by default a diff should be suppressed.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -28,6 +28,8 @@
 #   The protocol prefix of `ws://` or `wss://` are optional and will be determined
 #   based on `sensu::use_ssl` parameter by default.
 #   Passing `backend-url` as part of `config_hash` takes precedence.
+# @param show_diff
+#   Sets show_diff parameter for agent.yml configuration file
 #
 class sensu::agent (
   Optional[String] $version = undef,
@@ -37,6 +39,7 @@ class sensu::agent (
   Boolean $service_enable = true,
   Hash $config_hash = {},
   Array[Sensu::Backend_URL] $backends = ['localhost:8081'],
+  Boolean $show_diff = true,
 ) {
 
   include ::sensu
@@ -73,11 +76,12 @@ class sensu::agent (
   }
 
   file { 'sensu_agent_config':
-    ensure  => 'file',
-    path    => "${etc_dir}/agent.yml",
-    content => to_yaml($config),
-    require => Package['sensu-go-agent'],
-    notify  => Service['sensu-agent'],
+    ensure    => 'file',
+    path      => "${etc_dir}/agent.yml",
+    content   => to_yaml($config),
+    show_diff => $show_diff,
+    require   => Package['sensu-go-agent'],
+    notify    => Service['sensu-agent'],
   }
 
   service { 'sensu-agent':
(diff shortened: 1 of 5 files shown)


manifests/agent.pp
```diff
@@ -37,13 +40,15 @@
   Boolean $service_enable = true,
   Hash $config_hash = {},
   Array[Sensu::Backend_URL] $backends = ['localhost:8081'],
+  Optional[Boolean] $show_diff = undef,
```
ghoneycutt: The `Optional` makes this more complicated than it should be. Let's just use `Boolean` with the default set to `true`, which is the default anyhow for the file resource.

https://puppet.com/docs/puppet/6.1/types/file.html#file-attribute-show_diff

manifests/backend.pp
```diff
@@ -52,6 +55,7 @@
   String $ssl_key_source = $facts['puppet_hostprivkey'],
   String $password = 'P@ssw0rd!',
   Optional[String] $old_password = undef,
+  Optional[Boolean] $show_diff = undef,
```
ghoneycutt: see note above

manifests/init.pp
```diff
@@ -40,6 +40,9 @@
 #   Boolean that determines if SSL CA should be added
 #   to the system's trust store
 #
+# @param show_diff
```
ghoneycutt: where does this get used? Seems like agent and backend have their own settings.

manifests/agent.pp
treydock: Updated to default to `true` and documented that changing agent password should set `show_diff` to `false`.

manifests/backend.pp
treydock: Fixed.

spec/classes/agent_spec.rb
```diff
@@ -58,11 +58,12 @@
 
         it {
           should contain_file('sensu_agent_config').with({
-            'ensure'  => 'file',
-            'path'    => '/etc/sensu/agent.yml',
-            'content' => agent_content,
-            'require' => 'Package[sensu-go-agent]',
-            'notify'  => 'Service[sensu-agent]',
+            'ensure'    => 'file',
```
ghoneycutt: there should be a test that demonstrates when `show_diff` is set to `false`.

spec/classes/backend_spec.rb
```diff
@@ -108,11 +108,12 @@
 
         it {
           should contain_file('sensu_backend_config').with({
-            'ensure'  => 'file',
-            'path'    => '/etc/sensu/backend.yml',
-            'content' => backend_content,
-            'require' => 'Package[sensu-go-backend]',
-            'notify'  => 'Service[sensu-backend]',
+            'ensure'    => 'file',
```
ghoneycutt: there should be a test that demonstrates when `show_diff` is set to `false`.

spec/classes/agent_spec.rb
treydock: Test added

spec/classes/backend_spec.rb
treydock: Test added

PR #1044, opened 2019-01-11
Fix Amazon 2 to use EL7 repos

treydock:
# Pull Request Checklist

## Description

Use EL7 repo for Amazon 2.

## Motivation and Context

Amazon 2 is systemd so should use EL7 RPMs.  The logic matches logic used in sensu2 branch.


ghoneycutt:
Released in v2.62.0

PR #1053, opened 2019-01-22
Do not enable rabbitmq ssl for empty cert or key

### LINKED ISSUE #1050: TripleO and Sensu without ssl, ssl still enabled in rabbitmq.json
## Description of problem

- What did you do?
Use OpenStack TripleO+OpsTools to deploy OpenStack, with monitoring enabled but no SSL
https://github.com/openstack/tripleo-heat-templates/blob/stable/pike/environments/monitoring-environment.yaml

- What happened?
Sensu agents are not able to communicate with rabbitmq

Sensu agents to be able to communicate

Deploy OpenStack using TripleO with monitoring but set `MonitoringRabbitUseSSL` to false(or not set at all as its default value is false)

Sensu agent logs
```
{"timestamp":"2019-01-15T15:48:01.357967-0800","level":"warn","message":"transport connection error","reason":"tcp connection lost"}
```

- Your OS: CentOS 7.4.1708
- Your Ruby version: ruby-2.0.0.648-33.el7_4.x86_64
- Your version of Puppet: puppet-4.8.2-1.el7.noarch
- Your version of Sensu: sensu-0.27.0-1.el7.noarch
- Your version of this module: puppet-sensu-2.30.1-0.20171102090215.c99f0d2.el7.centos.noarch
- Pike version of TripleO/OpenStack

Removing `ssl{}` from sensu's rabbitmq config fixed it and the agents are able to communicate with the sensu server. I dont know much about ruby, but my hunch is this bit of code https://github.com/sensu/sensu-puppet/blob/master/lib/puppet/provider/sensu_rabbitmq_config/json.rb#L35-L55 
It looks like it only checks the existence of the key ssl in the hiera data, not it's value? So even though we are setting that to false via TripleO, somehow the sensu puppet module is putting `ssl{}` into the config, enabling ssl. But like I mention, I dont know ruby, so I could be wrong on that.

treydock:
# Pull Request Checklist

## Description

If `sensu:: rabbitmq_ssl_private_key` and `sensu:: rabbitmq_ssl_cert_chain` are empty, do not enable rabbitmq ssl if `sensu::rabbitmq_ssl` is `false`.

## Related Issue

Resolves #1050 

## Motivation and Context

This treats the module parameter values of empty strings for key and cert the same as if the `sensu_rabbitmq_config` type is used directly.  For that type empty cert and key will not enable SSL as those are the defaults.

## General

### DIFF
--- a/manifests/rabbitmq/config.pp
+++ b/manifests/rabbitmq/config.pp
@@ -90,11 +90,14 @@ class sensu::rabbitmq::config {
     } else {
       $ssl_private_key = $::sensu::rabbitmq_ssl_private_key
     }
-
-    $enable_ssl = true
   } else {
     $ssl_cert_chain = undef
     $ssl_private_key = undef
+  }
+
+  if ($ssl_cert_chain and $ssl_cert_chain != '') or ($ssl_private_key and $ssl_private_key != '') {
+    $enable_ssl = true
+  } else {
     $enable_ssl = $::sensu::rabbitmq_ssl
   }
 
--- a/spec/classes/sensu_rabbitmq_spec.rb
+++ b/spec/classes/sensu_rabbitmq_spec.rb
@@ -63,6 +63,7 @@ describe 'sensu', :type => :class do
   context 'rabbitmq config' do
     context 'no ssl (default)' do
       it { should contain_sensu_rabbitmq_config('hostname.domain.com').with(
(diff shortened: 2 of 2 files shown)


treydock:
Took existing `sensu-server` Vagrant VM and re-ran provision and no changes occurred so default behavior is retained.

PR #1054, opened 2019-01-24
Use sensu-go test-creds subcommand to test user passwords

treydock:
# Pull Request Checklist

## Description

Replace temporary net/http API call for testing user passwords with new `sensuctl user test-creds` subcommand.

## Related Issue

Fixes couple check boxes for #901 

## Motivation and Context

sensu-go 5.1.1 added `sensuctl user test-creds` subcommand to test passwords.  This is much simpler than temporary net/http approach and much more likely to be supported on future versions of sensu-go.

### DIFF
--- a/lib/puppet/provider/sensu_user/sensuctl.rb
+++ b/lib/puppet/provider/sensu_user/sensuctl.rb
@@ -97,6 +97,9 @@ Puppet::Type.type(:sensu_user).provide(:sensuctl, :parent => Puppet::Provider::S
         if ! resource[:old_password]
           fail("old_password is manditory when changing a password")
         end
+        if ! password_insync?(resource[:name], resource[:old_password])
+          fail("old_password given for #{resource[:name]} is incorrect")
+        end
         sensuctl('user', 'change-password', resource[:name], '--current-password', resource[:old_password], '--new-password', @property_flush[:password])
         if resource[:configure] == :true
           sensuctl('configure', '-n', '--url', resource[:configure_url], '--username', resource[:name], '--password', @property_flush[:password])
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/sensu_user/sensuctl.rb
```diff
@@ -52,32 +52,9 @@ def password_insync?(user, password)
...
-      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
-    end
-    request = Net::HTTP::Get.new('/auth')
-    request.add_field("Accept", "application/json")
-    request.basic_auth(user, password)
-    response = http.request(request)
-    Puppet.debug("GET /auth returned #{response.code}")
-    response.kind_of?(Net::HTTPSuccess)
+    ret = execute(['sensuctl', 'user', 'test-creds', user, '--password', password], failonfail: false)
+    exitstatus = ret.exitstatus
```
ghoneycutt: why get the status then set it to 0?

treydock:
Failures for acceptance I can't reproduce all of them locally.  The errors are unrelated to this change but are failing because getting "connection refused" when running sensuctl.  I'm guessing the backend is crashing.  The unit file for sensu-backend has `Restart=always` so more than likely if sensu-backend is crashing systemd is restarting the daemon before Puppet has a chance to check the service.

lib/puppet/provider/sensu_user/sensuctl.rb
ghoneycutt: do we have a test that shows what happens when you supply the wrong password and try to change passwords?

treydock: It would have to be an acceptance test, unit testing this wouldn't really prove much since when changing passwords we just pass old and new with no validation.  We could, in theory, validate the old password first using same function.

treydock: Added commit that tests the `old_password` before attempting to change the password.  Also added corresponding unit and acceptance tests.

treydock:
Had to add commits that fix tests to pass with latest sensu-go by removing resolve of keepalive events and instead resolving failing check.

PR #1057, opened 2019-01-26
Sensu2 agent ssl

treydock:
# Pull Request Checklist

## Description

Update agent SSL to define trusted-ca-file when SSL is enabled
The system trusted CAs are only updated for backends in order to support SSL with sensuctl.

## Motivation and Context

Agent supported `trusted-ca-file` like backend so use that when SSL is enabled instead of adding the CA to system trust store.  Now the only thing needing system trust store updates is sensuctl.  Once https://github.com/sensu/sensu-go/pull/2627 is released we can remove the dependency on trusted_ca module.

## How Has This Been Tested?

beaker and unit

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -45,15 +45,19 @@ class sensu::agent (
   include ::sensu
 
   $etc_dir = $::sensu::etc_dir
+  $ssl_dir = $::sensu::ssl_dir
   $use_ssl = $::sensu::use_ssl
   $_version = pick($version, $::sensu::version)
 
   if $use_ssl {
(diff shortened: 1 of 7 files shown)


treydock:
Tests will likely not pass until #1054 is merged and this rebased.

treydock:
The existing acceptance tests cover the behavior since we default to SSL and also have acceptance tests for no SSL.  The only real change in behavior is that we set a new config option in `agent.yml` instead of adding CA to system trust store and that is tested to be successful already by starting agent and ensuring it registered with backend which doesn't happen if SSL isn't working.

PR #1062, opened 2019-02-09
Fix sensu facts to work with sensu-go 5.2.0 output

treydock:
# Pull Request Checklist

## Description

Fix sensu facts to handle output for version subcommands in sensu-go 5.2.0.

## Motivation and Context

The output format for latest sensu-go release, 5.2.0, changed a bit and caused sensu facts `built` value to not be returned.  The fix is not assume built value is wrapped in quotes as 5.2.0 removed quotes around the timestamp.

## How Has This Been Tested?

Added unit tests.

### DIFF
--- a/lib/facter/sensu_facts.rb
+++ b/lib/facter/sensu_facts.rb
@@ -19,7 +19,8 @@ module SensuFacts
       build = $1
     end
     built = nil
-    if output =~ /built '(.*)'$/
+    # Match value that is optionally wrapped in single quotes
+    if output =~ /built (?:')?([^']+)(?:')?$/
       built = $1
     end
     return version, build, built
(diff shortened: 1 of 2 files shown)


lib/facter/sensu_facts.rb
```diff
@@ -19,7 +19,7 @@ def self.get_version_info(cmd)
       build = $1
     end
     built = nil
-    if output =~ /built '(.*)'$/
+    if output =~ /built (?:')?([^']+)(?:')?$/
```
ghoneycutt: Could you please add a comment that describes this regex

treydock: Added comment about matching optional single quotes.

PR #1065, opened 2019-02-10
Support trusting CA cert with sensuctl

treydock:
# Pull Request Checklist

## Description

Add `trusted_ca_file` property to `sensu_configure` to support sensuctl using sensu CA cert without adding the CA cert to system trusted CAs.  Remove dependency on trusted_ca module.

## Motivation and Context

sensu-go 5.2.0 added support for `--trusted-ca-file` with sensuctl and this PR takes advantage of that new feature.


treydock:
Added commit to set `api-url` in backend config because the docs for "Securing Sensu" now list that api-url must be changed when doing SSL.  https://docs.sensu.io/sensu-go/5.2/guides/securing-sensu/#securing-the-api-and-the-dashboard

Issue #1064, opened 2019-02-09
Support sensu-go enterprise features


treydock:
The documentation for sensu-go 5.2 outlines two enterprise features that need to be supported.

- [x] License
- [x]  Auth

The auth commands require the license be added first which can be handled with autorequires.

ghoneycutt:
@treydock Do we have a PR for Auth? Seems #1066 was for License.

treydock:
#1066 also had auth:

https://github.com/sensu/sensu-puppet/blob/sensu2/lib/puppet/type/sensu_ldap_auth.rb

ghoneycutt:
It mentioned that it only partially closed this issue. If we're all good, close this.

treydock:
Ya initial commit was just license file but later commits added auth.  Closing.

PR #1066, opened 2019-02-10
sensu-go enterprise support

treydock:
# Pull Request Checklist

## Description

So far this just adds support for adding sensu-go enterprise license.

## Related Issue

Partial fix for #1064

## Motivation and Context

This module should help support sensu-go enterprise features.

## How Has This Been Tested?

Unit only.

## Features needed

- [x] Add license
- [x] Support `sensuctl auth`
- [x] Update README
- [x] Add acceptance tests that will only run if a license file is provided


README.md
```diff
@@ -197,6 +198,16 @@ class { 'sensu':
 }
 ```
 
+### Enterprise Support
+
+In order to activate enterprise support the license file needs to be added:
+
+```puppet
+class { 'sensu::backend':
+  license_source => 'puppet:///modules/site_sensu/license.json',
```
ghoneycutt: This is the type of code you would have in a profile, so perhaps

`'puppet:///modules/profile/sensu/license.json'`

spec/spec_helper_acceptance.rb
```diff
@@ -11,8 +11,17 @@
 RSpec.configure do |c|
   c.add_setting :sensu_full, default: false
   c.add_setting :sensu_cluster, default: false
+  c.add_setting :sensu_enterprise_file, default: nil
+  c.add_setting :sensu_test_enterprise, default: false
   c.sensu_full = (ENV['BEAKER_sensu_full'] == 'yes' || ENV['BEAKER_sensu_full'] == 'true')
   c.sensu_cluster = (ENV['BEAKER_sensu_cluster'] == 'yes' || ENV['BEAKER_sensu_cluster'] == 'true')
+  if ENV['SENSU_ENTERPRISE_FILE']
```
ghoneycutt: Great approach for testing!

ghoneycutt:
Looks great so far!

README.md
treydock: Also updated other places where referenced `site_sensu` and replaced with `profile/sensu`.

treydock:
Added acceptance tests to verify adding a license file works.

manifests/backend.pp
```diff
@@ -64,8 +71,14 @@
   Optional[String] $agent_old_password = undef,
   Boolean $include_default_resources = true,
   Boolean $show_diff = true,
+  Optional[String] $license_source = undef,
+  Optional[String] $license_content = undef,
 ) {
 
+  if $license_source and $license_content {
+    fail('sensu::backend: Do not define both license_source and license_content')
```
ghoneycutt: I believe the error will already report that it came from the sensu::backend class.

treydock: Maybe if run from master but with `puppet apply` it does not when I removed reference to `sensu::backend: `.

```
Error: Evaluation Error: Error while evaluating a Function Call, Do not define both license_source and license_content (file: /etc/puppetlabs/code/environments/production/modules/sensu/manifests/backend.pp, line: 79, column: 5) on node sensu-backend.example.com
```

.travis.yml
```diff
@@ -31,121 +32,121 @@ matrix:
...
     services: docker
-    env: BEAKER_set="amazonlinux-2" BEAKER_PUPPET_COLLECTION=puppet5
+    env: BEAKER_set="amazonlinux-2" BEAKER_PUPPET_COLLECTION=puppet5 SENSU_ENTERPRISE_FILE=tests/sensu_license.json
     bundler_args:
     script: bundle exec rake beaker
   - rvm: 2.5.3
     sudo: required
     services: docker
-    env: BEAKER_set="amazonlinux-2" BEAKER_PUPPET_COLLECTION=puppet6
+    env: BEAKER_set="amazonlinux-2" BEAKER_PUPPET_COLLECTION=puppet6 SENSU_ENTERPRISE_FILE=tests/sensu_license.json
```
ghoneycutt: in the spec_helper or related place could we just specify that if the environment variable `SENSU_ENTERPRISE_FILE` is set to use it and if not set it to the default `tests/sensu_license.json`. That would DRY up the configuration.

treydock: Updated with that change.

spec/unit/provider/sensuctl_spec.rb
```diff
@@ -26,38 +26,38 @@
...
     it 'should have JSON with spec' do
-      subject.sensuctl_create('test', {name: 'test'}, {foo: 'bar'})
+      subject.sensuctl_create('Test', {name: 'test'}, {foo: 'bar'})
       j = JSON.parse(File.read(@tmp.path))
       expect(j['spec']).to eq({'foo' => 'bar'})
     end
 
     it 'should have JSON with only valid keys' do
-      subject.sensuctl_create('test', {name: 'test'}, {foo: 'bar'})
+      subject.sensuctl_create('Test', {name: 'test'}, {foo: 'bar'})
```
ghoneycutt: Are all these capitalization changes for style or did something change with Sensu to make it necessary?

treydock: Previously the function `sensuctl_create` was doing the capitalization but that breaks adding LDAP auth as `ldap` is the `type` and `Ldap` does not work.  Rather than add logic to disable the capitalization I just moved the logic to each provider so each provider passes the `type` that's required.

treydock: The actual JSON passed to `sensuctl create -f` hasn't changed, just moved the logic to each provider rather than assuming the logic in shared function.

treydock:
Added commit that adds type/provider for adding LDAP auth which is an enterprise feature.

treydock:
Removed WIP, all enterprise features are accounted for in this PR.

ghoneycutt:
Is this using the real license?

treydock:
Yep, I encrypted the real license once I verified the encryption keys and file didn't get exposed during a run using a fake license.

PR #1068, opened 2019-02-11
Sensu2: allow using internal repo for real

### LINKED ISSUE #1069: Sensu2: allow using internal repo work
## Description of problem

- I have set on hiera `sensu::manag_repo: false`
- The module fails to intsall, because it has `require` set against `sensu::repo`
- skip the code inside `sensu::repo` if false is set

- Your OS: CentOS / Ubuntu
- Your Ruby version: this is Sensu-Go
- Your version of Puppet: 6.2
- Your version of Sensu: 5.2
- Your version of this module: sensu2 branch, pulling the latest updates

please checl this proposed solution: PR #1068

maxadamo:
# Pull Request Checklist

please allow to use `manage_repo` set to `false`

## Description
I have set in hiera:
```
 sensu::manag_repo: false
```
but it fails, since `sensu::repo` has several `require` dependency set in the manifests. 

## Related Issue

Fixes #1069

## Motivation and Context

Cannot use my internal repository

## How Has This Been Tested?

it should be trivial enough :) 

## General

skips `repo` if any2bool($manage_repo) is not true


manifests/repo.pp
```diff
@@ -1,49 +1,53 @@
...
+class sensu::repo (
+  $manage_repo = $sensu::manage_repo
+) inherits sensu {
 
-  case $facts['os']['family'] {
-    'RedHat': {
-      if $facts['os']['name'] == 'Amazon' {
-        if $facts['os']['release']['major'] =~ /^201\d$/ {
-          $repo_release = '6'
+  if any2bool($manage_repo) {
```
treydock: Don't need `any2bool` since `$manage_repo` should have data type `Boolean`.

manifests/repo.pp
```diff
@@ -1,49 +1,53 @@
-# private class
-class sensu::repo {
+# @summary Private class to manage sensu repository resources
+# @api private
+#
+class sensu::repo (
+  $manage_repo = $sensu::manage_repo
```
treydock: I'd recommend moving the variable outside parameter list.

manifests/repo.pp
```diff
@@ -1,49 +1,53 @@
...
+) inherits sensu {
```
treydock: I'd recommend not inheriting sensu but instead doing `include ::sensu` and then pulling in value of `manage_repo` from that class so the value has to be changed in `sensu` class.

manifests/repo.pp
maxadamo: I did it, and now it's more close to the coding style of the rest of the module

manifests/repo.pp
maxadamo: it makes sense: remove any2bool, since it's always bool

manifests/repo.pp
maxadamo: done

manifests/repo.pp
maxadamo: done

manifests/repo.pp
maxadamo: done

maxadamo:
please note, that now I have also tested the code. 
The commit that I did earlier it was like the saying on the t-shirt: "I only test in production" :)

ghoneycutt:
Thanks! Replaced by https://github.com/sensu/sensu-puppet/pull/1070

PR #1058, opened 2019-01-29
Allow overriding default location for sensu checks

ndelic0:
# Pull Request Checklist

## Related Issue

Fixes # .

## General

- [ ] Update `README.md` with any necessary configuration snippets

- [ ] New parameters are documented

- [ ] New parameters have tests

- [ ] Tests pass - `bundle exec rake validate lint spec`

PR #1071, opened 2019-02-12
Remove hardcoded /etc/sensu

treydock:
# Pull Request Checklist

## Description

Several places in code did not respect changes to `sensu::etc_dir`, this change removes hardcoded `/etc/sensu` and `/etc/sensu/conf.d` from several locations.

## Related Issue

Replaces #1058 

## Motivation and Context

Support changing `sensu::etc_dir` and other resources doing the right thing.

## How Has This Been Tested?

Existing unit tests verify these changes don't change behavior but don't actually verify that changes to `sensu::etc_dir` do what the changes intended, those will be added.

### DIFF
--- a/manifests/check.pp
+++ b/manifests/check.pp
@@ -132,6 +132,8 @@ define sensu::check (
   Variant[Undef,Enum['absent'],Hash]    $hooks = undef,
 ) {
 
+  include ::sensu
+
   if $ensure == 'present' and !$command {
     fail("sensu::check{${name}}: a command must be given when ensure is present")
   }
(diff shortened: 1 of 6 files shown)


ghoneycutt:
@treydock are there other tests you are going to add or is this ready?

treydock:
@ghoneycutt I'm content with existing tests because they verify that these changes didn't mess with default behavior.  What we're not testing is what happens when `sensu::etc_dir` is changed, that behavior is now different as in the past some resources didn't respect `sensu::etc_dir` and now they do.  I'll add tests since this is a bug fix that fixes previously broken behavior.

PR #1081, opened 2019-02-15
(GH-1080) Fixes "missing custom" in sensu_enterprise_dashboard_config

### LINKED ISSUE #1080: Puppet 6.2 fails to apply sensu config, failing with: "Error 500 on SERVER: Server Error: no parameter named 'custom' (file: /etc/puppetlabs/code/environments/ta _dit_production/modules/sensu/manifests/enterprise/dashboard.pp, line: 56) on Sensu_enterprise_dashboard_config"
## Description of problem

- What did you do?
Tried to apply existing sensu config after an update to Puppet 6.2
- What happened?
Update did not apply, failing with: 
`Error` 500 on SERVER: Server Error: no parameter named 'custom' (file: /etc/puppetlabs/code/environments/ta
_dit_production/modules/sensu/manifests/enterprise/dashboard.pp, line: 56) on Sensu_enterprise_dashboard_config`
Update should apply cleanly.
Attempt to apply a sensu config with Puppet 6.2 and sensu-puppet 2.62.3.

[not able to share]
Master

- Your OS: 
RHEL 7.5
- Your Ruby version:
- Your version of Puppet: 
Open Source 6.2
- Your version of Sensu: 
0.26.5-2
- Your version of this module: 
2.62.3

I patched around what I think is the issue. Pull request to come.

jwoytek:
# Pull Request Checklist

## Description

Every other parameter defined in sensu_enterprise_dashboard_config has a getter that returns the config object. I'm not sure why custom had something different, but this appears to cause issues with applying the sensu config. I do not claim to understand the purpose of having a reject in the getter, but applying this fix does not seem to cause problems, and a bad custom config is still caught. 

## Related Issue

Fixes #1080 

## How Has This Been Tested?

This has been tested in our open source Puppet 6.2 master server environment across multiple RHEL7 client systems. 

## General

### DIFF
--- a/lib/puppet/provider/sensu_enterprise_dashboard_config/json.rb
+++ b/lib/puppet/provider/sensu_enterprise_dashboard_config/json.rb
@@ -226,7 +226,7 @@ Puppet::Type.type(:sensu_enterprise_dashboard_config).provide(:json) do
   end
 
   def custom
-    conf['dashboard'].reject { |k,v| is_property?(k) }
+    conf['dashboard']['custom']
   end
 
   def custom=(value)


treydock:
I think this could have unintended consequences.  The purpose of `is_property?` was to avoid having custom include values that are already properties for the type.

jwoytek:
OK, we configure the sensu module through hiera. Our client configuration is pretty simple, actually:

```
sensu::server: false
sensu::install_repo: false
sensu::purge:
  config: true
sensu::rabbitmq_host: [redacted]
sensu::rabbitmq_password: [redacted]
sensu::rabbitmq_port: 5672
sensu::rabbitmq_vhost: /sensu
```

ghoneycutt:
@jwoytek could you provide an acceptance test along with this? We want to see that this fails before your patch and succeeds after.

treydock:
@jwoytek Based on conversation in #1080 I am not convinced this change would actually fix your problem.  Also the `custom` key is supposed to be set to any keys that exist in the JSON that are not known properties.  This change would only assign `custom` to items nested under `custom` key which is not the intended behavior of the `custom` property.

jwoytek:
Yeah, I'm sorry, @ghoneycutt and @treydock. I should have closed this after we discussed things in the associated ticket. It appears to have been a type caching issue in the upstream puppet master. Asking the module to return the potentially-empty object instead of trying to do a validity check an undefined object DID fix the problem I was having, but the bug wasn't in the sensu module code. I had to get the upstream server manager to reset their server and ensure that the type cache was rebuilt, which they had not done after upgrading the module. 

I'm going to shortly delete the branch associated with this PR, but I'm going to leave the original bug with some additional details so that it can be found by others who might be experiencing the same problem.

PR #1076, opened 2019-02-13
fixes: couldn't find login name expanding tilde

maxadamo:
# Pull Request Checklist

puppet fails to run unattended

## Description

tildes expansion does not seem to work if the process runs unattended. Hence, we can guess the UID of the user running puppet, and then we guess the home directory for that user: 

the new code:
```ruby
require 'etc'
user_name = Etc.getpwuid(Process.uid).name
home = Etc.getpwnam(user_name).dir
```
the old code: 
```ruby
home = File.expand_path('~')
```
## Related Issue

Fixes #1072 

## Motivation and Context

puppet fails to run unattended

## How Has This Been Tested? 

I have added the code on my puppetserver and I tested it on one agent.

PR #1082, opened 2019-02-15
fixes: couldn't find login name  expanding tilde (cont.)

### LINKED ISSUE #1072: Could not evaluate: couldn't find login name -- expanding `~'
## Description of problem

- if puppet is run unattended (as per defined schedule) it throws this error: 
```
Could not evaluate: couldn't find login name -- expanding `~'
```
- If puppet is run from a terminal, it works properly. 
- The error is only shown in the puppetserver.log, or in puppetDB (I have found it in the puppetboard, reporting tool).

The problem seems to be in the file: `lib/puppet/provider/sensuctl.rb`, at line 10 
```
home = File.expand_path('~')
```
The tilde is not expanded, and variable interpolation fails

- Your OS: Linux CentOS 7
- Your Ruby version: Sensu-Go
- Your version of Puppet: 6.2
- Your version of Sensu: 5.2.1
- Your version of this module: latest, branch sensu2

treydock:
# Pull Request Checklist

## Description

Fix how sensuctl config path is determined to work when run in context that doesn't support expanding `~`.

## Related Issue

Fixes #1072
Replaces #1076

### DIFF
--- a/lib/puppet/provider/sensuctl.rb
+++ b/lib/puppet/provider/sensuctl.rb
@@ -1,3 +1,4 @@
+require 'etc'
 require 'json'
 require 'tempfile'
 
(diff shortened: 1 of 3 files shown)


treydock:
@maxadamo This is same as your PR but with one fewer call to `Etc` and added some unit tests around `config_path`.

maxadamo:
so far, so good. My coding skills with ruby is limited.

maxadamo:
hello @treydock @ghoneycutt  don't get me wrong as I don't want to appear push, but if puppet fails to run for everyone, I'd not consider this as a cosmetic issue. To me it seems quite an urgent PR, or is there any other reason that I am not aware? 
sensu_error

PR #1083, opened 2019-02-25
Support installing sensu plugins and extensions

treydock:
# Pull Request Checklist

## Description

Add both type and provider and class to manage sensu plugin and extension installation.

Had to have apt based systems have packages depend on `Class[apt::update]` as adding a second repo caused `apt-get update` to take place later and prevented install of core sensu-go packages.  The dependency change ensures `apt-get update` is run before trying to install any packages.

## Motivation and Context

sensu-go docs for plugins: https://docs.sensu.io/sensu-go/5.2/installation/plugins/

This change makes it easy to utilize plugins.

## General

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,8 @@
     * [Advanced agent](#advanced-agent)
     * [Advanced SSL](#advanced-ssl)
     * [Enterprise support](#enterprise-support)
+    * [Installing Plugins](#installing-plugins)
+    * [Installing Extensions](#installing-extensions)
     * [Exported resources](#exported-resources)
     * [Resource purging](#resource-purging)
     * [Sensu backend cluster](#sensu-backend-cluster)
@@ -114,7 +116,7 @@ associated to `linux` and `apache-servers` subscriptions.
   class { 'sensu::agent':
     backends    => ['sensu-backend.example.com:8081'],
     config_hash => {
-      'subscriptions => ['linux', 'apache-servers'],
+      'subscriptions' => ['linux', 'apache-servers'],
     },
   }
 ```
@@ -161,7 +163,7 @@ class { 'sensu':
 class { 'sensu::agent':
   backends    => ['sensu-backend.example.com:8081'],
   config_hash => {
-    'subscriptions => ['linux', 'apache-servers'],
+    'subscriptions' => ['linux', 'apache-servers'],
   },
 }
 ```
@@ -185,6 +187,108 @@ class { 'sensu::backend':
 ```
 
 The type `sensu_ldap_auth` requires a valid enterprise license.
+
+### Installing Plugins
+
+Plugin management is handled by the `sensu::plugins` class.
+
+Example installing plugins on agent:
+
+```puppet
+  class { 'sensu::agent':
+    backends    => ['sensu-backend.example.com:8081'],
+    config_hash => {
+      'subscriptions' => ['linux', 'apache-servers'],
+    },
+  }
+  class { 'sensu::plugins':
+    plugins => ['disk-checks'],
+  }
+```
+
+The `plugins` parameter can also be a Hash that sets the version:
+
+```puppet
+  class { 'sensu::agent':
+    backends    => ['sensu-backend.example.com:8081'],
+    config_hash => {
+      'subscriptions' => ['linux', 'apache-servers'],
+    },
+  }
+  class { 'sensu::plugins':
+    plugins => {
+      'disk-checks' => { 'version' => 'latest' },
+    },
+  }
+```
+
+Set `dependencies` to an empty Array to disable the `sensu::plugins` dependency management.
+
+```puppet
+  class { 'sensu::plugins':
+    dependencies => [],
+  }
+```
+
+You can uninstall plugins by passing `ensure` as `absent`.
+
+```puppet
+  class { 'sensu::agent':
+    backends    => ['sensu-backend.example.com:8081'],
+    config_hash => {
+      'subscriptions' => ['linux', 'apache-servers'],
+    },
+  }
+  class { 'sensu::plugins':
+    plugins => {
+      'disk-checks' => { 'ensure' => 'absent' },
+    },
+  }
+```
+
+### Installing Extensions
+
+Extension management is handled by the `sensu::plugins` class.
+
+Example installing extension on backend:
+
+```puppet
+  class { 'sensu::backend':
+    password     => 'supersecret',
+    old_password => 'P@ssw0rd!',
+  }
+  class { 'sensu::plugins':
+    extensions => ['graphite'],
+  }
+```
+
+The `extensions` parameter can also be a Hash that sets the version:
+
+```puppet
+  class { 'sensu::backend':
+    password     => 'supersecret',
+    old_password => 'P@ssw0rd!',
+  }
+  class { 'sensu::plugins':
+    extensions => {
+      'graphite' => { 'version' => 'latest' },
+    },
+  }
+```
+
+You can uninstall extensions by passing `ensure` as `absent`.
+
+```puppet
+  class { 'sensu::backend':
+    password     => 'supersecret',
+    old_password => 'P@ssw0rd!',
+  }
+  class { 'sensu::plugins':
+    extensions => {
+      'graphite' => { 'ensure' => 'absent' },
+    },
+  }
+```
 
 ### Exported resources
 
--- a/data/os/RedHat.yaml
+++ b/data/os/RedHat.yaml
@@ -1,0 +1,6 @@
+---
+sensu::plugins::dependencies:
+  - make
+  - gcc
+  - gcc-c++
+  - openssl-devel
--- a/lib/puppet/provider/sensu_plugin/sensu_install.rb
+++ b/lib/puppet/provider/sensu_plugin/sensu_install.rb
@@ -1,0 +1,162 @@
+Puppet::Type.type(:sensu_plugin).provide(:sensu_install) do
+  desc "Provider sensu_check using sensuctl"
+
+  mk_resource_methods
+
+  commands :gem => '/opt/sensu-plugins-ruby/embedded/bin/gem'
+  commands :sensu_install => 'sensu-install'
+
+  def self.instances
+    plugins = []
+
+    output = gem('list', '--local', '^sensu-(plugins|extensions)')
+    Puppet.debug("gem output: #{output}")
+    output.each_line do |o|
+      next unless o.start_with?('sensu-')
+      # This regex matches the gem list format.
+      # First capture group is non-white space
+      # Second capture group is anything preceeded by space(s) wrapped in parentheses
+      # Expected format:
+      # gem-name (version, version)
+      if o =~ /^(\S+)\s+\((.+)\)/
+        gem_name = $1
+        versions = $2.sub('default: ', '').split(/,\s*/)
+        if gem_name.start_with?('sensu-extensions')
+          extension = :true
+        else
+          extension = :false
+        end
+        plugin = {
+          :name      => gem_name.sub('sensu-plugins-', '').sub('sensu-extensions-', ''),
+          :ensure    => :present,
+          :version   => versions.map{|v| v.split[0]}[0],
+          :extension => extension,
+        }
+        Puppet.debug("plugin: #{plugin}")
+        plugins << new(plugin)
+      end
+    end
+    plugins
+  end
+
+  def self.prefetch(resources)
+    plugins = instances
+    resources.keys.each do |name|
+      if provider = plugins.find { |c| c.name == name }
+        resources[name].provider = provider
+      end
+    end
+  end
+
+  def self.latest_versions
+    return @latest_versions if @latest_versions
+    @latest_versions = {}
+    output = gem('search', '--remote', '--all', "^sensu-(plugins|extensions)-")
+    output.each_line do |o|
+      # This regex matches the gem list format.
+      # First capture group is non-white space
+      # Second capture group is anything preceeded by space(s) wrapped in parentheses
+      # Expected format:
+      # gem-name (version, version)
+      if o =~ /^(\S+)\s+\((.+)\)/
+        gem_name = $1
+        versions = $2.sub('default: ', '').split(/,\s*/)
+        Puppet.debug("#{gem_name} versions: #{versions}")
+        name = gem_name.sub('sensu-plugins-', '').sub('sensu-extensions-', '')
+        ver = versions.map { |v| v.split[0]}[0]
+        @latest_versions[name] = ver
+      end
+    end
+    @latest_versions
+  end
+
+  def exists?
+    @property_hash[:ensure] == :present
+  end
+
+  def initialize(value = {})
+    super(value)
+    @property_flush = {}
+  end
+
+  def version=(value)
+    @property_flush[:version] = value
+  end
+
+  def install(version)
+    args = []
+    if resource[:extension] == :true
+      args << '--extension'
+      prefix = 'sensu-extensions-'
+    else
+      args << '--plugin'
+      prefix = 'sensu-plugins-'
+    end
+    if version == :latest
+      latest_versions = self.class.latest_versions
+      version = latest_versions[resource[:name]]
+    end
+    if version 
+      name = "#{prefix}#{resource[:name]}:#{version}"
+    else
+      name = resource[:name]
+    end
+    args << name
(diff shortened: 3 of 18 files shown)


manifests/init.pp
```diff
@@ -70,11 +63,21 @@
 
   case $facts['os']['family'] {
     'RedHat': {
+      $os_package_require = []
```
ghoneycutt: does this not work if it is set to `undef` ?

manifests/init.pp
```diff
@@ -70,11 +63,21 @@
...
     default: {
       fail("Detected osfamily <${::osfamily}>. Only RedHat and Debian are supported.")
     }
   }
+
+  if $manage_repo {
+    include ::sensu::repo
+    $package_require = [Class['::sensu::repo']] + $os_package_require
+  } else {
+    $package_require = undef
```
ghoneycutt: Perhaps we should document which class(es) this `$package_require` variable is meant to be used in. Since it is just specified here and not used, it makes the code harder to understand.

manifests/plugins.pp
```diff
@@ -0,0 +1,92 @@
...
+#   Default is OS dependent.
+# @param plugins
+#   Plugins to install
+# @param extensions
+#   Extensions to install
+#
+class sensu::plugins (
+  String $package_ensure = 'installed',
+  String $package_name = 'sensu-plugins-ruby',
+  Array $dependencies = [],
```
ghoneycutt: We should not use empty assignments such as `''`, `[]` and `{}` and instead use `undef`.

manifests/plugins.pp
```diff
@@ -0,0 +1,92 @@
...
+  Variant[Array, Hash] $plugins = {},
+  Variant[Array, Hash] $extensions = {},
+) {
+
+  include ::sensu
+
+  if $::sensu::manage_repo {
```
ghoneycutt: this seems to duplicate the code already existing in the sensu class.

manifests/repo/community.pp
```diff
@@ -0,0 +1,50 @@
...
+        baseurl         => "https://packagecloud.io/sensu/community/el/${repo_release}/\$basearch",
+        descr           => 'sensu_community',
+        enabled         => 1,
+        gpgcheck        => 0,
+        gpgkey          => 'https://packagecloud.io/sensu/community/gpgkey',
+        metadata_expire => 300,
+        repo_gpgcheck   => 1,
+        sslcacert       => '/etc/pki/tls/certs/ca-bundle.crt',
+        sslverify       => 1,
+}
```
ghoneycutt: style, i think this paran is misaligned..

manifests/repo/community.pp
```diff
@@ -0,0 +1,50 @@
...
+        include  => {
+          'src' => true,
+        },
+        key      => {
+          'id'     => '7F54E8A5C0CB51DBE612D2F50156BD72FEC8CD59',
+          'source' => 'https://packagecloud.io/sensu/community/gpgkey',
+        },
+      }
+    }
+    default: {
```
ghoneycutt: perhaps the default should be to emit a `warning()`

data/os/RedHat.yaml
```diff
@@ -0,0 +1,6 @@
+---
+sensu::plugins::dependencies:
```
ghoneycutt: the dependencies are missing from the spec tests

manifests/init.pp
treydock: I do array merge later so wanted all values to be arrays to keep thing simpler.

manifests/repo/community.pp
treydock: It's not possible to ever hit this default, it's just there for lint.  The `sensu` class will fail if `osfamily` is not `RedHat` or `Debian`.

manifests/plugins.pp
treydock: So using `undef` would require nesting a lot of logic inside an `if` statements which could make the rest of manifest a bit harder to read without changing behavior.  Still want `undef` as default?  Also in case of `$dependencies` the real default is specified in Hiera data.

manifests/repo/community.pp
treydock: It's not, at least doesn't look like it in VIM.  Everything lines up with `metadata_expire` which is longest parameter name.

manifests/init.pp
treydock: Add comment about where `$package_require` is used.

data/os/RedHat.yaml
treydock: Added these to unit tests.

manifests/repo/community.pp
ghoneycutt: That's what should happen :)

manifests/plugins.pp
ghoneycutt: Yes, please use `undef` and update the data types. I think it will only introduce one layer of an `if` statement around the current code.

spec/classes/plugins_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'spec_helper'
+
+describe 'sensu::plugins', :type => :class do
+  on_supported_os({facterversion: '3.8.0'}).each do |os, facts|
+    if facts[:osfamily] == 'Debian'
+      package_require = ['Class[Sensu::Repo::Community]', 'Class[Apt::Update]']
```
ghoneycutt: What I've done in the past is outside the describe set the data such as

```ruby
platforms => {
  debian => {
    :package_require => ['Class[Sensu::Repo::Community]', 'Class[Apt::Update]'],
    :dependencies => ['make','gcc','g++','libssl-dev'],
  },
  el => {
    :package_require => ['Class[Sensu::Repo::Community]'],
    :dependencies => ['make','gcc','gcc-c++','openssl-devel'],
  },
}
```

Then keep your if statement and use that to assign the data

```
package_require = platforms[#{facts[:osfamily]}[:package_require]
```

This is useful as the data changes and we add more platforms. With the current code you assume if osfamily is not Debian then it must be RedHat. By using the above method, it makes it very clear which platform's data is being used.

manifests/plugins.pp
treydock: Updated to default to `undef`.

manifests/init.pp
ghoneycutt: makes sense to me, especially since this is a private variable and not a param

manifests/init.pp
```diff
@@ -70,11 +63,23 @@
 
   case $facts['os']['family'] {
     'RedHat': {
+      $os_package_require = []
     }
     'Debian': {
+      $os_package_require = [Class['::apt::update']]
     }
     default: {
       fail("Detected osfamily <${::osfamily}>. Only RedHat and Debian are supported.")
```
ghoneycutt: I dont think osfamily is used anymore in lieu of `$facts['os']['family']`

manifests/plugins.pp
```diff
@@ -0,0 +1,94 @@
...
+# @param package_ensure
+#   Ensure property for sensu plugins package.
+# @param package_name
+#   Name of the Sensu plugins ruby package.
+# @param dependencies
+#   Package dependencies needed to install plugins and extensions.
+#   Default is OS dependent.
+# @param plugins
+#   Plugins to install
+# @param extensions
```
ghoneycutt: what's a plugin vs an extension?

manifests/repo/community.pp
ghoneycutt: check you aren't using hard tabs..  this paran is definitely off as it should line up with `yumrepo`

manifests/repo/community.pp
ghoneycutt: To avoid the empty case statement, suggest doing two if statements

```puppet
if $facts['os']['family'] == 'RedHat' {
  ...
}

if $facts['os']['family'] == 'Debian' {
  ...
}
```

This is also easier to read than a if-elsif

spec/acceptance/04_plugins_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'spec_helper_acceptance'
+
+describe 'sensu::plugins class', unless: RSpec.configuration.sensu_cluster do
+  agent = hosts_as('sensu_agent')[0]
+  backend = hosts_as('sensu_backend')[0]
+  context 'on agent' do
+    it 'should work without errors' do
```
ghoneycutt: The `it` text should be updated to show that we're checking idempotency.

`it should work without errors and be idempotent` or you can wrap that idempotency check like this

```ruby
it 'should be idempotent' do
  apply_manifest_on(agent, pp, :catch_changes  => true)
end
```

spec/classes/plugins_spec.rb
treydock: Took similar approach but defined a function in spec_helper that can be used by all the unit tests so these are all in one place.

manifests/plugins.pp
treydock: I'm not sure but the command `sensu-install` makes a distinction as plugins use `--plugin` flag and extensions use `--extension` flag and gem names have different prefixes.

manifests/init.pp
treydock: Changed to match how fact is referenced for actual case statement to use `$facts['os']['family']`.

spec/acceptance/04_plugins_spec.rb
treydock: Used the `it should work without errors and be idempotent`.

treydock:
Tests are failing because Debian lacks `sensu-plugins-ruby` package for Debian.  See https://github.com/sensu/sensu-go/issues/2757

ghoneycutt:
Thanks for creating the ticket with sensu-go. Suggest skipping those tests with Debian and adding a comment like

```
# TODO: package is missing on Debian - See https://github.com/sensu/sensu-go/issues/2757
```

treydock:
Add commit to skip all plugin related tests for Debian.

treydock:
Added commit that documents `sensu::plugins` in README.

README.md
```diff
@@ -173,6 +174,40 @@ class { 'sensu':
 }
 ```
 
+### Installing Plugins
+
+Plugin installation is handled by the `sensu::plugins` class.
```
ghoneycutt: s/installation/management

Installation is a non-idempotent term

README.md
```diff
@@ -173,6 +174,40 @@ class { 'sensu':
 }
 ```
 
+### Installing Plugins
```
ghoneycutt: Need a similar section for Managing Extensions

README.md
treydock: Change made

spec/acceptance/04_plugins_spec.rb
```diff
@@ -0,0 +1,83 @@
+require 'spec_helper_acceptance'
+
+describe 'sensu::plugins class', unless: RSpec.configuration.sensu_cluster do
+  agent = hosts_as('sensu_agent')[0]
+  backend = hosts_as('sensu_backend')[0]
+  before do
+    if fact_on(agent, 'operatingsystem') == 'Debian'
+      skip("TODO: package is missing on Debian - See https://github.com/sensu/sensu-go/issues/2757")
```
ghoneycutt: Should probably just link to https://github.com/sensu/sensu-plugins-omnibus/issues/3

spec/acceptance/sensu_plugin_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'spec_helper_acceptance'
+
+describe 'sensu_plugin', if: RSpec.configuration.sensu_full do
+  agent = hosts_as('sensu_agent')[0]
+  before do
+    if fact_on(agent, 'operatingsystem') == 'Debian'
+      skip("TODO: package is missing on Debian - See https://github.com/sensu/sensu-go/issues/2757")
```
ghoneycutt: Should probably just link to https://github.com/sensu/sensu-plugins-omnibus/issues/3

README.md
treydock: Added section.

spec/acceptance/04_plugins_spec.rb
treydock: Updated

spec/acceptance/sensu_plugin_spec.rb
treydock: Updated

manifests/repo/community.pp
treydock: Implemented.  I'll open separate pull request to do this in `sensu::repo`.

manifests/repo/community.pp
treydock: I fixed that when switching to if statements.

lib/puppet/provider/sensu_plugin/sensu_install.rb
```diff
@@ -0,0 +1,152 @@
...
+  commands :sensu_install => 'sensu-install'
+
+  def self.instances
+    plugins = []
+
+    output = gem('list', '--local', '^sensu-(plugins|extensions)')
+    Puppet.debug("gem output: #{output}")
+    output.each_line do |o|
+      next unless o.start_with?('sensu-')
+      if o =~ /^(\S+)\s+\((.+)\)/
```
ghoneycutt: please document this regex in plain english

README.md
```diff
@@ -173,6 +175,70 @@ class { 'sensu':
...
```
ghoneycutt: If you can specify plugins and extensions as absent, this should be `### Managing Plugins`

README.md
```diff
@@ -173,6 +175,70 @@ class { 'sensu':
...
+
+Plugin management is handled by the `sensu::plugins` class.
+
+Example installing plugins on agent:
+
+```puppet
```
ghoneycutt: should show how to disable the dependencies as many folks will want to do this because the packages can conflict

lib/puppet/provider/sensu_plugin/sensu_install.rb
```diff
@@ -0,0 +1,152 @@
...
+      end
+    end
+  end
+
+  def self.latest_versions
+    return @latest_versions if @latest_versions
+    @latest_versions = {}
+    output = gem('search', '--remote', '--all', "^sensu-(plugins|extensions)-")
+    output.each_line do |o|
+      if o =~ /^(\S+)\s+\((.+)\)/
```
ghoneycutt: document regex

lib/puppet/type/sensu_plugin.rb
```diff
@@ -0,0 +1,82 @@
...
+@summary Manages Sensu plugins
+@example Install a sensu plugin
+  sensu_plugin { 'disk-checks':
+    ensure  => 'present',
+  }
+
+@example Install latest version of a sensu plugin
+  sensu_plugin { 'disk-checks':
+    ensure  => 'present',
+    version => 'latest',
```
ghoneycutt: example should show how to pin to a specific version as sensu recommends that you pin your check versions

manifests/plugins.pp
```diff
@@ -0,0 +1,94 @@
...
+#   Extensions to install
+#
+class sensu::plugins (
+  String $package_ensure = 'installed',
+  String $package_name = 'sensu-plugins-ruby',
+  Optional[Array] $dependencies = undef,
+  Optional[Variant[Array, Hash]] $plugins = undef,
+  Optional[Variant[Array, Hash]] $extensions = undef,
```
ghoneycutt: Seems that VP is advocating using `[]` and `{}`. Should this be changed?

lib/puppet/provider/sensu_plugin/sensu_install.rb
treydock: Added comments.

README.md
treydock: Added docs about setting `absent` for both plugins and extensions.

README.md
treydock: Add docs to plugins section about passing `dependencies => []`.

lib/puppet/provider/sensu_plugin/sensu_install.rb
treydock: Added comments

lib/puppet/type/sensu_plugin.rb
treydock: Added example of passing specific version.

manifests/plugins.pp
treydock: I switched all 3 from using `Optional` to defaulting to empty Array.

PR #1084, opened 2019-02-26
Remove dependency pinnings from Gemfile

ghoneycutt:
Attempt to simplify the Gemfile by removing unnecessary version pinning.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,17 +7,17 @@ else
 end
 
 group :development, :unit_tests do
-  gem 'rake',                                             '< 11.0.0'
-  gem 'rspec-puppet', '~> 2.6.0',                         :require => false
+  gem 'rake'
+  gem 'rspec-puppet',                                     :require => false
   gem 'rspec-puppet-facts',                               :require => false
   gem 'rspec-mocks',                                      :require => false
-  gem 'puppetlabs_spec_helper', '>= 2.11.0',              :require => false
-  gem 'puppet-lint', "~> 2.0",                            :require => false
-  gem 'json', "~> 1.8.3",                                 :require => false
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -7,17 +7,17 @@ else
...
-  gem 'puppetlabs_spec_helper', '>= 2.11.0',              :require => false
-  gem 'puppet-lint', "~> 2.0",                            :require => false
-  gem 'json', "~> 1.8.3",                                 :require => false
-  gem 'json_pure', "~> 1.8.3",                            :require => false
+  gem 'puppetlabs_spec_helper',                           :require => false
   gem 'metadata-json-lint',                               :require => false
   gem 'puppet-blacksmith',                                :require => false
-  gem 'rest-client', ">= 1.7.4",                          :require => false
+  gem 'json',                                             :require => false
+  gem 'json_pure',                                        :require => false
```
treydock: We don't directly use `json` or `json_pure`, they should probably be pulled in by whatever else needs them.  Maybe remove them?  I think same could be said for `rest-client`.

Gemfile
```diff
@@ -41,18 +41,18 @@ group :documentation do
 end
 
 group :system_tests do
-  gem 'beaker', '~> 4.x',             :require => false
```
treydock: This would only become a problem if/when beaker 5 comes out, the 3 to 4 transition wasn't painful but required changes to testing.

ghoneycutt: We'll find that during the automated tests and can always pin if needed.

Gemfile
ghoneycutt: Let's do this in another PR after this one passes

PR #1088, opened 2019-03-03
Add env_vars parameter

### LINKED ISSUE #1074: Allow adding arbitrary config variables in /etc/default/sensu
It would be useful o have more flexibility in managing **/etc/default/sensu** The way it's currently set is limiting us on injecting more configuration variables.  As a reference I'm adding [PR852](https://github.com/sensu/sensu-puppet/pull/852/) - think that one could work.

treydock:
# Pull Request Checklist

## Description

Add `env_vars` parameter that provides environment variables to `/etc/default/sensu`.

## Related Issue

Fixes #1074

## Motivation and Context

Support arbitrary environment variables in `/etc/default/sensu`.

### DIFF
--- a/manifests/package.pp
+++ b/manifests/package.pp
@@ -28,6 +28,8 @@
 #
 # @param path Used to set PATH in /etc/default/sensu
 #
+# @param env_vars Additional environment variables for /etc/default/sensu
+#
 # @param rubyopt Ruby opts to be passed to the sensu services
 #
 # @param use_embedded_ruby If the embedded ruby should be used, e.g. to install the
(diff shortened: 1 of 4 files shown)


manifests/package.pp
```diff
@@ -28,6 +28,8 @@
 #
 # @param path Used to set PATH in /etc/default/sensu
 #
+# @param env_vars Additional environment variables for /etc/default/sensu
```
ghoneycutt: why is this specified in two places since there is only one `/etc/default/sensu` ?

treydock: Was just following convention of parameters being documented in `sensu` and `sensu::package`.  Since `sensu::package` is private I think having in `sensu` is sufficient.

treydock: Want me to remove or leave?  All the other parameters in `sensu::package` are documented in 2 places.

ghoneycutt: why is the param in both?

treydock: The public access is in `sensu` class and the `sensu::package` parameter pulls the value from `$::sensu::env_vars` so that template can use `@env_vars`.  Also just how all the parameters in `sensu::package` are handled, was just following the convention.  Could define in just `sensu` class and access via `scope['sensu::env_vars']` in template but this would be the first time that pattern is used for parameters.

PR #1094, opened 2019-03-21
Add sensu_ad_auth type to support Active Directory authentication

treydock:
# Pull Request Checklist

## Description

Support adding Active Directory auth source.

## Motivation and Context

Sensu Go 5.3.0 added Enterprise support for Active Directory.  This is nearly identical to the existing LDAP support as far as Puppet code is concerned.


spec/fixtures/unit/provider/sensuctl/auths.txt
```diff
@@ -0,0 +1,77 @@
+type: AD
```
ghoneycutt: how does this work? why are there what seems like three different fixtures in one file?

spec/unit/sensu_ad_auth_spec.rb
```diff
@@ -0,0 +1,322 @@
...
+    }
+  end
+  let(:config) do
+    default_config
+  end
+  let(:auth) do
+    described_class.new(config)
+  end
+
+  it 'should add to catalog with raising an error' do
```
ghoneycutt: it should add to the catalog with raising an error and then expect to not raise an error? huh?

spec/fixtures/unit/provider/sensuctl/auths.txt
treydock: So the output of `sensuctl auth list --format json` doesn't distinguish between LDAP or AD, the `type` value isn't exposed.  I have to instead run `sensuctl auth list --format yaml` and iterate over the returned values which does expose `type` and build a hash of `{ name => type }` so that the code knows if auth listed with JSON are AD or LDAP.

This is the first type where `sensuctl <type> list` contains multiple `type` objects.  Everything else is 1:1 where list is just the type from <type>.  With auth there are two types.  Doing `sensuctl check list --format json` also doesn't expose `type` but not necessary, so not convinced this is really a bug, maybe just annoyance.

I added 3 to file because I wanted to make sure in unit tests that the code was correctly handling a mix of LDAP and AD and multiple of one type.

spec/unit/provider/sensuctl_spec.rb
```diff
@@ -79,4 +79,11 @@
       subject.sensuctl_delete('check','test')
     end
   end
+
+  context 'sensuctl_auth_types' do
+    it 'should return auth and their types' do
+      allow(subject).to receive(:sensuctl).with(['auth','list','--format','yaml']).and_return(my_fixture_read('auths.txt'))
```
treydock: @ghoneycutt This is where that fixture you asked about is loaded.

spec/unit/sensu_ad_auth_spec.rb
treydock: Really unfortunate typo,  `s/with/without/`.  Fixing, but I think this may be present in all the unit tests as I usually copy existing tests for new types.  I added commit to fix other tests where this is also present.

spec/fixtures/unit/provider/sensuctl/auths.txt
ghoneycutt: This sounds like a sensu bug if the items returned differ between json and yaml

treydock: I think they are supposed to differ.  All `--format json` returns is the `spec` and sometimes `metadata` for an object, which is consistent across all objects.  The `type` is stored outside `spec` as top-level key that only shows up with `--format yaml` and --format wrapped-json`.  Those two formats are not valid YAML or JSON because they print each object as a new YAML or JSON document so you can't do something like this: `sensuctl list auth --format wrapped-json | python -m json.tool`.  The code I had to write split `--format yaml` on each `---` which is used by sensuctl to separate records.

Imagine taking multiple YAML files then doing "cat file >> new.yaml" , that's kind of what it's like for `--format yaml`.  Each object returned is returned as a separate YAML document and not another object in one big hash or array.  I raised this issue with wrapped-json but at the time the yaml format had not been added so I think same issue exists for both: https://github.com/sensu/sensu-go/issues/1817

I'll double check that all output from `--format json` is consistent, if all types do indeed exclude the `type` value then I think the most we could ask of upstream is to include `type` as key in `--format json`.

PR #1095, opened 2019-03-21
Ensure presence of sensuclassic doesn't break this module

treydock:
# Pull Request Checklist

## Description

Installing sensuclassic module during acceptance tests to ensure this module won't break when sensuclassic is installed.

I will do same to sensuclassic once sensu-go changes are in forge.


ghoneycutt:
Sensu-go 3.0.0 is on the forge now

PR #1097, opened 2019-03-24
Hiera resources

treydock:
# Pull Request Checklist

## Description

This adds ability to define all types via Hiera.  Renamed `sensu::backend::resources` to `sensu::backend::default_resources` to better match purpose of the class and allow these user defined resources to be defined via `sensu::backend::resources`.

## Motivation and Context

This should make it easier for this module to be used with less possible changes needed to something like a profile class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@
     * [Installing Plugins](#installing-plugins)
     * [Installing Extensions](#installing-extensions)
     * [Exported resources](#exported-resources)
+    * [Hiera resources](#hiera-resources)
     * [Resource purging](#resource-purging)
     * [Sensu backend cluster](#sensu-backend-cluster)
         * [Adding backend members to an existing cluster](#adding-backend-members-to-an-existing-cluster)
@@ -306,6 +307,60 @@ The backend system would collect all `sensu_check` resources.
 
 ```puppet
   Sensu_check <<||>>
+```
+
+### Hiera resources
+
+All the types provided by this module can have their resources defined via Hiera. A type such as `sensu_check` would be defined via `sensu::backend::checks`.
+
+The following example adds an asset, filter, handler and checks via Hiera:
+
+```yaml
+sensu::backend::assets:
+  sensu-email-handler:
+    ensure: present
+    url: 'https://github.com/sensu/sensu-email-handler/releases/download/0.1.0/sensu-email-handler_0.1.0_linux_amd64.tar.gz'
+    sha512: '755c7a673d94997ab9613ec5969666e808f8b4a8eec1ba998ee7071606c96946ca2947de5189b24ac34a962713d156619453ff7ea43c95dae62bf0fcbe766f2e'
+    filters:
+      - "entity.system.os == 'linux'"
+      - "entity.system.arch == 'amd64'"
+sensu::backend::filters:
+  hourly:
+    ensure: present
+    action: allow
+    expressions:
+      - 'event.check.occurrences == 1 || event.check.occurrences % (3600 / event.check.interval) == 0'
+sensu::backend::handlers:
+  email:
+    ensure: present
+    type: pipe
+    command: "sensu-email-handler -f root@localhost -t user@example.com -s localhost -i"
+    timeout: 10
+    runtime_assets:
+      - sensu-email-handler
+    filters:
+      - is_incident
+      - not_silenced
+      - hourly
+sensu::backend::checks:
+  check-cpu:
+    ensure: present
+    command: check-cpu.sh -w 75 -c 90
+    interval: 60
+    subscriptions:
+      - linux
+    handlers:
+      - email
+    publish: true
+  check-disks:
+    ensure: present
+    command: "/opt/sensu-plugins-ruby/embedded/bin/check-disk-usage.rb -t '(xfs|ext4)'"
+    subscriptions:
+      - linux
+    handlers:
+      - email
+    interval: 1800
+    publish: true
 ```
 
 ### Resource purging
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -51,6 +51,44 @@
 # @param license_content
 #   The content of sensu-go enterprise license
 #   Do not define with license_source
+# @param ad_auths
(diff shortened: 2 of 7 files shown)


manifests/backend/resources.pp
```diff
@@ -1,152 +1,25 @@
...
-    ensure   => 'present',
-    role_ref => 'system:user',
-    subjects => [
-      {
-        'type' => 'Group',
-        'name' => 'system:users'
-      },
-    ],
-  }
+  create_resources('sensu_ad_auth', $::sensu::backend::ad_auths)
```
ghoneycutt: `create_resources()` is the old way, the current way would be to use iteration

https://www.devco.net/archives/2015/12/16/iterating-in-puppet.php

manifests/backend/resources.pp
```diff
@@ -1,152 +1,25 @@
-# @summary Default sensu resources
+# @summary Define sensu resources
```
ghoneycutt: why does this class stick around?

manifests/backend.pp
```diff
@@ -73,13 +111,33 @@
   Boolean $show_diff = true,
   Optional[String] $license_source = undef,
   Optional[String] $license_content = undef,
+  Hash $ad_auths = {},
```
ghoneycutt: are these all of the sensu_ types? is the mapping to just drop `sensu_` from the name of the type?

ghoneycutt: would any of these make sense for the agent instead of the backend?

manifests/backend/resources.pp
treydock: I wanted to define the resources in separate class to avoid cluttering up `sensu::backend`.  Will be even more important if using iteration as that will take up more lines of code.

manifests/backend.pp
treydock: All these are dropping `sensu_` and making plural except silencing is a bit weird.  These all require executing `sensuctl` and thus not supported on agent, this module only supports running `sensuctl` types and providers on backend.  The only custom type that will make sense on agent is `sensu_plugin` and that already supports defining via parameters.

manifests/backend/resources.pp
treydock: Redid using iterations.

manifests/backend/resources.pp
```diff
@@ -1,152 +1,101 @@
...
-  sensu_cluster_role_binding { 'system:user':
-    ensure   => 'present',
-    role_ref => 'system:user',
-    subjects => [
-      {
-        'type' => 'Group',
-        'name' => 'system:users'
-      },
-    ],
+  $::sensu::backend::ad_auths.each |$name, $ad_auth| {
```
ghoneycutt: These should be in alphabetical order

manifests/backend.pp
```diff
@@ -73,13 +111,33 @@
...
+  Hash $assets = {},
```
ghoneycutt: These should be in alphabetical order

README.md
```diff
@@ -308,6 +309,22 @@ The backend system would collect all `sensu_check` resources.
...
+The following example adds a check via Hiera:
+
+```yaml
+sensu::backend::checks:
+  check-cpu:
+    ensure: present
+    command: check-cpu.sh -w 75 -c 90
+    interval: 60
+    subscriptions:
+      - linux
```
ghoneycutt: add another common one here to drive home that this work for any type and provider

manifests/backend/resources.pp
treydock: Are they not?  I did "ls -la" on lib/puppet/types and did it in that order.

manifests/backend.pp
```diff
@@ -51,6 +51,44 @@
...
+#   Hash of sensu_ldap_auth resources
+# @param mutators
+#   Hash of sensu_mutator resources
+# @param namespaces
+#   Hash of sensu_namespace resources
+# @param role_bindings
+#   Hash of sensu_role_binding resources
+# @param roles
+#   Hash of sensu_role resources
+# @param silencing
```
ghoneycutt: please use `silencings` with a `s` to ensure consistency

treydock: Fixed.

README.md
treydock: Added some examples based on real world usage and examples from sensu-go docs.

README.md
```diff
@@ -308,6 +309,60 @@ The backend system would collect all `sensu_check` resources.
   Sensu_check <<||>>
 ```
 
+### Hiera resources
+
+All the types provided by this module can have their resources defined via Hiera. A type such as `sensu_check` would be defined via `sensu::backend::checks`.
+
+The following example adds a check via Hiera:
```
ghoneycutt: This should be rephrased as the example does a lot more then adding a check

spec/classes/backend_spec.rb
```diff
@@ -247,6 +252,20 @@
            is_expected.to compile.and_raise_error(/Do not define both license_source and license_content/)
         end
       end
+
+      context 'checks defined' do
```
ghoneycutt: we should have similar tests for each type.

README.md
treydock: Updated wording.

spec/classes/backend_spec.rb
treydock: Because of amount of tests needed and fact resources actually defined in a different class, I moved tests for all 19 parameters to separate file.

spec/classes/backend_resources_spec.rb
```diff
@@ -4,181 +4,275 @@
   on_supported_os({facterversion: '3.8.0'}).each do |os, facts|
```
ghoneycutt: There's no logic wrapped around this class. Having this loop over every supported platform seems like it is just adding a lot of testing time without real benefit. Suggest setting this to EL7.

treydock: Limited those tests and ones for `sensu::backend::default_resources` to just EL7 since both are really just defining types with no OS specific logic.

PR #1101, opened 2019-04-07
Support opting out of tessen phone home

treydock:
# Pull Request Checklist

## Description

Support the ability to opt out of tessen phone home.

## Motivation and Context

Sensu Go 5.5 added Tessen phone-home and is opt-in by default.  This change maintains default behavior of opt-in but makes it easy to opt-out.

## How Has This Been Tested?

Unit and acceptance.

### DIFF
--- a/manifests/backend/tessen.pp
+++ b/manifests/backend/tessen.pp
@@ -1,0 +1,24 @@
+# @summary Manage tessen phone home
+# @api private
+#
+class sensu::backend::tessen {
+  include ::sensu::backend
+
+  if $::sensu::backend::tessen_ensure == 'present' {
+    $command = 'sensuctl tessen opt-in'
+    $match = 'true' # lint:ignore:quoted_booleans
+  } else {
+    $command = 'sensuctl tessen opt-out --skip-confirm'
+    $match = 'false' # lint:ignore:quoted_booleans
+  }
+
+  exec { $command:
+    path    => '/usr/bin:/bin:/usr/sbin:/sbin',
+    onlyif  => "sensuctl tessen info --format json | grep 'opt_out' | grep -q ${match}",
+    require => [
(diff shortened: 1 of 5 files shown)


spec/classes/backend_tessen_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'spec_helper'
+
+describe 'sensu::backend::tessen', :type => :class do
+  on_supported_os({
+    facterversion: '3.8.0',
```
ghoneycutt: Do we have other tests that are always the same for each OS that we are testing with each OS? Perhaps there is an opportunity to speed things up.

ghoneycutt:
Looks great, just waiting on Travis. Thanks @treydock

manifests/backend/tessen.pp
```diff
@@ -0,0 +1,24 @@
+# @summary Manage tessen phone home
+# @api private
+#
+class sensu::backend::tessen {
+  include ::sensu::backend
+
+  if $::sensu::backend::tessen_ensure == 'present' {
+    $command = 'sensuctl tessen opt-in'
+    $match = 'true'
```
ghoneycutt: these true/false lines need a puppet-lint control comment as they're currently failing

treydock: Added control comments

spec/classes/backend_tessen_spec.rb
treydock: The only place we currently change behavior for OS is the repo classes.  Everything else is not OS specific but not sure that will always be the case, especially when Windows support gets added.

treydock: One way to speed up at least unit tests is to use parallel rspec which would execute each spec file as separate process.

PR #1105, opened 2019-05-05
Support Sensu Go 5.6

treydock:
# Pull Request Checklist

## Description

* Fix issue when sensuctl <resource> list returns null
* Use shared code to handle parsing output from sensuctl <resource> list
* Support trusted_ca_file, client_cert_file and client_key_file for AD and LDAP auth
* Make server_binding optional for AD and LDAP auth

## Related Issue

Works around issue https://github.com/sensu/sensu-go/issues/2915

## Motivation and Context

Sensu-Go 5.6 made AD and LDAP authentication optional. Also new properties for AD and LDAP were added to supply CA certs and keys.  An issue was discovered where `sensuctl <resource> list` will return `null` instead of empty array and this PR addresses that and moves the JSON parsing logic to shared function.

## How Has This Been Tested?

Unit and acceptance


lib/puppet/provider/sensu_ad_auth/sensuctl.rb
```diff
@@ -108,7 +104,7 @@ def flush
         if @property_flush[:server_binding]
           server['binding'] = @property_flush[:server_binding][host]
         else
-          server['binding'] = resource[:server_binding][host]
+          server['binding'] = resource[:server_binding][host] if resource[:server_binding]
```
ghoneycutt: does this conditional have a test?

lib/puppet/provider/sensu_entity/sensuctl.rb
```diff
@@ -8,14 +8,7 @@
   def self.instances
     entities = []
 
-    output = sensuctl_list('entity')
-    Puppet.debug("sensu entities: #{output}")
```
ghoneycutt: why has this block been removed?

spec/unit/provider/sensu_entity/sensuctl_spec.rb
```diff
@@ -8,12 +8,12 @@
 
   describe 'self.instances' do
     it 'should create instances' do
-      allow(@provider).to receive(:sensuctl_list).with('entity').and_return(my_fixture_read('entity_list.json'))
+      allow(@provider).to receive(:sensuctl_list).with('entity').and_return(JSON.parse(my_fixture_read('entity_list.json')))
```
ghoneycutt: why did we just read the json before and now we read and then parse?

lib/puppet/provider/sensu_ad_auth/sensuctl.rb
treydock: Indirectly, yes.  I added more explicit test.

spec/unit/provider/sensu_entity/sensuctl_spec.rb
treydock: Because before `sensuctl_list` would return a JSON string that had to be parsed in each provider.  That parsing of the string was moved to shared function so now that function returns parsed JSON so rather than rewrite how the tests are written I just parse the same fixtures.

lib/puppet/provider/sensu_entity/sensuctl.rb
treydock: Previously the function `sensuctl_list` would return raw output and each provider had to parse the JSON string.  I moved that common logic into the function and so now every provider just gets back the parsed JSON.  This put the code for parsing JSON and handling edge cases in one place.

treydock:
Added commit that should fix failures with clusters.  The JSON from `--format json` for cluster commands is actually two separate JSON structures so updated logic to handle old and new formats.

ghoneycutt:
Released in v3.2.0

PR #1108, opened 2019-05-10
Add Windows support for Sensu Go agent

treydock:
# Pull Request Checklist

## Description

Add Windows support for agent

## Related Issue

Fixes #1106

## Motivation and Context

Sensu Go 5.7 added better support for Windows, this PR adds that support.

PR #1109, opened 2019-05-11
Fix cluster tests to work with Sensu Go 5.7

treydock:
# Pull Request Checklist

## Description

Fix cluster acceptance tests to work with sensu go 5.7

The Puppet provider has extra logic to handle both old and fixed JSON but the acceptance tests were targeted at latest release which now fixes JSON output.

Fix upstream: https://github.com/sensu/sensu-go/pull/2924


treydock:
Borrowed commit from #1108 to limit travis tests so only one build happens during a PR.  This PR will be needed before #1108 fully passes.

PR #1114, opened 2019-05-25
Support listing sensuctl resources using chunk-size

treydock:
# Pull Request Checklist

## Description

Add support for passing `--chunk-size` when listing sensuctl resources.

Adds `sensuctl_chunk_size` parameter to `sensu::backend` class.  Default is not to use `--chunk-size` to maintain compatibility with versions of Sensu Go before 5.8.

The `sensuctl_config` type is abstract type capable of configuring other types so that one value can be passed to multiple types and used in prefetch before a type's parameters are evaluated.

## Motivation and Context

Sensu Go 5.8 added `--chunk-size` flag to list commands and this adds support for that new flag. This would be useful for sites with large number of resources in their backend where chunking requests would be needed to avoid timeouts.

WIP:

- [x] Remove commented out code if not needed
- [x] Document in README?
- [x] Add acceptance tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@
     * [Resource purging](#resource-purging)
     * [Sensu backend cluster](#sensu-backend-cluster)
         * [Adding backend members to an existing cluster](#adding-backend-members-to-an-existing-cluster)
+    * [Large Environment Considerations](#large-environment-considerations)
 4. [Reference](#reference)
     * [Facts](#facts)
 5. [Limitations - OS compatibility, etc.](#limitations)
@@ -470,6 +471,16 @@ sensu::backend::config_hash:
 ```
 
 The first step will not fully add the node to the cluster until the second step is performed.
+
+### Large Environment Considerations
+
+If the backend system has a large number of resources it may be necessary to query resources using chunk size added in Sensu Go 5.8.
+
+```
+class { '::sensu::backend':
+  sensuctl_chunk_size => 100,
+}
+```
 
 ## Reference
 
--- a/lib/puppet/type/sensuctl_config.rb
+++ b/lib/puppet/type/sensuctl_config.rb
@@ -1,0 +1,35 @@
+require_relative '../../puppet_x/sensu/type'
+require_relative '../../puppet_x/sensu/array_property'
+require_relative '../../puppet_x/sensu/hash_property'
+require_relative '../../puppet_x/sensu/integer_property'
+
+Puppet::Type.newtype(:sensuctl_config) do
+  desc <<-DESC
+@summary Abstract type to configure other types
+DESC
+
+  newparam(:name, :namevar => true) do
+    desc "The name of the resource."
+  end
+
+  newparam(:chunk_size, :parent => PuppetX::Sensu::IntegerProperty) do
+    desc "sensuctl chunk-size"
+  end
+
+  # First collect all types with sensuctl provider that come from this module
+  # For each sensuctl type, set the class variable 'chunk_size' used by
+  # each provider to list resources
+  # Return empty array since we are not actually generating resources
+  def generate
+    sensuctl_types = []
+    Dir[File.join(File.dirname(__FILE__), '../provider/sensu_*/sensuctl.rb')].each do |file|
+      type = File.basename(File.dirname(file))
+      sensuctl_types << type.to_sym
+    end
+    sensuctl_types.each do |type|
+      provider_class = Puppet::Type.type(type).provider(:sensuctl)
+      provider_class.chunk_size = self[:chunk_size]
+    end
+    []
+  end
+end
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -93,6 +93,8 @@
 #   Hash of sensu_silenced resources
 # @param users
 #   Hash of sensu_user resources
+# @param sensuctl_chunk_size
+#   Chunk size to use when listing sensuctl resources
 #
 class sensu::backend (
   Optional[String] $version = undef,
@@ -136,6 +138,7 @@ class sensu::backend (
   Hash $roles = {},
   Hash $silencings = {},
   Hash $users = {},
+  Optional[Integer] $sensuctl_chunk_size = undef,
 ) {
 
   if $license_source and $license_content {
@@ -193,6 +196,10 @@ class sensu::backend (
     ensure  => $_version,
     name    => $cli_package_name,
     require => $::sensu::package_require,
+  }
+
+  sensuctl_config { 'sensu':
+    chunk_size => $sensuctl_chunk_size,
   }
 
   sensu_api_validator { 'sensu':
--- a/spec/acceptance/sensu_check_spec.rb
+++ b/spec/acceptance/sensu_check_spec.rb
@@ -54,6 +54,32 @@ describe 'sensu_check', if: RSpec.configuration.sensu_full do
     end
   end
 
+  context 'with chunk size' do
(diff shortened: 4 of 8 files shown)


treydock:
Flagged as WIP because I left some commented code in place as wanted to see how tests behave without it.  Also wasn't sure how we should document this.  I was thinking of adding section to README like "Large Environment examples" or something specific to scaling this module in large environments.

lib/puppet/type/sensuctl_config.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  newparam(:name, :namevar => true) do
+    desc "The name of the resource."
+  end
+
+  newparam(:chunk_size, :parent => PuppetX::Sensu::IntegerProperty) do
+    desc "sensuctl chunk-size"
+  end
+
+  def generate
```
ghoneycutt: i dont follow how this works

manifests/backend.pp
```diff
@@ -136,6 +138,7 @@
   Hash $roles = {},
   Hash $silencings = {},
   Hash $users = {},
+  Optional[Integer] $sensuctl_chunk_size = undef,
```
ghoneycutt: should we be more specific with regards to the integer? do the docs give a valid range?

spec/spec_helper.rb
```diff
@@ -64,6 +64,15 @@
   end
 }
 
+def sensuctl_types
```
ghoneycutt: what's going on here?

lib/puppet/type/sensuctl_config.rb
treydock: The `generate` function I think is intended to allow a type to define other types or in this case modify other types. What we end up doing here is setting class variable `chunk_size` for all types.  That is used when `sensuctl <resource> list` is executed.

manifests/backend.pp
treydock: Nothing in docs about a range, I would assume it's 0 to some large number.

https://docs.sensu.io/sensu-go/5.8/sensuctl/reference/#handling-large-datasets

spec/spec_helper.rb
treydock: This mimics logic in `sensuctl_config` type that finds all types that have provider `sensuctl` since that's the ones we want to deal with.  Rather than hardcode the list of types this finds them dynamically.

lib/puppet/type/sensuctl_config.rb
treydock: The logic on looking at files is listing all providers that use `sensuctl` since those are the ones we are modifying the class variable for.  This is one approach to make things dynamic, other modules I've seen use `generate` have just hardcoded the list of types that should be modified.

manifests/backend.pp
treydock: I just tested in Vagrant, `--chunk-size` takes negative numbers and 0 as well as expected positive integers.  All returned all resources. I am inclined not to impose a range if sensuctl isn't imposing one.

treydock:
Added one acceptance test that uses chunk size, also added small section to README and cleaned up commented out code.

lib/puppet/type/sensuctl_config.rb
ghoneycutt: since this is not standard to how the rest of the params work, could you document this in the code.

manifests/backend.pp
ghoneycutt: Looks like any integer is valid - https://github.com/sensu/sensu-go/search?q=ChunkSize&unscoped_q=ChunkSize

spec/spec_helper.rb
ghoneycutt: could you please document that in the code

README.md
```diff
@@ -471,6 +472,16 @@ sensu::backend::config_hash:
 
 The first step will not fully add the node to the cluster until the second step is performed.
 
+### Large Environment Considerations
+
+If the backend system has a large number of resources it may be necessary to query resources using chunk size added in Sensu Go 5.8.
+
+```
+class { '::sensu::backend':
+  sensuctl_chunk_size => 100,
```
ghoneycutt: @portertech what's a real world number we should use as an example?

lib/puppet/type/sensuctl_config.rb
treydock: Added comments to code

spec/spec_helper.rb
treydock: Documented in code

spec/acceptance/sensu_check_spec.rb
```diff
@@ -54,6 +54,32 @@
...
+        command       => 'check-cpu4.rb',
+        subscriptions => ['demo'],
+        handlers      => ['email'],
+        interval      => 60,
+      }
+      EOS
+
+      # Run it twice and test for idempotency
+      apply_manifest_on(node, pp, :catch_failures => true)
+      apply_manifest_on(node, pp, :catch_changes  => true)
```
ghoneycutt: should this test verify that chunk setting matches `1` ?

lib/puppet/type/sensuctl_config.rb
```diff
@@ -0,0 +1,34 @@
...
+    sensuctl_types = []
+    Dir[File.join(File.dirname(__FILE__), '../provider/sensu_*/sensuctl.rb')].each do |file|
+      type = File.basename(File.dirname(file))
+      sensuctl_types << type.to_sym
+    end
+    sensuctl_types.each do |type|
+      provider_class = Puppet::Type.type(type).provider(:sensuctl)
+      provider_class.chunk_size = self[:chunk_size]
+    end
+    []
```
ghoneycutt: what's with the empty array? more comments on what's happening in here would greatly help me out :)

spec/acceptance/sensu_check_spec.rb
treydock: We can't, acceptance tests execute the real commands so what we get back is all resources and no way to verify chunking was used since end result is the same as not using chunking.  The unit tests are what ensure we're actually adding the `--chunk-size` flag but no way to verify this with acceptance tests.

treydock: The idea behind using 1 was to ensure that even really small chunk size returns all records which is done because the catalog contains more than 1 resource and if chunking returned just 1 then it wouldn't be idempotent as Puppet would keep wanting to add one resource.

lib/puppet/type/sensuctl_config.rb
treydock: generate returns array of generated resources, we aren't actually generating resources so have to return empty array.

Added line to comments at start of function why returning empty array.

treydock:
Don't know if it helps but other places this trick of Puppet is used:

https://github.com/treydock/puppet-module-keycloak/blob/master/lib/puppet/type/keycloak_api.rb

The above modifies all Keycloak types to change how commands are executed by modifying class variables, not actually generating resources.

https://github.com/voxpupuli/puppet-splunk/blob/master/lib/puppet/type/splunk_config.rb

The Splunk module sets class variables and also generates resources.  The Splunk module is where I first saw this used and adapted it for Keycloak module.

ghoneycutt:
@treydock this failed appveyor

treydock:
I think  that failure was transient and I don't have permission to rerun the one that failed. I'll try a force push to trigger new build unless your able to re-run the one failure?

treydock:
@ghoneycutt Did forced push and everything appears fine.

PR #1115, opened 2019-06-15
Add ability to run acceptance tests against Sensu-Go CI builds

treydock:
# Pull Request Checklist

## Description

Add ability to run acceptance tests in Travis against Sensu-Go CI builds.

## Motivation and Context

This is going to allow this module to run full acceptance test suite against unreleased versions of Sensu-Go.


treydock:
Current travis matrix is very minimal to see if this actually works.  If we wanted to test all operating systems and both Puppet 5 and Puppet 6.

treydock:
Encrypting multiple files requires using single tar archive of all files that is then encrypted.  This is why previous license was removed, now in tar archive.  The script `ci_build.sh` is also in tar archive and that contains necessary logic to download and install CI build repos from package cloud.

treydock:
@ghoneycutt It seems that this PR has broken all other PRs because I had to modify the encryption key contents to support new encrypted file and travis CLI re-uses same key name so now the PRs without this change are broken.

I think what's left for this PR is to determine the matrix we want to use with CI builds.  Do we want to test the full matrix with CI builds or just a select few?  Also it seems allow_failures will not trigger a failure notification on slack.

treydock:
I may have fixed the issue, not sure how but managed to get different environment variables assigned by travis CLI but had to update #1117 with newly encrypted file.

treydock:
Removed WIP, added CI builds for all operating systems supported

.gitignore
```diff
@@ -53,3 +53,5 @@ doc/**
 
 # secrets
 tests/sensu_license.json
+tests/ci_build.sh
```
ghoneycutt: is this being auto generated? where?

treydock: It's in `tests/secrets.tar`. It has the secret token for packagecloud so only exists inside encrypted tar archive. We can only encrypt one file and now require two files be encrypted and travis docs say in that case to encrypt a tar archive with files needing to be kept secret.

ghoneycutt: Could you add instructions on how this file would be decrypted for local development

treydock: I added scripts and added those scripts to `tests/README.md`. The decryption of those files is only necessary if one wishes to run tests against a CI build.  The absence of the secrets being decrypted and certain env variables will just test latest production package.

tests/ci_build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
ghoneycutt: should this have `-e` so it fails on error

tests/decrypt-secrets.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+travis encrypt-file ${DIR}/secrets.tar.enc --decrypt --key $TRAVIS_CI_KEY --iv $TRAVIS_CI_IV
```
ghoneycutt: what's `IV` in the environment variable stand for?

tests/README.md
```diff
@@ -21,3 +21,17 @@ Copy certs from vagrant instance to this repo
 \cp -r /etc/puppetlabs/puppet/ssl/* /vagrant/tests/ssl/
 ```
 
+# Secrets
```
ghoneycutt: Could you add some more text around what files these touch and when you might need to use them.

tests/decrypt-secrets.sh
treydock: Part of openssl. I couldn't find in man page but googled it: https://stackoverflow.com/a/39412887. According to unofficial sources it's Initialization Vector which is part of the AES encryption.

tests/README.md
treydock: Added additional text. Also updated decrypt script to throw error if needed environment variables are not set.

tests/ci_build.sh
```diff
@@ -0,0 +1,15 @@
...
+
+which apt 1>/dev/null 2>&1
+if [ $? -eq 0 ]; then
+  script='script.deb.sh'
+else
+  script='script.rpm.sh'
+fi
+
+curl -s https://${CI_BUILD_TOKEN}:@packagecloud.io/install/repositories/sensu/ci-builds/${script} | bash 1>/dev/null 2>&1
+
```
ghoneycutt: Notice your files often have a trailing newline. This should not be there. Maybe an editor plugin can help?

PR #1122, opened 2019-07-13
Fix to support Puppetserver 5

### LINKED ISSUE #1102: Could not autoload
## Description of problem

- What did you do?
I've downloaded modul sensu and trying to install backend. Module was added to environment production on puppet server.

- What happened?
While run puppet, I get error:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Type-Name, Could not autoload puppet/type/sensu_configure: Could not autoload puppet/provider/sensu_configure/sensuctl: /etc/puppetlabs/code/environments/production/modules/sensu/lib/puppet/provider/sensuctl.rb:84: syntax error, unexpected tLABEL
  def self.sensuctl_set(command, name, property, value: nil, flags:  nil)
                                                       ^ at /etc/puppetlabs/code/environments/production/modules/sensu/manifests/backend/tessen.pp:19:7
```

I've tried to restart puppetserver, but error stay the same.

Expect sensu backend to be installed and configured.

Configuration in puppet:
`class { 'sensu::backend': }`

```
Puppetserver: puppetserver-2.8.1-1.el6
Puppet agent: puppet-agent-1.10.14-1.el6
```

- Your OS: Amazon Linux 
- Your Ruby version:
- Your version of Puppet: puppetserver-2.8.1-1.el6
- Your version of this module: master

treydock:
# Pull Request Checklist

## Description

Use syntax that works with JRuby 1.9.3 in order to support Puppetserver 5.

## Related Issue

Fixes #1102

### DIFF
--- a/lib/puppet/provider/sensuctl.rb
+++ b/lib/puppet/provider/sensuctl.rb
@@ -99,28 +99,6 @@ class Puppet::Provider::Sensuctl < Puppet::Provider
     self.class.sensuctl_delete(*args)
   end
 
-  def self.sensuctl_set(command, name, property, value: nil, flags:  nil)
-    args = [command]
-    args << "set-" + property.gsub('_', '-')
-    args << name
-    args << value unless value.nil?
-    args += flags unless flags.nil?
-    sensuctl(args)
-  end
-  def sensuctl_set(*args)
-    self.class.sensuctl_set(*args)
-  end
-
-  def self.sensuctl_remove(command, name, property)
-    args = [command]
-    args << "remove-" + property.gsub('_', '-')
-    args << name
-    sensuctl(args)
-  end
-  def sensuctl_remove(*args)
-    self.class.sensuctl_remove(*args)
-  end
-
   def self.sensuctl_auth_types()
     output = sensuctl(['auth','list','--format','yaml'])
     Puppet.debug("YAML auth list: #{output}")


lib/puppet/provider/sensuctl.rb
```diff
@@ -99,7 +99,7 @@ def sensuctl_delete(*args)
     self.class.sensuctl_delete(*args)
   end
 
-  def self.sensuctl_set(command, name, property, value: nil, flags:  nil)
+  def self.sensuctl_set(command, name, property, value = nil, flags =  nil)
```
ghoneycutt: Could you please add a comment that links to this PR letting folks know that puppetserver 5 uses jruby 1.9.3 or whatever and so we need it this way

treydock: I checked and we're not actually using this function. We may have used it early on before all resources supported creation through JSON files. Should I just delete this function?

treydock: There are two functions in `sensuctl.rb` that were helper functions no longer used. Deleting them wouldn't break anything.

ghoneycutt: If we can delete them, even better

treydock: Updated PR with functions removed.

PR #1123, opened 2019-07-13
Add acceptance tests that use puppetserver

### LINKED ISSUE #1120: Add acceptance tests that use Puppetserver
In several cases it seems we've run into issues where the Puppet code in this module is compatible with `puppet-agent` 5.x and 6.x but not `puppetserver` 5.x. This seems to be due to older Ruby version for puppetserver.

The acceptance tests for this module should be expanded to include testing the custom types/providers using full master/agent setup. Currently our tests only really test masterless so also testing interactions with puppetserver will ensure we are fully testing the more common master/agent setups.

One issue this would help with is #1102

treydock:
# Pull Request Checklist

## Description

Add acceptance tests that use puppetserver to test master/agent.

## Related Issue

Fixes #1120 also relates to #1102 

## Motivation and Context

We have run into issues where some behaviors with JRuby 1.9.3 in Puppetserver 5 will break this module so this should test both master and masterless Puppet.

## How Has This Been Tested?

The currently added acceptance tests covers backend,agent,facts and all type/provides. Testing just one OS is sufficient since what we're really testing is the puppetserver handling of catalogs.


treydock:
I marked as WIP as I'm still working to add master/agent for the cluster testing.  Currently the acceptance tests have been updated to cover everything except backend clustering.

treydock:
This has commit from #1122 in order to ensure the puppetserver acceptance tests will pass.

treydock:
Removed WIP, got cluster acceptance testing to work

lib/puppet/provider/sensuctl.rb
```diff
@@ -99,7 +99,7 @@ def sensuctl_delete(*args)
     self.class.sensuctl_delete(*args)
   end
 
-  def self.sensuctl_set(command, name, property, value: nil, flags:  nil)
+  def self.sensuctl_set(command, name, property, value = nil, flags =  nil)
```
ghoneycutt: did you confirm it fails without this?

treydock: I did, I ran beaker acceptance tests for puppetserver without this change and got errors.

Issue #1125, opened 2019-08-01
Feature Request: Change namespace property in sensu_check from string to array


csoleimani:
And automatically create the check in every namespace in the array. I'm currently doing this:
```
$namespaces = ['test1', 'test2']
  $namespaces.each |$namespace| {
  sensu_check {"${namespace}-check-hardware":
      command       => "${ruby} ${install_path}/check-hardware-fail.rb -L emerg, alert, crit",
      handlers      => lookup('page::for::production::only'),
      interval      => 60,
      annotations   => {
        'fatigue_check/occurrences' => '5',
      },
      subscriptions => 'hardware',
      namespace     => $namespace,
    }
  }
```

It would be nice if we could do this instead:

```

  sensu_check {'check-hardware':
    command       => "${ruby} ${install_path}/check-hardware-fail.rb -L emerg, alert, crit",
    handlers      => lookup('page::for::production::only'),
    interval      => 60,
    annotations   => {
      'fatigue_check/occurrences' => '5',
    },
    subscriptions => 'hardware',
    namespace     =>['test1', 'test2'],
    }
  }
```

Maybe rename namespace parameter to namespaces if this happens?

treydock:
A few comments.

First I'd prefer to add a new property called `namespaces` that would accept an array of namespaces. Most if not all properties of the types are directly mapped to resource properties in Sensu Go so changing `namespace` to be something different than what Sensu Go accepts is not ideal.

Also your second example wouldn't work because of how we prefetch resources from sensuctl. We look just at name so `test1` or `test2` would be seen and the other ignored if name is the same. I'll have to investigate if there's a way around this. I think this would require us to employ compound titles so the name would be seen as `check-hardware in test1`.

Lastly, because Puppet dictates state there is no way to find the state of a property like `namespaces` because it's not a state in Sensu Go. We'd have to do something like iterate over all checks and group ones with same name and different environments and build `namespaces` property. Not 100% sure that's possible.

treydock:
@csoleimani I am still investigating if something like a `namespaces` property would be possible.  I have fixed several issues with namespace handling in #1126 so now you could do something like this:

```puppet
$namespaces = ['test1', 'test2']
  $namespaces.each |$namespace| {
  sensu_check {"check-hardware in ${namespace}":
      command       => "${ruby} ${install_path}/check-hardware-fail.rb -L emerg, alert, crit",
      handlers      => lookup('page::for::production::only'),
      interval      => 60,
[...]

PR #1126, opened 2019-08-03
Better support for resources in different namespaces

treydock:
# Pull Request Checklist

## Description

Resources that can belong to a namespace now support composite names in form of "<name> in <namespace>". This change is backwards compatible. I did have to change how some properties are validated as doing `validate do` for an entire type was not working with composite names. The official Puppet docs for custom types only document the use of `pre_run_check` which essentially does the same thing, performs validation after the catalog is compiled but before the catalog is applied. This is how properties are validated against other properties.

This also resolves a bug where it would have been impossible to delete a resource if it was not in the `default` namespace.

## Related Issue

Partial fix for #1125 

## Motivation and Context

In Sensu Go resources that belong to namespaces can have the same name so long as they are in different namespaces. This change ensures the Puppet types support this same behavior by matching resources by name + namespace and not just name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@
     * [Sensu backend cluster](#sensu-backend-cluster)
         * [Adding backend members to an existing cluster](#adding-backend-members-to-an-existing-cluster)
     * [Large Environment Considerations](#large-environment-considerations)
+    * [Composite Names for Namespaces](#composite-names-for-namespaces)
 4. [Reference](#reference)
     * [Facts](#facts)
 5. [Limitations - OS compatibility, etc.](#limitations)
@@ -481,6 +482,35 @@ class { '::sensu::backend':
   sensuctl_chunk_size => 100,
 }
 ```
+
+### Composite Names for Namespaces
+
+All resources that support having a `namespace` also support a composite name to define the namespace.
+
+For example, the `sensu_check` with name `check-cpu in team1` would be named `check-cpu` and put into the `team1` namespace.
+
+Using composite names is necessary if you wish to have multiple resources with the same name but in different namespaces.
+
+For example to define the same check in two namespaces using the same check name:
+
+```puppet
+sensu_check { 'check-cpu in default':
+  ensure        => 'present',
+  command       => 'check-cpu.sh -w 75 -c 90',
+  interval      => 60,
+  subscriptions => ['linux'],
+}
+sensu_check { 'check-cpu in team1':
+  ensure        => 'present',
+  command       => 'check-cpu.sh -w 75 -c 90',
+  interval      => 60,
+  subscriptions => ['linux'],
+}
+```
+
+The example above would add the `check-cpu` check to both the `default` and `team1` namespaces.
+
+**NOTE:** If you use composite names for namespaces, the `namespace` property takes precedence.
 
 ## Reference
 
--- a/spec/unit/sensu_role_spec.rb
+++ b/spec/unit/sensu_role_spec.rb
@@ -28,6 +28,33 @@ describe Puppet::Type.type(:sensu_role) do
     }.to raise_error(Puppet::Error, 'Title or name must be provided')
   end
 
+  it 'should handle composite title' do
+    config.delete(:namespace)
+    config[:name] = 'test in dev'
+    expect(role[:name]).to eq('test in dev')
+    expect(role[:resource_name]).to eq('test')
+    expect(role[:namespace]).to eq('dev')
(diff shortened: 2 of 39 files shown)


spec/unit/sensu_role_spec.rb
```diff
@@ -28,6 +28,21 @@
     }.to raise_error(Puppet::Error, 'Title or name must be provided')
   end
 
+  it 'should handle composite title' do
```
ghoneycutt: what happens if a composite title and a namespace are specified? what happens if they are specified incorrectly, ie `foo in qa` is the title and the namespace is `default`.

what happens if the title is named funny? `in in qa` or `weird       spacing                   in qa` or they use hard tabs?

README.md
```diff
@@ -482,6 +483,33 @@ class { '::sensu::backend':
...
+All resources that support having a `namespace` also support a composite name to define the namespace.
+
+For example, the `sensu_check` with name `check-cpu in qa` would be named `check-cpu` and put into the `qa` namespace.
+
+Using composite names is necessary if you wish to have multiple resources with the same name but in different namespaces.
+
+For example to define the same check in two namespaces using the same check name:
+
+```puppet
+sensu_check { 'check-cpu in default':
```
ghoneycutt: if you don't specify a composite name would the namespace be specified in the type/provider as `default` or would it be unspecified?

spec/unit/sensu_role_spec.rb
treydock: For the first issue, `namespace` property takes priority but I'll add a test for that. So name `foo in qa` with `namespace` of `default` will keep namespace `default`.

If the pattern doesn't match it just puts the entire name in to `name` property and that becomes default for `resource_name` and `resource_name` doesn't allow spaces so those 2 examples would raise an error. We already test for this indirectly but I'll see if I can add some reasonable tests for this with compound aspect.

README.md
treydock: The default `namespace` is always `default` so if the name is just `check-cpu` the namespace is `default` if not specified.  I put the default in the name as thought it'd be confusing to mix non-composite and composite in the same example.

spec/unit/sensu_role_spec.rb
ghoneycutt: Thanks! Since this introduces a regex on the title, we want to ensure we are testing the regex as well as the edge cases of people mixing composite naming with namespaces.

Mixing could happen inadvertently with Hiera pretty easily.

treydock: Added some extra unit tests for composite names.

README.md
```diff
@@ -482,6 +483,33 @@ class { '::sensu::backend':
 }
 ```
 
+### Composite Names for Namespaces
+
+All resources that support having a `namespace` also support a composite name to define the namespace.
+
+For example, the `sensu_check` with name `check-cpu in qa` would be named `check-cpu` and put into the `qa` namespace.
+
+Using composite names is necessary if you wish to have multiple resources with the same name but in different namespaces.
```
ghoneycutt: Add something about if you use composite namespaces and specify `namespace` that `namespace` will take precedence. 

Perhaps also some guidance that if you use namespaces then you should follow this composite pattern. Ideally you wouldn't use namespaces for different environments, you would use hiera data. Is there another use case for the namespace? Don't want to confuse people.

treydock: I added a note about `namespace` property taking precedence.

I think namespaces can probably be different than puppet environments and unrelated to Hiera. Because roles define namespaces it's easy to imagine giving access controls to different namespaces that would all be defined in one Hiera environment so that the production sensu-backend could manage multiple namespaces for different roles or entities. As for composite vs non-composite, it's eluded to in README that the main benefit to composite is that you can have something like a check have the same name is Sensu for different environments. Besides having the same name in Sensu, there's not real benefit to using composite names if sites decide they want namespaces to be in the name they could just name checks `check-cpu-qa` or something.

ghoneycutt: so for the example should we change `qa` to something like `tenant1` or `namespace_foo`? I don't want to encourage folks to use namespace's for their deployment environments.

treydock: Ya I can change that to be something unique to a team or geographical location or something not specific to a sort of development workflow.

treydock: Updated examples in README.

ghoneycutt:
Looks great. Ready for merge once Travis is happy.

treydock:
Travis appears to be good now.

PR #1129, opened 2019-08-11
Do not resolve absent sensu_events

treydock:
# Pull Request Checklist

## Description

Attempt to fix occasional failures with `sensu_event` that may be due to timing.


treydock:
The real issue is that code was trying to resolve non-existent events so fixed type to not try and resolve an event if it doesn't exist.

Issue #1131, opened 2019-08-15
Unable to Create sensu_role_binding that maps to a ClusterRole


csoleimani:
- What did you do?
**Create a role binding, and referenced a cluster-role in the 'role_ref' parameter**
- What happened?
**The role_binding always assumes the type is Role,**
- What did you expect to happen?
**I want to either be able to specify role or cluster role.**
- How can someone reproduce the problem?
**use a sensu_role_binding resource type, and put a cluster role in the role_ref parameter**

- What does your puppet config look like (including any hiera config)

```
    sensu_role_binding { $team:
      role_ref  => 'edit_no_handlers',
      subjects  => {
        'type' => 'Group',
        'name' => "ad:${team}",
      },
      namespace => $team,
    }
```

I also tried turning role_ref into a hash with a 'name' and 'type' param like below, but puppet errors:

```
  sensu_role_binding { $team:
   role_ref  => {
        'name' => 'edit_no_handlers',
        'type' => 'ClusterRole'
      },
      subjects  => {
        'type' => 'Group',
        'name' => "ad:${team}",
      },
```
This is the error I'd get:

**Error: /Stage[main]/Profile::Sensugo::Server::Rbac/Sensu_role_binding[ts-corpsys]: Could not evaluate: sensuctl create ts-corpsys failed
Error message: Execution of '/usr/bin/sensuctl create --file /tmp/sensuctl20190815-12612-1gyy33s' returned 1: resource 0: (offset 57): json: cannot unmarshal object into Go struct field RoleRef.name of type string
Error: some resources couldn't be parsed**

- Is this a masterless or master based puppet setup?
Master
## Platform and version information

- Your OS: Debian 9
- Your Ruby version: 2.3.3
- Your version of Puppet:6.4.3
- Your version of Sensu: 5.11.1-5015
- Your version of this module: 978f6e2c6ce05d5d4e6b409e10678bebe14342c2 (namespaces branch)

treydock:
@csoleimani 

Your first example should work but I found the problem.  You need to ensure the `namespace` for `$team` exists. Notice in the error below the `PUT` has the namespace in URL which has to exist.  The types that use namespaces will auto-require the namespace if it's defined in Puppet.

BEFORE creating a namespace for `$team`:
```
Error: sensuctl create devs failed
[...]

PR #1132, opened 2019-08-17
Improved Validations

treydock:
# Pull Request Checklist

## Description

This improves several validations being performed.

First verify that `role_ref` for the role binding types is just a String. Those are the properties that don't match exactly to Sensu Go specs so this validations ensures we're getting just the role name and not a Hash.

The second validation is executed once the catalog is compiled and will check the catalog to ensure the defined `namespace` is in the catalog.

## Related Issue

This addresses some issues brought up in #1131

### DIFF
--- a/lib/puppet/provider/sensuctl.rb
+++ b/lib/puppet/provider/sensuctl.rb
@@ -116,5 +116,21 @@ class Puppet::Provider::Sensuctl < Puppet::Provider
     Puppet.debug("auth_types: #{auth_types}")
     auth_types
   end
+
+  def self.namespaces()
+    begin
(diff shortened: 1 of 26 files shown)


treydock:
I marked WIP because I want to see if it's possible to check both catalog and sensuctl for available namespaces.

treydock:
Namespace validation will check existing namespaces (defined outside Puppet) and ones defined in Puppet, one of those must exist to pass the validation

treydock:
Reverted commit that forced `role_ref` as string because that's going to have to be changed in upcoming pull request.

PR #1136, opened 2019-08-18
Support PostgreSQL datastore

### LINKED ISSUE #1135: Support event datastore
Sensu Go supports PostgreSQL for event datastore: https://docs.sensu.io/sensu-go/5.11/reference/datastore/

To help sites get started it would be handy to also manage PostgreSQL with Puppetlab's module so sites.

treydock:
# Pull Request Checklist

## Description

Add support for PostgreSQL datastore.

## Related Issue

Fixes #1135 

## Motivation and Context

Sensu Go 5.10 added support for PostgreSQL datastore that has to be setup with `sensuctl`. This adds support for that as well as optionally manages the actual postgresql database.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -10,6 +10,18 @@ fixtures:
       repo: git://github.com/puppetlabs/puppetlabs-yumrepo_core
       ref: 1.0.1
       puppet_version: ">= 6.0.0"
+    # Need by postgresql
+    augeas_core:
+      repo: git://github.com/puppetlabs/puppetlabs-augeas_core
+      ref: 1.0.4
+      puppet_version: ">= 6.0.0"
+    # Need by postgresql
+    concat:
+      repo: git://github.com/puppetlabs/puppetlabs-concat.git
+      ref: v6.0.0
+    postgresql:
+      repo: git://github.com/puppetlabs/puppetlabs-postgresql.git
+      ref: v6.0.0
     archive:
       repo: git://github.com/voxpupuli/puppet-archive.git
       ref: 'v3.0.0'
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@
     * [Advanced agent](#advanced-agent)
     * [Advanced SSL](#advanced-ssl)
     * [Enterprise support](#enterprise-support)
+    * [PostgreSQL datastore support](#postgresql-datastore-support)
     * [Installing Plugins](#installing-plugins)
     * [Installing Extensions](#installing-extensions)
     * [Exported resources](#exported-resources)
@@ -69,6 +70,8 @@ This module has a soft dependency on the [puppetlabs/apt](https://forge.puppet.c
 If using Puppet >= 6.0.0 there is a soft dependency on the [puppetlabs/yumrepo_core](https://forge.puppet.com/puppetlabs/yumrepo_core) module (`>= 1.0.1 < 2.0.0`) for systems using `yum`.
 
 If managing Windows there is a soft dependency on the [puppet/archive](https://forge.puppet.com/puppet/archive) module (`>= 3.0.0 < 4.0.0`).
+
+For PostgreSQL datastore support there is a soft dependency on [puppetlabs/postgresql](https://forge.puppet.com/puppetlabs/postgresql) module (`>= 6.0.0 < 7.0.0`).
 
 ### Beginning with sensu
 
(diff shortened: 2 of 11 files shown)


.fixtures.yml
```diff
@@ -10,6 +10,12 @@ fixtures:
       repo: git://github.com/puppetlabs/puppetlabs-yumrepo_core
       ref: 1.0.1
       puppet_version: ">= 6.0.0"
+    # Need by postgresql
+    concat:
+      repo: git://github.com/puppetlabs/puppetlabs-concat.git
```
ghoneycutt: this should use a tag

spec/acceptance/06_postgresql_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'spec_helper_acceptance'
+
+describe 'sensu::backend class', unless: RSpec.configuration.sensu_cluster do
```
ghoneycutt: How about a test that shows that after a client runs, there is expected data in the datastore. This would allow us to test from the user perspective and verify that the connection between the sensu backend and postgres are working as intended.

.fixtures.yml
treydock: Added a tag here.

spec/acceptance/06_postgresql_spec.rb
treydock: I added basic check that exits 0 to generate an event and check the event with and without postgresql support enabled.

.fixtures.yml
ghoneycutt: tag is not there.. forgot to push?

treydock: Fixed, not sure where it ended up but also squashed commits

.fixtures.yml
```diff
@@ -10,6 +10,18 @@ fixtures:
...
+      repo: git://github.com/puppetlabs/puppetlabs-augeas_core
+      ref: 1.0.4
+      puppet_version: ">= 6.0.0"
+    # Need by postgresql
+    concat:
+      repo: git://github.com/puppetlabs/puppetlabs-concat.git
+      ref: v6.0.0
+    postgresql:
+      repo: git://github.com/puppetlabs/puppetlabs-postgresql.git
+      ref: v6.0.0
```
ghoneycutt: 6.1.0 - be sure to check all the new additions that we are using the latest release

README.md
```diff
@@ -219,7 +222,40 @@ class { 'sensu::backend':
...
+**NOTE**: This features require a valid Sensu Go enterprise license.
+
+The following example will add a PostgreSQL server and database to the sensu-backend host and configure Sensu Go to use PostgreSQL as the event datastore.
+
+```puppet
+class { 'postgresql::globals':
+  manage_package_repo => true,
+  version             => '9.6',
+}
+class { 'postgresql::server': }
```
ghoneycutt: Is there a reason to use `class{}` with no params instead of `include` ?

spec/acceptance/sensu_check_spec.rb
```diff
@@ -68,13 +68,6 @@
         expect(data['metadata']['namespace']).to eq('test')
       end
     end
-
```
ghoneycutt: why would this change? it seems unrelated.

ghoneycutt:
This should also have a manifest for vagrant so it can easily be consumed there.

.fixtures.yml
treydock: This file has historically been the lower bound of version range, we test against the upper bound in `.fixtures-latest.yml`.

README.md
treydock: Because that's what puppetlabs documents. Also ensuring we don't do something to break the ability for people to use `class { } ` since that's easier to break than just `include`.

spec/acceptance/sensu_check_spec.rb
treydock: Because it's fragile. We add/remove checks in other files and to constantly having to update this number based on unrelated tests seems like a bad idea. Also we are directly testing that the checks we created all exist so checking the count just adds fragility to the acceptance tests.

treydock:
We can't use Vagrant because postgresql support requires a license, so for someone to use it in Vagrant would require they supply a license file.

treydock:
Do you want me to make it so an environment variable can define the path to license file or something? We don't currently have any enterprise licensed support in the vagrant images.

spec/acceptance/sensu_check_spec.rb
treydock: This became an issue when I added checks to test the postgresql still worked and so an external test to this file was impacting this test which sort of violates the idea that tests are self contained.

PR #1133, opened 2019-08-17
Support role_ref property being Hash

treydock:
# Pull Request Checklist

## Description

Allow `role_ref` to be String or Hash for `sensu_role_binding` and `sensu_cluster_role_binding`.

This is backwards compatible as `role_ref` can still be a string and munged to maintain old behavior.

## Related Issue

Fixes #1131

## Motivation and Context

The `role_ref` hash can have either `Role` or `ClusterRole` for both role bindings and cluster role bindings.

PR #1139, opened 2019-08-31
Testing improvements

treydock:
# Pull Request Checklist

## Description

Add a test for all contexts to ensure the catalog can compile.
Use parallel_spec to run unit tests
Update travis matrix to make it a bit clearer which test is which by using `$CHECK` pattern.

## Motivation and Context

I discovered that invalid resources would not cause test failures without testing that the catalog compiles. This noticed while working on #1133 and `role_ref` was still a String for some tests but it wasn't failing because we were not testing compile.

Using parallel_spec is causing the tests to go from 4 minutes to 1 minute on my desktop.

Using the `$CHECK` pattern is something that makes it really clear in Travis which tests failed since the env is what shows in the matrix UI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,20 +15,24 @@ before_install:
   - 'if [ -n "$encrypted_f942601034d6_key" -a -n "$encrypted_f942601034d6_iv" ]; then openssl aes-256-cbc -K $encrypted_f942601034d6_key -iv $encrypted_f942601034d6_iv -in tests/secrets.tar.enc -out tests/secrets.tar -d; cd tests && tar xvf secrets.tar ; fi'
 sudo: false
 
-script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec strings:generate reference'
+script: 'bundle exec rake $CHECK'
 
 matrix:
   fast_finish: true
   include:
-  - rvm: 2.4.4
-    env: PUPPET_GEM_VERSION="~> 5"
-  - rvm: 2.4.4
-    env: PUPPET_GEM_VERSION="~> 5" FIXTURES_YML=".fixtures-latest.yml"
+  - rvm: 2.4.5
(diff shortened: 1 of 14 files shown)


.travis.yml
```diff
@@ -15,19 +15,23 @@ before_install:
   - 'if [ -n "$encrypted_f942601034d6_key" -a -n "$encrypted_f942601034d6_iv" ]; then openssl aes-256-cbc -K $encrypted_f942601034d6_key -iv $encrypted_f942601034d6_iv -in tests/secrets.tar.enc -out tests/secrets.tar -d; cd tests && tar xvf secrets.tar ; fi'
 sudo: false
 
-script: 'SPEC_OPTS="--format documentation" bundle exec rake validate lint spec strings:generate reference'
+script: 'bundle exec rake $CHECK'
 
 matrix:
   fast_finish: true
   include:
   - rvm: 2.4.4
```
ghoneycutt: Puppet 5 is now using ruby 2.4.5. Want to go ahead and make that part of this?

https://puppet.com/docs/puppet/5.5/about_agent.html

https://puppet.com/docs/puppet/latest/about_agent.html

spec/classes/agent_spec.rb
```diff
@@ -157,6 +157,9 @@
         let(:pre_condition) do
           "class { 'sensu': manage_repo => false }"
         end
+        if facts[:os]['family'] != 'windows'
```
ghoneycutt: please add a comment as to why windows is not tested

.travis.yml
treydock: I replaced all 2.4.4 with 2.4.5.

spec/classes/agent_spec.rb
treydock: Comment added

PR #1140, opened 2019-09-02
Deprecate defining single asset builds

treydock:
# Pull Request Checklist

## Description

Deprecate ability to define assets via `url`, `sha512`, `filters` and `headers` property in favor of just using `builds` for defining assets with multiple builds. All previous properties still supported but deprecated ones will issue Puppet warning.

## Motivation and Context

Sensu Go 5.12 deprecated defining single assets not using `builds` so this change matches that deprecation.

### DIFF
--- a/lib/puppet/type/sensu_asset.rb
+++ b/lib/puppet/type/sensu_asset.rb
@@ -7,21 +7,6 @@ require_relative '../../puppet_x/sensu/integer_property'
 Puppet::Type.newtype(:sensu_asset) do
   desc <<-DESC
 @summary Manages Sensu assets
-@example Create an asset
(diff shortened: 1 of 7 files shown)


ghoneycutt:
Did the functionality just change or is there a deprecation process where the old and new way still work until the next major version bump of Sensu?

treydock:
Functionality did not change, but the docs were updated to primarily document multiple build and document deprecation of single builds. This change allows old behavior in Puppet but issues warning that old way is deprecated. I don't think Sensu Go issues warning except in docs.

treydock:
So I think next major version of Sensu Go will remove single builds.

PR #1141, opened 2019-09-02
BREAKING: Remove sensu_event and sensu_silenced types

treydock:
# Pull Request Checklist

## Description

Replace `sensu_event` and `sensu_silenced` with bolt tasks.

This is a breaking change.

## Motivation and Context

Both Sensu Go events and silenced entries are more for ad-hoc management and not a real state that is best suited for Puppet.


treydock:
I removed bolt tasks, they are not a breaking change and would be helpful to have them added in different pull request with other tasks. This pull request is just removing sensu_event and sensu_silenced

PR #1143, opened 2019-09-05
Replace unit test instance variables with let

treydock:
# Pull Request Checklist

## Description

Replace use of instance variables with `let` for unit tests. This is a better practice and something a tool like rubocop for rspec would have pointed out.

I opted to not modify the files for sensu_event and sensu_silenced to avoid conflicts with #1141. This will likely conflict with #1142 and #1140


ghoneycutt:
This would produce a lot of code churn. What will this get us?

treydock:
I've actually had issues in past where using instance variables instead of `let` would cause data to leak between test cases and required ugly hacks to reset variables between tests.  This is really just a future-proof change as well as best practice.

PR #1146, opened 2019-09-12
Fix sensu_asset to work with Sensu Go >= 5.13

treydock:
# Pull Request Checklist

## Description

Change how assets are queried to list all assets then use the `info` subcommand to get information about each asset.

## Related Issue

Part of #1145 

## Motivation and Context

It seems in Sensu Go 5.13 a change was made so that assets defined using the `builds` property do not show the builds property with `list` but rather show multiple assets. The only way to see the actual properties value is using `info` subcommand.

PR #1148, opened 2019-09-14
Fix sensu_plugin version insync? check

treydock:
# Pull Request Checklist

## Description

The previous logic of `insync?` for `sensu_plugin` version property was not capable of upgrades or downgrades when the new value for `version` was something other than `latest`.

I ran the acceptance tests with new changes but not updating `insync?` and the upgrade and downgrade tests failed because nothing actually happened.


treydock:
Tests will likely fail until #1146 is merged.

Issue #1145, opened 2019-09-10
Support Sensu Go 5.13


treydock:
This is issue that we need to support Sensu Go 5.13. There appear to be issues with the automated acceptance tests too that need to be addressed or issues opened with Sensu Go.

treydock:
Asset `builds` property is broken or changed: https://github.com/sensu/sensu-go/issues/3270

It appears that a single asset with `builds` is creating multiple objects in Sensu Go where `builds` is `null` and deprecated properties are set.

The ability to check for outdated bonsai assets appears to be broken so hard to add this feature: https://github.com/sensu/sensu-go/issues/3269

treydock:
This was handled with fixes available in 5.14, checking for outdated assets will be tracked in #1147

PR #1149, opened 2019-09-14
Add sensu_bonsai_asset type

treydock:
# Pull Request Checklist

## Description

Support installing Bonsai assets directly using `sensuctl asset add`. If `version` is set to `latest` the versions available are pulled from Bonsai API.

## Related Issue

Part of #1145 

## Motivation and Context

Sensu Go 5.13 added supporting for installing assets directly from Bonsai. This adds a type to support that operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@
         * [Adding backend members to an existing cluster](#adding-backend-members-to-an-existing-cluster)
     * [Large Environment Considerations](#large-environment-considerations)
     * [Composite Names for Namespaces](#composite-names-for-namespaces)
+    * [Installing Bonsai Assets](#installing-bonsai-assets)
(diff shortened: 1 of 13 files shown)


treydock:
Marked WIP as I need to figure out a way to prevent `sensu_asset` resource purging from deleting assets defined via this new type. It's possible with this current code that someone could get into a situation where Puppet is constantly adding bonsai assets then deleting that asset via sensuctl if they do resource purging.

Tests will fail until #1146 is merged.

treydock:
Removed WIP. The `sensu_bonsai_asset` type will generate a basic `sensu_asset` resource that doesn't change any properties in order to ensure resource purging won't delete the asset when the bonsai asset is defined.

PR #1156, opened 2019-09-22
Add sensu_gem package provider

### LINKED ISSUE #1155: allow to remove the prefix "^sensu-(plugins|extensions)" from sensu::plugins
## Description of problem

- What did you do? 

I installed `sensu-plugins-memory-checks`
one of the scripts contained in the plugin is called: `check-ram.rb`
`check-ram.rb` requires a gem called `vmstat`
You can check the documentation on Bonsai: [sensu-plugins-memory-checks](https://bonsai.sensu.io/assets/sensu-plugins/sensu-plugins-memory-checks)

- What happened? 

I tried to install the gem `vmstat` using `sensu::plugins` but it fails because it prepends the prefix and it's not possible to customize this prefix. 

I'd like to have a parameter called, for instance, `prefix`, which defaults to  `^sensu-(plugins|extensions)` but, for instance, it can be set to null: 
`prefix => '',`

treydock:
# Pull Request Checklist

## Description

Add `sensu_gem` package provider and add `gem_dependencies` parameter to `sensu::plugins`.

## Related Issue

Fixes #1155 .

## Motivation and Context

Allow Sensu plugins gem dependencies to be managed

### DIFF
--- a/manifests/plugins.pp
+++ b/manifests/plugins.pp
@@ -28,6 +28,8 @@
 # @param dependencies
 #   Package dependencies needed to install plugins and extensions.
 #   Default is OS dependent.
+# @param gem_dependencies
+#   Gem dependencies.
 # @param plugins
 #   Plugins to install
 # @param extensions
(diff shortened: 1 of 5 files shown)


manifests/plugins.pp
```diff
@@ -28,6 +28,8 @@
 # @param dependencies
 #   Package dependencies needed to install plugins and extensions.
 #   Default is OS dependent.
+# @param gem_dependencies
```
ghoneycutt: This should be added to the README since it is a common use case.

treydock: Added README example of usage

PR #1159, opened 2019-09-28
BREAKING: Rename sensu::backend url_host and url_port to api_host and api_port

treydock:
# Pull Request Checklist

## Description

Rename sensu::backend url_host and url_port to api_host and api_port

## Motivation and Context

This just renames the parameters to be more specific about their function. There are many URLs defined for backend and this just makes it clear the ones for these parameters are the API URL.


treydock:
Implemented in #1164 , closing this.

PR #1160, opened 2019-09-28
Support defining agent and backend service environment variables

treydock:
# Pull Request Checklist

## Description

Supporting setting environment variables that are loaded by agent and backend services.

## Motivation and Context

The daemons for both agent and daemon load environment variables from `/etc/default/` or `/etc/sysconfig/` files and this adds support for defining those environment variables. Also added support for doing the same with Windows by setting Windows environment variables.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -95,7 +95,7 @@ class sensu::agent (
   $_service_env_vars = $service_env_vars.map |$key,$value| {
     "${key}=\"${value}\""
   }
-  $_service_env_vars_content = ['# File managed by Puppet'] + $_service_env_vars
+  $_service_env_vars_content = ['# This file is being maintained by Puppet.','# DO NOT EDIT'] + $_service_env_vars
 
   if $facts['os']['family'] == 'windows' {
     $sensu_agent_exe = "C:\\Program Files\\sensu\\sensu-agent\\bin\\sensu-agent.exe"
@@ -176,6 +176,7 @@ class sensu::agent (
       notify    => Service['sensu-agent'],
     }
   }
+  # No built in way to read environment variables from a file for Windows
   if $facts['os']['family'] == 'windows' {
     $service_env_vars.each |$key,$value| {
       windows_env { $key:
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -22,6 +22,12 @@
 #   This parameter only used when `package_source` is an URL or when it's a puppet source (`puppet://`).
 # @param package_name
 #   Name of Sensu agent package.
+# @param service_env_vars_file
+#   Path to the agent service ENV variables file.
+#   Debian based default: `/etc/default/sensu-agent`
+#   RedHat based default: `/etc/sysconfig/sensu-agent`
+# @param service_env_vars
+#   Hash of environment variables loaded by sensu-agent service
 # @param service_name
 #   Name of the Sensu agent service.
 # @param service_ensure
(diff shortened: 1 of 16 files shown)


treydock:
WIP as sensu-docs PR documents all lower case ENV vars but testing shows all upper case is used.

https://github.com/sensu/sensu-docs/pull/1724

Also backend SENSU_* ENV vars are not loaded: https://github.com/sensu/sensu-go/issues/2680

treydock:
Testing shows the agent does load SENSU_* ENV vars and that is tested in acceptance tests.

ghoneycutt:
Waiting for those issues to be fixed upstream before we merge support.

treydock:
@ghoneycutt The required changes upstream were released in Sensu Go 5.15 and I re-ran tests and all pass now.

manifests/agent.pp
```diff
@@ -142,6 +154,30 @@
...
+      content   => join($_service_env_vars_content, "\n"),
+      owner     => $::sensu::sensu_user,
+      group     => $::sensu::sensu_group,
+      mode      => $::sensu::file_mode,
+      show_diff => $show_diff,
+      require   => Package['sensu-go-agent'],
+      notify    => Service['sensu-agent'],
+    }
+  }
+  if $facts['os']['family'] == 'windows' {
```
ghoneycutt: Does windows not support reading the values from a file like the other platforms? If so, that seems like a bug in sensu-go.

spec/classes/agent_spec.rb
```diff
@@ -63,6 +63,25 @@
           })
         }
 
+        service_env_vars_content = ['# File managed by Puppet'].join("\n")
```
ghoneycutt: Here's the header that I use everywhere

```
# This file is being maintained by Puppet.
# DO NOT EDIT
```

manifests/agent.pp
treydock: I think this is Windows issue because the application reading values from file is systemd or sysVinit on Linux. All sensu-go does is read environment variables.

ghoneycutt: perhaps just a comment that sensu-go does not read environment variables from a file on windows

spec/classes/agent_spec.rb
treydock: comments updated

manifests/agent.pp
treydock: comment added

spec/classes/agent_spec.rb
```diff
@@ -63,6 +63,25 @@
           })
         }
 
+        service_env_vars_content = ['# This file is being maintained by Puppet.','# DO NOT EDIT'].join("\n")
```
ghoneycutt: This is a programmatic way to build the content which is ok for the code, but not as great for the spec testing as it might miss something. Prefer doing it in such a way that you could copy/paste from an actual file and read it would be in the generated file. Here is an example.

https://github.com/ghoneycutt/puppet-module-bind/blob/master/spec/classes/init_spec.rb#L33-L78

treydock: That's harder to read in my opinion. I build the string using array then join with newlines and say `with_content(...)` which matches exact content. It's the same end result as your example but without the complication of handling indented HEREDOC.  Both approaches match exact content, so won't miss anything. Should I still change to the example from bind module?

ghoneycutt: Please do. I prefer the HEREDOC approach for testing so that you can read the document as you would when it is generated. For things larger than ~10 lines or so, I would use a fixture file.

treydock: Updated to use HEREDOC.

ghoneycutt:
Looks good! Needs rebase.

treydock:
Rebase pushed , just waiting on tests.

PR #1153, opened 2019-09-14
Add bolt tasks

treydock:
# Pull Request Checklist

## Description

Add bolt tasks to manage sensuctl events and silencings. These will replace sensu_event and sensu_silenced types removed in #1141 

## Related Issue

Part of #1147 

## Motivation and Context

Events and silencings are ad-hoc tasks so best to move into bolt tasks.

Issue #1147, opened 2019-09-14
Add Bolt tasks


treydock:
There are several bolt tasks that would be useful:

- [x] Manage events (resolve, delete) #1153 
- [x] Manage silencing #1153 
- [x] Execute checks #1153
- [x] Agent events #1153
- [x] List outdated assets #1162 
- [x]  Deploy agent defining backend, namespace and subscription #1153 
- [x] Inventory from sensu #1164

calebhailey:
And above all - use Bolt to deploy the agent, minimally configuring the backend(s) the agent should connect to, the namespace it should join, and the subscriptions it should subscribe to. 👍

treydock:
Looks like might be possible to define bolt inventory using external sources so in theory could set bolt inventory to be pulled from sensuctl: https://groups.google.com/forum/#!topic/puppet-users/cDYbir1JlgQ

PR #1162, opened 2019-10-31
Additional bolt tasks

treydock:
Added task to list outdated bonsai assets. This is part of fixing #1147 

Add a few comments that didn't get pushed with #1153 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@
     * [Large Environment Considerations](#large-environment-considerations)
     * [Composite Names for Namespaces](#composite-names-for-namespaces)
     * [Installing Bonsai Assets](#installing-bonsai-assets)
+    * [Bolt Tasks](#bolt-tasks)
 4. [Reference](#reference)
     * [Facts](#facts)
 5. [Limitations - OS compatibility, etc.](#limitations)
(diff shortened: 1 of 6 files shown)


treydock:
This PR is additional bolt tasks and a commit that was missing from #1153

ghoneycutt:
We should have a bolt tasks section of the readme that lists the tasks and explains what they do and how to use them, just like the basics for the classes.

treydock:
Wouldn't that duplicate the puppet strings auto generated docs? The docs are generated when pushed to forge based on the JSON metadata. I just wasn't sure how much we wanted to duplicate automatic docs. Still think warrants more than Strings docs?

treydock:
https://forge.puppet.com/sensu/sensu/tasks

I don't see tasks in the strings docs on github but when I generate the docs locally I see tasks getting put into REFERENCE.md and the docs directory. Maybe release for bolt tasks didn't receive automatic doc updates?

I'd prefer to rely on code to document things especially since metadata is a core part of bolt tasks. If there is something lacking from the auto docs that can't be added to the auto docs then I can update readme.

ghoneycutt:
We don't want the docs in the README to just replicate puppet-strings generated docs. Listing what the tasks are with a sentence or two about it let's folks you read the README understand what is possible from a glance.

treydock:
Added section to README

Issue #1163, opened 2019-11-02
Password changing and running `sensuctl configure` is fragile


treydock:
I have run into issues when testing where changing the password and going from SSL to non-SSL at the same time results in errors. The logic for `sensu_configure` to determine which password to use is a bit fragile.

treydock:
This is going to require using only the Sensu API to manage users because if the password is changed when SSL is changed then sensuctl config is removed to fix SSL settings but `sensuctl configure` won't work until the user password is updated and `sensu_user` comes after `sensu configure`. If the `sensu_user` resource uses just the API then it can be run before `sensuctl configure` which would also remove all bootstrap logic added in #1164

maxadamo:
how about changing these into `Sensitive` inside `backend.pp`? 
```
String $password = 'P@ssw0rd!',
String $agent_password = 'P@ssw0rd!',
Optional[String] $agent_old_password = undef,
String $postgresql_password = 'changeme',
```

treydock:
That would be useful but not solve the problem. There is a fix in #1164 . Should be noting that using the API directly didn't help, real fix was just extra provider logic in sensu_configure.

maxadamo:
yep, sorry. I didn't meant to provide a solution, but I just wanted to add another topic.

maxadamo:
my two cents on the Sensitive datatype. In my modules, in order to avoid breaking changes, for one or two release, I use `Variant[String|Sensitive]` and if `string` is being used I emit a warning. After a couple of release I force the usage of `Sensitive`

PR #1164, opened 2019-11-02
BREAKING: Move cli resources to sensu::cli class

treydock:
# Pull Request Checklist

## Description

Move cli resources to sensu::cli class
Manage sensu user/group so to support sensu::cli on host not running agent or backend

* Move sensu::backend::cli_package_name to sensu::cli::package_name
* Move sensu::backend::url_host to sensu::api_host
* Move sensu::backend::url_port to sensu::api_port
* Move sensu::backend::password to sensu::password
* Move sensu::backend::old_password to sensu::old_password
* Move sensu::backend::agent_password to sensu::agent_password
* Move sensu::backend::agent_old_password to sensu::agent_old_password
* Move sensu::backend::sensuctl_chunk_size to sensu::cli::sensuctl_chunk_size
* Replace `sensu_config` type with `sensu::cli::config_format` and `sensu::cli::config_namespace` parameters.

Always manage `sensu:agent` user to ensure password set
Set `password` in agent.yml

## Motivation and Context

This will allow non-backends to have sensuctl setup which is supported by Sensu Go. This also would make it easier to have a Bolt task that can use sensuctl to populate Bolt Inventory v2.

The management of sensu user/group was necessary to support using sensu::ctl class on a host that doesn't include sensu::agent or sensu::backend as the agent and backend packages had previously been what adds the user/group.


treydock:
Marked WIP as needs support for Windows and possibly addressing #1163 as it was the acceptance tests for sensu::cli where I ran into #1163.

manifests/cli.pp
```diff
@@ -0,0 +1,72 @@
...
+
+  sensu_configure { 'puppet':
+    url                => $url,
+    username           => 'admin',
+    password           => $password,
+    bootstrap          => $bootstrap,
+    bootstrap_password => 'P@ssw0rd!',
+    trusted_ca_file    => $trusted_ca_file,
+  }
+
```
ghoneycutt: no empty lines between `}`

manifests/cli.pp
```diff
@@ -0,0 +1,72 @@
...
+  sensuctl_config { 'sensu':
+    chunk_size => $sensuctl_chunk_size,
+  }
+
+  sensu_configure { 'puppet':
+    url                => $url,
+    username           => 'admin',
+    password           => $password,
+    bootstrap          => $bootstrap,
+    bootstrap_password => 'P@ssw0rd!',
```
ghoneycutt: should be a param

README.md
```diff
@@ -130,6 +131,19 @@ associated to `linux` and `apache-servers` subscriptions.
...
+
+```puppet
+class { '::sensu::cli':
+  url_host => 'sensu-backend.example.com',
+  password => 'supersecret',
+}
+```
+
+**NOTE**: The `sensu::backend` class calls the `sensu::cli` class so it is only necessary to directly call the `sensu::cli` class on hosts not using the `sensu::backend` class.
+
```
ghoneycutt: We should further explain what bootstrapping entrails and why it is needed.

manifests/cli.pp
treydock: I would disagree, changing the value would not work. This value is the default value from Sensu Go and anything but this value will break the module on fresh installs. We have to bootstrap using default password.

README.md
treydock: The bootstrapping only happens on fresh install of backend. It's something we abstract away and not something I would expect anyone to need to worry about since we just recommend not using default password. The bootstrapping is not needed for just using cli class like this section documents. This is kinda of like a private type functionality. I can document but if module behaves correctly people should be unaware of bootstrap step.

treydock: Also I hope to rework some the logic to avoid need to pass bootstrap parameters. The current process is semi fragile per #1163. Somewhere still have to hardcode the default bootstrap password though.

manifests/cli.pp
treydock: Fixed.

manifests/cli.pp
treydock: To avoid confusion I will hardcode the value into the Ruby code where it's used. If the value is changed by anyone using this module it will break that functionality on fresh installs so probably better to avoid confusion and just not expose it at all as parameter. The only time we'd have to change the value is if upstream Sensu Go starts using a different default password.

manifests/cli.pp
```diff
@@ -0,0 +1,113 @@
...
+# @param install_path
+#   Where to install sensuctl for Windows. Default to `C:\Program Files\Sensu`.
+# @param url_host
+#   Sensu backend host used to configure sensuctl and verify API access.
+# @param url_port
+#   Sensu backend port used to configure sensuctl and verify API access.
+# @param password
+#   Sensu backend admin password used to confiure sensuctl.
+# @param bootstrap
+#   Should sensuctl be bootstrapped. This is a private parameter used by sensu::backend class.
```
ghoneycutt: it is used in this class, too

manifests/cli.pp
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  if $configure {
+    sensuctl_config { 'sensu':
+      chunk_size => $sensuctl_chunk_size,
+      path       => $sensuctl_path,
+    }
+
+    sensu_configure { 'puppet':
```
ghoneycutt: is this duplicating code in `sensu::bootstrap` ?

README.md
ghoneycutt: This can be a commented out section in the code or the README as you see fit, though a few sentences regarding How Sensu ships with a default password and we have to go through this bootstrap process to change the password would help users understand what is happening.

manifests/cli.pp
treydock: I found a way to avoid bootstrap parameters. This has been removed.

manifests/cli.pp
treydock: This used to be in `sensu::backend` and was moved here and backend just class `cli` class.  I removed all bootstrapping parameters from this PR.

manifests/cli.pp
treydock: Bootstrap parameters removed.

README.md
treydock: I improved the bootstrap process so it's entirely hidden. We document using non-default password and everything else is handled by the provider for `sensu_configure`.  See d2ef64b07805262733c87ccd2bd28de89d2c9a4f

treydock:
Removed WIP. I will fix #1163 in a different pull request that will have to incorporate this PR. The changes necessary to fix #1163 are rather involved and potentially not backwards compatible as a parameter will have to get removed.

treydock:
I am struggling to get simple "Does file exist?" to work in appveyor. I got 2 appveyor instances stuck on RDP but didn't accept my password so have to wait till they time out. I'll cleanup windows testing once I'm able to debug why checking file existence is not working.

.fixtures.yml
```diff
@@ -36,5 +36,8 @@ fixtures:
     archive:
       repo: git://github.com/voxpupuli/puppet-archive.git
       ref: 'v3.0.0'
+    windows_env:
```
ghoneycutt: Why is this needed?

appveyor.yml
```diff
@@ -48,16 +49,19 @@ test_script:
...
 environment:
+  APPVEYOR_RDP_PASSWORD: Sensu1
   matrix:
     - PUPPET_GEM_VERSION: '~>5.x'
       PUPPET_REPO: puppet5
       RUBY_VERSION: 24-x64
-    - PUPPET_GEM_VERSION: '~>6.x'
-      PUPPET_REPO: puppet6
-      RUBY_VERSION: 25-x64
+#    - PUPPET_GEM_VERSION: '~>6.x'
```
ghoneycutt: why are these commented out?

lib/puppet/provider/sensu_api.rb
```diff
@@ -0,0 +1,28 @@
...
+    else
+      Puppet.notice "Error contacting #{auth_test_url}"
+      return false
+    end
+  rescue Exception => e
+    Puppet.notice "Unable to connect to #{auth_test_url}: #{e.message}"
+    return false
+  end
+end
+
```
ghoneycutt: trailing newline

lib/puppet/type/sensu_configure.rb
```diff
@@ -35,11 +35,6 @@
     desc "Password to use with sensuctl configure"
   end
 
-  newparam(:bootstrap_password) do
```
ghoneycutt: Changing the public interface feels like a breaking change.

manifests/backend.pp
```diff
@@ -176,6 +179,7 @@
   Stdlib::Port $postgresql_port = 5432,
   String $postgresql_dbname = 'sensu',
   Integer $postgresql_pool_size = 20,
+  Boolean $declare_cli_class = true,
```
ghoneycutt: This is a new patter, why should the user care how the class is included?

.fixtures.yml
treydock: This is needed to set `PATH` for sensuctl on Windows since the install isn't a package, it's just a zip file.

appveyor.yml
treydock: See my recent comment yesterday, I am debugging appeveyor failures and it's easier to iterate over changes with fewer changes running. I had to pause my work last night due to issues with RDP in appveyor builds that allows me to log into VM and see why tests are failing. I will revert these changes once I get appveyor working.

lib/puppet/type/sensu_configure.rb
treydock: If anyone ever changed this value then it would have resulted in errors. The default value was the only value that would work as it's the Sensu Go default admin password. Also we do not document the use of `sensu_configure` as it's kind of a private type that we use internally in the backend class and now moved to the cli class. If anyone changed this value then their install would have not worked so I think it's safe to consider this backwards compatible.

manifests/backend.pp
treydock: Because if someone wants to do this:

```puppet
class { '::sensu::cli': ... }
class { '::sensu::backend': ... }
```

Then they will duplicate resource errors unless they set `declare_cli_class => false`.

ghoneycutt: Could we use data in module to work around this?

treydock: I don't know how that would work? This is something I think we just have to allow someone to turn on/off based on how they use this module.

treydock: I think only way to avoid this would be to do something like `if ! defined(Class['sensu::cli'])` but I thought such order dependent hacks were discouraged in Puppet.

lib/puppet/provider/sensu_api.rb
treydock: Fixed.

treydock:
Fixed Windows tests and restored previous test matrix. This PR should be good to go once Travis tests complete.

manifests/backend.pp
ghoneycutt: Why do we have to use `class` at all instead of using`include`?

treydock: Because then instead of this:

```
class { 'sensu::backend': url_host => 'foo' }
```

A site has to do this with just include: 

```
class { 'sensu::cli': url_host => 'foo' }
class { 'sensu::backend': url_host => 'foo' }
```

There are parameters that we pass to backend class that are also needed by CLI class and this maintains backwards compatibility for behavior too.

treydock: One idea is to move `url_host` and `url_port` to `sensu` class since shared between backend agent and now CLI. This would be easier once #1159 is merged and #1159 will have to be updated once this PR is merged. If we move these and other parameters to `sensu` class we could simplify with `include` but would be breaking change. I can make that change as part of #1159

ghoneycutt: I like that approach. Let's do this as part of the breaking change.

treydock: So implement that breaking change in this PR or keep this current behavior and do new PR for breaking change?

treydock: This was implemented.

treydock:
Closed #1159 as this had similar changes that made that PR not necessary.

treydock:
I've rebased with master and added Bolt v2 inventory task that requires the CLI changes in order to run inventory on non-backend node.

treydock:
Added commit to make managing agent passwords much more consistent. Updated PR description as moved `agent_password` and `agent_old_password` from `sensu::backend` class to `sensu` class. This way both backends and agents could use same value and ensure that value is propagated to `sensu:agent` user and also updated in `agent.yml`. Previous behavior was flawed and made changing the password very difficult.

treydock:
Added commit to consolidate `sensu_config` which was just a way to set a few `sensuctl` settings and moved those settings into `sensuctl_configure` and exposed them via `sensu::cli` class. The previous `sensu_config` would allow invalid settings and had no validation and didn't really do much that couldn't be handled by the type managing sensuctl configuration.

PR #1168, opened 2019-11-09
Support Windows 2019

treydock:
# Pull Request Checklist

## Description

Add Support for Windows 2019

## Motivation and Context

Sensu Go says "2008 R2 and later" and now Appveyor has Windows 2019 images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -721,6 +721,10 @@ Linux.
 * Ubuntu 18.04 LTS
 * Amazon 2018.03
 * Amazon 2
+* Windows Server 2008 R2
+* Windows Server 2012 R2
+* Windows Server 2016
+* Windows Server 2019
 
 ## Development
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -51,6 +51,8 @@ image:
   - Visual Studio 2015
   # Windows 2016
   - Visual Studio 2017
+  # Windows 2019
+  - Visual Studio 2019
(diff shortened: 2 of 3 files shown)


ghoneycutt:
This is great! This should also be in the README under Limitations.

treydock:
Limitations doesn't document supported platforms, but a different place in README does and added Windows versions to that list.

README.md
```diff
@@ -721,6 +721,10 @@ Linux.
 * Ubuntu 18.04 LTS
 * Amazon 2018.03
 * Amazon 2
+* Windows Server 2008 R2
+* Windows Server 2012 R2
+* Windows Server 2016
+* Windows Server 2019
```
ghoneycutt: We should get it added to Vagrant as well.

ghoneycutt:
An appveyor test is failing and I can't tell if it is transient or not. Not able to restart the job. Should restart though when you rebase/squash with the Vagrant addition.

treydock:
There are no reliable Windows 2019 Vagrant boxes. I haven't managed to get any to work. I think we will just have to rely on Appveyor for testing Windows 2019.  The appveyor failure is transient but I can't see output until #1164 is merged which fixes appveyor to not omit output.  This PR just added items to test matrix and one of the old tests failed so see no way it could be related, especially since previous commit was passing and all I changed was README.

PR #1171, opened 2019-11-12
Add bolt task to manage API keys

treydock:
# Pull Request Checklist

## Description

Add bolt task to create, list and delete API keys for Sensu Go.

## Motivation and Context

Sensu Go 5.15 (not yet released) will have support for managing API keys. The way they are created and defined the key is generated at creation time thus can't be managed as a custom type/provider so must be handled via Bolt tasks where Bolt can return the key when created as part of the output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -616,6 +616,12 @@ The following Bolt tasks are provided by this Module:
 
 Example: `bolt task run sensu::agent_event name=bolttest status=1 output=test --nodes sensu_agent`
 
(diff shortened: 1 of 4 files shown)


treydock:
Tests will fail until Sensu Go 5.15 is released. The tests based on CI builds should pass though.

ghoneycutt:
Sensu-go 5.15 has been released, merging.

PR #1172, opened 2019-11-12
Update steps to release software

ghoneycutt:


ghoneycutt:
Updating docs after automating release process.

PR #1173, opened 2019-11-16
Improve name validations to match Sensu Go

treydock:
# Pull Request Checklist

## Description

Update all resource name validations to match Sensu Go validations for names.

## Motivation and Context

Sensu Go 5.15 (not yet released) added a colon as valid name character. This PR updates updates regex used by types but in a way where most types share the same regex so future updates require fewer code changes.

Upstream change: https://github.com/sensu/sensu-go/pull/3384/files

Any changes to our regex should be considered a bugfix as previous lack of regex or incorrect regex would have allowed Puppet to define things that Sensu Go would reject.

### DIFF
--- a/lib/puppet/type/sensu_asset.rb
+++ b/lib/puppet/type/sensu_asset.rb
@@ -76,6 +76,7 @@ DESC
   newparam(:resource_name, :namevar => true) do
     desc "The name of the asset."
     validate do |value|
+      # Must only contain lower case letters, numbers, underscores, periods, hyphens and colons
       unless value =~ %r{^[a-z0-9\/\_\.\-\:]+$}
         raise ArgumentError, "sensu_asset name invalid"
       end
--- a/lib/puppet/type/sensu_asset.rb
+++ b/lib/puppet/type/sensu_asset.rb
@@ -76,7 +76,7 @@ DESC
   newparam(:resource_name, :namevar => true) do
     desc "The name of the asset."
     validate do |value|
-      unless value =~ /^[\w\.\-\/]+$/
+      unless value =~ %r{^[a-z0-9\/\_\.\-\:]+$}
         raise ArgumentError, "sensu_asset name invalid"
       end
     end
--- a/lib/puppet/type/sensu_user.rb
+++ b/lib/puppet/type/sensu_user.rb
@@ -44,6 +44,7 @@ DESC
   newparam(:name, :namevar => true) do
     desc "The name of the user."
     validate do |value|
+      # Match only upper case letters, lowercase letters, numbers, underscores, periods and hyphens
       unless value =~ %r{^[\w.\-]+$}
         raise ArgumentError, "sensu_user name invalid"
       end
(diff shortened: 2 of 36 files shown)


treydock:
Marked WIP until Sensu Go 5.15 is released

lib/puppet/type/sensu_user.rb
```diff
@@ -44,7 +44,7 @@
   newparam(:name, :namevar => true) do
     desc "The name of the user."
     validate do |value|
-      unless value =~ /^[\w\.\-]+$/
+      unless value =~ %r{^[\w.\-]+$}
```
ghoneycutt: I thought users can't have `.` or `:` in the username

lib/puppet/type/sensu_asset.rb
```diff
@@ -76,7 +76,7 @@
   newparam(:resource_name, :namevar => true) do
     desc "The name of the asset."
     validate do |value|
-      unless value =~ /^[\w\.\-\/]+$/
+      unless value =~ %r{^[a-z0-9\/\_\.\-\:]+$}
```
ghoneycutt: why does asset use a regex instead of `PuppetX::Sensu::Type.name_regex` ?

lib/puppet/type/sensu_user.rb
ghoneycutt: should this be a type?

ghoneycutt:
Please add a comment next to any regex that describes what the regex is meant to match.

lib/puppet/type/sensu_user.rb
treydock: They can not, this regex matches regex from Sensu Go which allows period but not colon.

This is a one-off regex not shared by other types so kept separate.

lib/puppet/type/sensu_asset.rb
treydock: Because this is how Sensu Go does it. The name regex used by Asset type is not the one shared by other resources. The Asset does not support uppercase letters and also allows slashes when others don't share that pattern. I pulled regex from Sensu Go source.

treydock:
It's really difficult to document a regex that isn't just the regex expanded with commas. Describing `%r{^[a-z0-9\/\_\.\-\:]+$}` ends up looking like `# Match a-z, 0-9, /, _, ., -, and :` which maybe more confusing than reading the regex. Without commas then the comment is the same as the regex and seems not necessary.

treydock:
Added commit for comments but can revert since they may cause confusion due to commas and essentially documenting a regex that's pretty clear.

treydock:
@ghoneycutt Removed WIP. I'm concerned the regex comments will cause confusion due to commas, the regex'es themselves are not overly complicated, so wonder if we should just document valid in tests which is done already.

lib/puppet/type/sensu_asset.rb
ghoneycutt: Each regex should be documented in a way that is easy to read. By using symbols and commas it makes it hard to understand. Below is an example comment for a regex.

Match strings that begin with and contain only letters, numbers, underscores, periods and colons. The string may also be empty.

treydock: Updated comments per suggestion.

PR #1175, opened 2019-11-23
Add support for Sensu etcd replicator

treydock:
# Pull Request Checklist

## Description

Add new type `sensu_etcd_replicator`.

## Related Issue

Part of #1174 

## Motivation and Context

Sensu Go added support for Etcd replicators and this adds a custom type to manage replicators.

PR #1176, opened 2019-11-23
Move PostgresConfig to a type

treydock:
# Pull Request Checklist

## Description

Move the configuration of Postgres in Sensu to a native type.

## Motivation and Context

This replaces an file + exec with a custom type to make the postgres config management more streamlined.

This is backwards compatible as no behavior will change except if a site used this code we will remove a file that's no longer needed that was fed into sensuctl from the private class.  All parameters and behavior remain the same, just replacing File + Exec with Sensu_postgres_config type.

### DIFF
--- a/lib/puppet/provider/sensuctl.rb
+++ b/lib/puppet/provider/sensuctl.rb
@@ -72,7 +72,7 @@ class Puppet::Provider::Sensuctl < Puppet::Provider
     data
   end
 
-  def self.sensuctl_create(type, metadata, spec, api_version = 'core/v2')
+  def self.resource_file(type, metadata, spec, api_version = 'core/v2')
     data = {}
     data['type'] = type
     data['api_version'] = api_version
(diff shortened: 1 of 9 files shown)


treydock:
This will have to be rebased if #1175 is merged first or that will have to rebased if this merged first. Both add the `dump` function.

treydock:
Actually #1175 should get merged first because it has tests for `dump` while this PR just has the function.

lib/puppet/provider/sensuctl.rb
```diff
@@ -117,6 +128,17 @@ def self.sensuctl_auth_types()
     auth_types
   end
 
+  def self.dump(resource_type)
+    output = sensuctl(['dump',resource_type,'--format','yaml','--all-namespaces'])
+    Puppet.debug("YAML dump of #{resource_type}:\n#{output}")
+    resources = []
+    dumps = output.split('---')
+    dumps.each do |d|
+      resources << YAML.load(d)
```
ghoneycutt: why are we doing yaml output here though using json output above?

manifests/backend/datastore/postgresql.pp
```diff
@@ -5,58 +5,24 @@
   include ::sensu
   include ::sensu::backend
 
-  $config_path = "${::sensu::etc_dir}/postgresql.yaml"
   $user = $::sensu::backend::postgresql_user
   $password = $::sensu::backend::postgresql_password
   $host = $::sensu::backend::postgresql_host
   $port = $::sensu::backend::postgresql_port
   $dbname = $::sensu::backend::postgresql_dbname
-  $config = {
```
ghoneycutt: if you were specifying $config options, wouldn't those be lost?

lib/puppet/provider/sensuctl.rb
treydock: Because `sensuctl dump` doesn't support `--format json`. It supports `wrapped-json` and `yaml` and neither can be parsed by Ruby without hacks and easier to split the YAML output on `---` and then ingest each item.

manifests/backend/datastore/postgresql.pp
treydock: That was a private variable that was not accessible to anyone, it's not a parameter. It was only possible to change and the value based on parameters that have not changed. The `$config` was used to generate YAML which is no longer needed.

lib/puppet/provider/sensuctl.rb
ghoneycutt: Isn't there an open issue you raised against sensu-go related to this? Suggest commenting that this is done because of what you noted and if there is an issue link add that, too.

treydock: Sort of? The issue I opened pre-dated the `dump` command. Taking the YAML and splitting on `---` isn't that difficult or expensive. This feature works regardless. I'd rather not wait on upstream to support something we can work around. Opened new issue for this but I think this PR should be good to go without upstream making our lives easier.

https://github.com/sensu/sensu-go/issues/3424

treydock: Also #1175 uses the dump + YAML which is all we can do for that PR. Just the behavior of upstream and we can remove the YAML code if / when upstream updates to support json for dump.

ghoneycutt: Using the split and yaml is fine, I just want to document why we do it next to the code along with the link to the issue you noted.

treydock: Added comments to all locations where we execute `sensuctl dump`.

PR #1177, opened 2019-11-23
Add lint plugin

ghoneycutt:
## Description
Add a puppet-lint plugin to check for the presence of a newline character at the end of a file.

## Motivation and Context

To ensure style compliance and lessen back and forth with pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
   - gem update bundler
   - gem --version
   - bundle -v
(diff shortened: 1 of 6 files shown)


treydock:
That plugin just ensures `\n` is last character of manifest, it won't catch extra newlines where file ends with `\n\n`.

ghoneycutt:
Right, though I saw we were not using it and should.

Issue #1183, opened 2019-11-25
info about client configuration


bovy89:
Hi,

using sensuclassic, we can dynamically add subscriptions (sensuclassic::subscription), also with custom attributes, and write custom client config (sensuclassic:write_json).

In this way, we can include sensuclassic module in the "root" hostgroup with default subscriptions and client attributes (we use foreman as ENC), while leaving other subscriptions and custom attributes inside other modules (e.g. tomcat subscription is inside our tomcat puppet module). This is possible due sensu config hash merge.

How can we translate this in sensu go (subscription, label, annotation etc)? This is a crucial feature for us.

Thanks

treydock:
Does Foreman ENC do any merging? Merging in the Puppet module isn't necessary as we just write out a hash from `config_hash` to agent YAML. Here you can specify subscriptions, labels and annotations. You could specify a default for top level hostgroup then change else where. If you used Hiera then be very easy to merge values, just not sure if Foreman supports it.

I will open PR to make setting important values like subscription a bit easier.

treydock:
Also be useful to see example of Foreman ENC yaml for sensuclassic and some manifest code that illustrates the different places you're setting something like subscription.

ghoneycutt:
Perhaps a profile class is in order to potentially `lookup()` and munge the data?

bovy89:
Just an example.


These are my modules.

```
mybasemodule/
├── manifests
    ├── init.pp

    # init.pp
    class mybasemodule {
        class { '::sensuclassic':
            <other client params here>
            subscriptions => ['all', 'linux'],
            client_custom => {
                'environment' => 'prod',
            },
        }
    }

myapachemodule/
├── manifests
    ├── init.pp

    # init.pp
    class myapachemodule {
        class { '::apache':
            server_signature => 'Off',
        }

        sensuclassic::subscription {'httpd': }
    }
```


In foreman I do the following (we are not using yaml):

```
include ::mybasemodule
include ::myapachemodule
```

This generate 2 config files: 1 for client and 1 for every subscription (sensu hash config merge will do the magic)

This allow us to dynamically add things (like subscription) without modify mybasemodule or its parameters

treydock:
This is a fundamental change from Sensu to Sensu Go. The old way used files and you could deploy many files and they all got merged together. The new Sensu Go way is either defining agent config via a single YAML file or by modifying the agent's entity with sensuctl. The sensuclassic module approach simply is not possible with Sensu Go.

[...]

PR #1185, opened 2019-11-26
Add several parameters to sensu::agent class

treydock:
# Pull Request Checklist

## Description

Add several parameters to `sensu::agent` that are the more important configuration options. The old way of using `config_hash` is still valid and takes priority and is documented and unit tested.

## Related Issue

Was thought of while thinking of solution for #1183 but won't actually solve the problem

## Motivation and Context

The very important config options for agents should be first class parameters with data types.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -5,7 +5,7 @@ fixtures:
       ref: 5.0.1
     stdlib:
       repo: git://github.com/puppetlabs/puppetlabs-stdlib.git
-      ref: 4.25.1
+      ref: 5.1.0
     yumrepo_core:
       repo: git://github.com/puppetlabs/puppetlabs-yumrepo_core
       ref: 1.0.1
--- a/README.md
+++ b/README.md
@@ -179,10 +179,8 @@ associated to `linux` and `apache-servers` subscriptions.
 
 ```puppet
   class { 'sensu::agent':
-    backends    => ['sensu-backend.example.com:8081'],
-    config_hash => {
-      'subscriptions' => ['linux', 'apache-servers'],
-    },
+    backends      => ['sensu-backend.example.com:8081'],
+    subscriptions => ['linux', 'apache-servers'],
   }
 ```
 
@@ -192,10 +190,8 @@ This module supports Windows Sensu Go agent via chocolatey beginning with versio
 
 ```puppet
 class { 'sensu::agent':
-  backends    => ['sensu-backend.example.com:8081'],
-  config_hash => {
-    'subscriptions' => ['windows'],
-  },
+  backends      => ['sensu-backend.example.com:8081'],
+  subscriptions => ['windows'],
 }
 ```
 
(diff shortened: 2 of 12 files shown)


.fixtures.yml
```diff
@@ -5,7 +5,7 @@ fixtures:
       ref: 5.0.1
     stdlib:
       repo: git://github.com/puppetlabs/puppetlabs-stdlib.git
-      ref: 4.25.1
```
ghoneycutt: `metadata.json` would also need to be updated.

README.md
```diff
@@ -157,10 +157,8 @@ associated to `linux` and `apache-servers` subscriptions.
 
 ```puppet
   class { 'sensu::agent':
-    backends    => ['sensu-backend.example.com:8081'],
-    config_hash => {
```
ghoneycutt: Does this bring us closer to using the same options as Sensu Go or further?

manifests/agent.pp
```diff
@@ -80,9 +101,14 @@
       "${backend_protocol}://${backend}"
     }
   }
-  $default_config = {
-    'backend-url' => $backend_urls,
-  }
+  $default_config = delete_undef_values({
```
ghoneycutt: Will this need to change as new types are added? If so, this would be good to call out in the DESIGN doc or similar.

.fixtures.yml
treydock: It was. Now has `>= 5.1.0 < 7.0.0`

metadata.json
```diff
@@ -10,7 +10,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.25.1 < 7.0.0"
+      "version_requirement": ">= 5.1.0 < 7.0.0"
```
treydock: @ghoneycutt This is metadata update

README.md
treydock: Neither? This just makes "important" or more commonly changed agent configs actual parameters instead of requiring everything be in `config_hash`. Several places in docs put emphasis on things like `subscriptions` and also this makes it easier to set values that are sent to the backend entity for the agent, like `labes` and `annotations` as well as `subscriptions`.

manifests/agent.pp
treydock: No, it won't. This is not impacted by types.

README.md
ghoneycutt: If we are moving this way in the new version, perhaps we should wait until then as this might confuse people.

treydock: So save merge till v4?

ghoneycutt: Yes. Want to also ensure that as a design decision we are mimicking the data structures used by sensu-go. We started down the road of making our own and mapping them back in order to make things easy to consume and it led to a lot of drift and our need for this major version bump.

treydock: Before it was possible to define something like `config_hash` with `subscriptions` that was invalid. I don't think it's realistic to expose all config options but this exposes the important ones and adds extra validation ensuring the correct data types are used.

treydock:
Appveyor failures are hard to debug until change in #1164 is merged and then rebased into this PR.

PR #1192, opened 2019-12-12
Support 'sensu-backend init' added in Sensu Go 5.16

treydock:
# Pull Request Checklist

## Description

Execute `sensu-backend init` before executing `sensuctl configure` if `sensu-backend` is in $PATH and if none of the passwords configured in Puppet work for the admin user.

## Motivation and Context

This is necessary to support Sensu Go 5.16

### DIFF
--- a/lib/puppet/provider/sensuctl_configure/sensuctl.rb
+++ b/lib/puppet/provider/sensuctl_configure/sensuctl.rb
@@ -66,6 +66,20 @@ Puppet::Type.type(:sensuctl_configure).provide(:sensuctl, :parent => Puppet::Pro
     @property_flush[:config_namespace] = value
   end
 
+  def backend_init
+    backend = which('sensu-backend')
+    return if backend.nil?
+    return if api.auth_test(resource[:url], resource[:username], resource[:password])
+    return if api.auth_test(resource[:url], resource[:username], resource[:old_password])
+    return if api.auth_test(resource[:url], resource[:username], bootstrap_password)
+    custom_environment = {
+      'SENSU_BACKEND_CLUSTER_ADMIN_USERNAME' => resource[:username],
+      'SENSU_BACKEND_CLUSTER_ADMIN_PASSWORD' => resource[:password],
+    }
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/sensuctl_configure/sensuctl.rb
```diff
@@ -66,6 +66,16 @@ def config_namespace=(value)
     @property_flush[:config_namespace] = value
   end
 
+  def backend_init
```
ghoneycutt: is this temporary, ie is it only for 5.16 and will go away with another version that has the return code?

treydock: No, this is permanent. After some testing the return code won't help because we still have to execute `init` to get the code.

lib/puppet/provider/sensuctl_configure/sensuctl.rb
```diff
@@ -66,6 +66,16 @@ def config_namespace=(value)
...
+    backend = which('sensu-backend')
+    return if backend.nil?
+    return if api.auth_test(resource[:url], resource[:username], resource[:password])
+    return if api.auth_test(resource[:url], resource[:username], resource[:old_password])
+    return if api.auth_test(resource[:url], resource[:username], bootstrap_password)
+    cmd = [backend, 'init', '--cluster-admin-username',resource[:username],'--cluster-admin-password',resource[:password]]
+    execute(cmd, :failonfail => false)
```
treydock: Need to pass `custom_environment` hash to this execution with `SENSU_BACKEND_CLUSTER_ADMIN_USERNAME ` and `SENSU_BACKEND_CLUSTER_ADMIN_PASSWORD ` and remove the `--cluster-admin-username` and --cluster-admin-password` flags.

ghoneycutt: Sounds like this is still a WIP then.

treydock: I made this change, this is ready.

PR #1195, opened 2020-01-12
Add support for sensuctl 'command' subcommand

### LINKED ISSUE #1187: Add support for 'sensuctl commands'
This was originally part of #1174 but is delayed due to issues with the documentation for how to interact with Bonsai using `sensuctl commands`. Blocking issues:

<s>sensu/sensu-go#3420</s>
<s>sensu/sensu-go#3421</s>

treydock:
# Pull Request Checklist

## Description

Add the `sensu_command` type that provides a way to manage the `sensuctl command` subcommand.

## Related Issue

Fixes #1187


treydock:
Tests failing because the license we test with has expired

treydock:
If https://github.com/sensu/sensu-go/pull/3495 is released before this is merged it will simplify the logic in this new provider. Because there is no `--format json` I had to hack things a bit to work around that.

ghoneycutt:
Failing due to expired license.

ghoneycutt:
Please add a `# TODO` comment next to the code that will change after the sensu-go update along with a link to the PR/issue.

spec/acceptance/06_postgresql_spec.rb
```diff
@@ -16,8 +16,7 @@ class { 'postgresql::globals':
       }
       class { 'postgresql::server':}
       class { 'sensu::backend':
-        license_source => '/root/sensu_license.json',
-        datastore      => 'postgresql',
```
ghoneycutt: why is license_source no longer needed?

treydock: Because all features are available to Sensu Go without a license until you reach 100 entities. It is something that changed in a more recent version of Sensu Go.  The license is only required if you have more than 100 entities. So I removed it from these tests since these tests are testing things like Postgresql and not the license which isn't necessary.

ghoneycutt:
puppet 6 is failing in windows. Perhaps a gem needs to be pinned.

treydock:
The issue seems to be Puppet agent on Windows isn't using AIO gems or is using some gems from Appveyor installs of Ruby. I can't tell for sure but as ffi gem update maybe the issue.

Issue #1196, opened 2020-01-16
[Feature Request] - Add possibility to use hiera when adding ssl cert & keys


tdmike:
Currently, you can only add SSL certs and keys using the constraints of source. When using GIT as a source for the puppet code this would mean that private keys are uploaded to GIT without any encryption. I would like to see the use of content to allow usage of eYAML through hieradata.

Example:

https://github.com/sensu/sensu-puppet/blob/3a4d35fe5131949ca6be71de2d37071da9e85924/manifests/backend.pp#L215

With the license, you do allow to either use source or content parameter:
https://github.com/sensu/sensu-puppet/blob/3a4d35fe5131949ca6be71de2d37071da9e85924/manifests/backend.pp#L184
https://github.com/sensu/sensu-puppet/blob/3a4d35fe5131949ca6be71de2d37071da9e85924/manifests/backend.pp#L185

Allowing both parameters would allow us to add a private key to the backend in a secure way.

ghoneycutt:
Thank you for reporting this! While we work on modifying the code to better support this, you can work around it by adding those files in your profile code and then changing the sensu parameters to read the files locally. So the source would use `file:///path/to/file`

tdmike:
Hi, I already started working on modifying the code (https://github.com/sensu/sensu-puppet/pull/1197) However I have no proper way to test since I currently don't have a testing environment (I ran into this problem in our production environment)

treydock:
@tdmike I continued your pull request in #1198

ghoneycutt:
@tdmike This new functionality is in v4.2.0. You can check out the new parameter names in the commit at https://github.com/sensu/sensu-puppet/pull/1198/files

PR #1197, opened 2020-01-16
Added content parameters to allow usage of both source and content in cert and keys

tdmike:
## Description

Changes to the backend.pp
- Added content parameters for ssl_cert and ssl_key
- Added if statements to check if at least one variable is used in combination with `$use_ssl`
- Added if statements to check if not both source and content are in use

Changes to init.pp
- Added content parameters for ssl_ca
- Added if statements to check if at least one variable is used in combination with `$use_ssl`
- Added if statements to check if not both source and content are in use

## Related Issue
https://github.com/sensu/sensu-puppet/issues/1196

## Motivation and Context
These changes allow the use of either content or source parameters for the ssl_ca, ssl_cert and ssl_key. 

When using a puppet server which pulls the puppet code from a git repository, it would result in a private key to be uploaded to git to allow the puppet server access to the code. When that git server is accessible by multiple users it allows for the private key to be leaked and therefore be insecure.

## How Has This Been Tested?
I have not yet been able to test this.

### DIFF
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -128,17 +128,17 @@ class sensu::backend (
   $api_protocol = $sensu::api_protocol
   $password = $sensu::password
 
-  if $use_ssl and ! ($ssl_cert_source or $ssl_cert_content) {
-    fail('sensu::backend: ssl_cert_source or ssl_cert_content must be defined when sensu::use_ssl is true')
-  }
   if $use_ssl and $ssl_cert_source and $ssl_cert_content {
     fail('sensu::backend: Do not define both ssl_cert_source and ssl_cert_content_content')
   }
-  if $use_ssl and ! ($ssl_key_source or $ssl_key_content) {
-    fail('sensu::backend: ssl_key_source or ssl_cert_content must be defined when sensu::use_ssl is true')
+  if $use_ssl and ! ($ssl_cert_source and $ssl_cert_content) {
+    $ssl_cert_source = $facts['puppet_hostcert']
   }
   if $use_ssl and $ssl_key_source and $ssl_key_content {
     fail('sensu::backend: Do not define both ssl_key_source and ssl_key_content')
+  }
+  if $use_ssl and ! ($ssl_key_source and $ssl_key_content) {
+    $ssl_key_source = $facts['puppet_hostprivkey']
   }
 
   if $use_ssl {
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -28,8 +28,16 @@
 #   Sensu backend configuration hash used to define backend.yml.
 # @param ssl_cert_source
 #   The SSL certificate source
+#   Do not define with ssl_cert_content
+# @param ssl_cert_content
+#   The SSL certificate content
+#   Do not define with ssl_cert_source
 # @param ssl_key_source
 #   The SSL private key source
+#   Do not define with ssl_key_content
+# @param ssl_key_content
+#   The SSL private key content
+#   Do not define with ssl_key_content
 # @param include_default_resources
 #   Sets if default sensu resources should be included
 # @param show_diff
@@ -77,8 +85,10 @@ class sensu::backend (
   Boolean $service_enable = true,
   Stdlib::Absolutepath $state_dir = '/var/lib/sensu/sensu-backend',
   Hash $config_hash = {},
-  Optional[String] $ssl_cert_source = $facts['puppet_hostcert'],
-  Optional[String] $ssl_key_source = $facts['puppet_hostprivkey'],
+  Optional[String] $ssl_cert_source = undef,
+  Optional[String] $ssl_cert_content = undef,
+  Optional[String] $ssl_key_source = undef,
+  Optional[String] $ssl_key_content = undef,
   Boolean $include_default_resources = true,
   Boolean $show_diff = true,
   Optional[String] $license_source = undef,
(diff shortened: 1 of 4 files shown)


manifests/backend.pp
```diff
@@ -77,8 +85,10 @@
   Boolean $service_enable = true,
   Stdlib::Absolutepath $state_dir = '/var/lib/sensu/sensu-backend',
   Hash $config_hash = {},
-  Optional[String] $ssl_cert_source = $facts['puppet_hostcert'],
-  Optional[String] $ssl_key_source = $facts['puppet_hostprivkey'],
```
treydock: Changing this default would be a breaking change. It's possible to make `undef` if the manifest later checks $ssl_*_content and sets $ssl_*_source to the previous defaults if content is undefined.

treydock:
Please remove the changes to REFERENCE.md. We make that change as part of the release process.

manifests/backend.pp
tdmike: I applied a fix, now when both parameters are empty the source parameters get a default variable

manifests/init.pp
```diff
@@ -75,8 +80,11 @@
   Optional[String] $agent_old_password = undef,
 ) {
 
-  if $use_ssl and ! $ssl_ca_source {
-    fail('sensu: ssl_ca_source must be defined when use_ssl is true')
+  if $use_ssl and $ssl_ca_source and $ssl_ca_content {
+    fail('sensu::backend: Do not define both ssl_ca_source and ssl_ca_content_content')
+  }
+  if $use_ssl and ! ($ssl_ca_source or $ssl_ca_content) {
+    $ssl_ca_source = $facts['puppet_localcacert']
```
treydock: Can't reassign a variable in Puppet. This would have to be named `$_ssl_ca_source` and that value used.

Simpler approach:

```
if $ssl_ca_content {
  $_ssl_ca_source = undef
} else {
  $_ssl_ca_source = pick($ssl_ca_source, $facts['puppet_localcacert'])
}
```

Will need to have two validation conditions

```
if $use_ssl {
  if $ssl_ca_source and $ssl_ca_content {
    ... error they can't be defined at same time ...
  }
  if ! ($ssl_ca_source and $ssl_ca_content) {
    ... ssl_ca_source or ssl_ca_content must be defined when use_ssl is true ...
  }
}
```

manifests/backend.pp
```diff
@@ -118,11 +128,17 @@
...
-  if $use_ssl and ! $ssl_key_source {
-    fail('sensu::backend: ssl_key_source must be defined when sensu::use_ssl is true')
+  if $use_ssl and ! ($ssl_cert_source and $ssl_cert_content) {
+    $ssl_cert_source = $facts['puppet_hostcert']
+  }
+  if $use_ssl and $ssl_key_source and $ssl_key_content {
+    fail('sensu::backend: Do not define both ssl_key_source and ssl_key_content')
+  }
+  if $use_ssl and ! ($ssl_key_source and $ssl_key_content) {
+    $ssl_key_source = $facts['puppet_hostprivkey']
```
treydock: Similar to my other commands, you can't redfine the same variable twice in Puppet.  If `$ssl_key_content` is undefined then use `pick`, else make the `$_ssl_key_source` set to undefined.

manifests/backend.pp
```diff
@@ -118,11 +128,17 @@
...
 
-  if $use_ssl and ! $ssl_cert_source {
-    fail('sensu::backend: ssl_cert_source must be defined when sensu::use_ssl is true')
+  if $use_ssl and $ssl_cert_source and $ssl_cert_content {
+    fail('sensu::backend: Do not define both ssl_cert_source and ssl_cert_content_content')
   }
-  if $use_ssl and ! $ssl_key_source {
-    fail('sensu::backend: ssl_key_source must be defined when sensu::use_ssl is true')
+  if $use_ssl and ! ($ssl_cert_source and $ssl_cert_content) {
+    $ssl_cert_source = $facts['puppet_hostcert']
```
treydock: Similar to my other commands, you can't redfine the same variable twice in Puppet. If $ssl_cert_content is undefined then use pick, else make the $_ssl_cert_source set to undefined.

manifests/backend.pp
```diff
@@ -118,11 +128,17 @@
   $api_protocol = $sensu::api_protocol
   $password = $sensu::password
 
-  if $use_ssl and ! $ssl_cert_source {
-    fail('sensu::backend: ssl_cert_source must be defined when sensu::use_ssl is true')
+  if $use_ssl and $ssl_cert_source and $ssl_cert_content {
```
treydock:
```
if $use_ssl {
  if $ssl_cert_source and $ssl_cert_content {
    ... error they can't be defined at same time ...
  }
  if ! ($ssl_cert_source and $ssl_cert_content) {
    ... ssl_cert_source or ssl_cert_content must be defined when use_ssl is true ...
  }
  if $ssl_key_source and $ssl_key_content {
    ... error they can't be defined at same time ...
  }
  if ! ($ssl_key_source and $ssl_key_content) {
    ... ssl_key_source or ssl_key_content must be defined when use_ssl is true ...
  }
}
```

ghoneycutt:
After reviewing the code, I'm not keen to add more configuration options to specify secrets. The way it currently works, you pass in paths to files that are already on disk and then the module copies them to a sensu specific ssl directory. That way you can use existing certs as we do, or handle them however you would like such as file resources with content from eyaml or managed by packages. Going to close this PR and we will instead better document and show examples of how to use the module for these use cases.

PR #1198, opened 2020-01-19
Allow SSL files to be defined via content parameters

### LINKED ISSUE #1196: [Feature Request] - Add possibility to use hiera when adding ssl cert & keys

treydock:
# Pull Request Checklist

## Description

Allow SSL CA cert, cert and key to be defined via content.

## Related Issue

Fixes #1196 and replaces #1197 

## Motivation and Context

This allows one to define SSL file contents via Hiera using EYAML.

### DIFF
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -28,16 +28,16 @@
 #   Sensu backend configuration hash used to define backend.yml.
 # @param ssl_cert_source
 #   The SSL certificate source
-#   Do not define with ssl_cert_content
+#   This parameter is mutually exclusive with ssl_cert_content
 # @param ssl_cert_content
 #   The SSL certificate content
-#   Do not define with ssl_cert_source
+#   This parameter is mutually exclusive with ssl_cert_source
 # @param ssl_key_source
 #   The SSL private key source
-#   Do not define with ssl_key_content
+#   This parameter is mutually exclusive with ssl_key_content
 # @param ssl_key_content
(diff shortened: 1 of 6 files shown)


manifests/backend.pp
```diff
@@ -28,8 +28,16 @@
 #   Sensu backend configuration hash used to define backend.yml.
 # @param ssl_cert_source
 #   The SSL certificate source
```
ghoneycutt: Under both `_content` and `_source` params put the note 'This parameter is mutually exclusive with X'

treydock: Changes made

PR #1199, opened 2020-01-19
Add examples for Slack and InfluxDB

treydock:
# Pull Request Checklist

## Description

Adds examples for Sensu Go guides such, for Slack and InfluxDB. Also moved contact routing example into examples directory.

### DIFF
--- a/examples/contact_routing.pp
+++ b/examples/contact_routing.pp
@@ -1,0 +1,59 @@
+include sensu::backend
+
+sensu_bonsai_asset { 'sensu/sensu-go-has-contact-filter':
+  ensure  => 'present',
+  version => 'latest',
+}
+
+sensu_filter { 'contact_dev':
+  ensure         => 'present',
+  action         => 'allow',
+  runtime_assets => ['sensu/sensu-go-has-contact-filter'],
+  expressions    => ['has_contact(event, "dev")'],
+}
+sensu_filter { 'contact_ops':
+  ensure         => 'present',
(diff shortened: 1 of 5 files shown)


examples/contact_routing.pp
```diff
@@ -0,0 +1,59 @@
...
+sensu_handler { 'email':
+  ensure   => 'present',
+  type     => 'set',
+  handlers => ['email_dev','email_ops'],
+}
+
+sensu_check { 'check_cpu':
+  ensure         => 'present',
+  labels         => {
+    'contacts' => 'dev, ops',
```
ghoneycutt: This is a string? I would think it would be an array.

spec/acceptance/examples_spec.rb
```diff
@@ -2,65 +2,10 @@
...
-  command        => 'check-cpu.rb -w 75 -c 90',
-  handlers       => ['email'],
-  interval       => 30,
-  publish        => true,
-  subscriptions  => ['linux'],
-  runtime_assets => ['sensu-plugins-cpu-checks','sensu-ruby-runtime'],
-}
-      EOS
 
+  RSpec.configuration.sensu_examples.each do |example|
```
ghoneycutt: This is awesome!

examples/contact_routing.pp
treydock: Yes, label values are all strings. The handler that looks for `contacts` does the necessary splitting.

PR #1201, opened 2020-01-25
Better support for Sensu Go upgrades

treydock:
# Pull Request Checklist

## Description

Ensure that upgrading packages will notify the necessary services and for systemd ensure to run `systemctl daemon-reload` per Sensu Go docs.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -173,7 +173,9 @@ class sensu::agent (
   }
 
   # See https://docs.sensu.io/sensu-go/latest/installation/upgrade/
-  if $facts['service_provider'] == 'systemd' {
+  # Only necessary for Puppet < 6.1.0,
+  # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
+  if versioncmp($facts['puppetversion'],'6.1.0') < 0 and $facts['service_provider'] == 'systemd' {
     Package['sensu-go-agent'] ~> Exec['sensu systemctl daemon-reload']
     Exec['sensu systemctl daemon-reload'] -> Service['sensu-agent']
   }
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -172,6 +172,12 @@ class sensu::agent (
     $_package_source = undef
   }
 
+  # See https://docs.sensu.io/sensu-go/latest/installation/upgrade/
+  if $facts['service_provider'] == 'systemd' {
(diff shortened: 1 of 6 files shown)


treydock:
Travis failures will be resolved once #1202 is merged and this is rebased

manifests/agent.pp
```diff
@@ -172,13 +172,19 @@
     $_package_source = undef
   }
 
+  if $facts['service_provider'] == 'systemd' {
```
ghoneycutt: This seems very clunky. If the docs say this is what should be done, let's put a link to the docs here and a short explanation.

treydock: I'll add the link but this is a very common pattern I've encountered. Normally this would be handled by a module like camptocamp/systemd if we defined the unit files in Puppet or by the RPM installing the unit files. 

FWIW Puppet >= 6.1.0 will auto execute `systemctl daemon-reload` before restarting a service.  I am updating PR to only execute the command for Pupppet < 6.1.0 which is a pattern in some voxpupuli modules dealing with systemd.

Puppet change: https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4

manifests/agent.pp
```diff
@@ -173,7 +173,9 @@
   }
 
   # See https://docs.sensu.io/sensu-go/latest/installation/upgrade/
-  if $facts['service_provider'] == 'systemd' {
+  # Only necessary for Puppet < 6.1.0,
```
ghoneycutt: We only support the latest versions of puppet 5 and 6. Given this is an old version, I don't think we should support it. We should document this though.

treydock: This is necessary to ensure we execute `systemctl daemon-reload` on Puppet 5. Only Puppet >= 6.1.0 has this fix. The latest Puppet 5 does NOT execute `systemctl daemon-reload`. The condition I added ensures we are not executing this for latest Puppet 6 versions where it's unnecessary.

spec/classes/backend_spec.rb
```diff
@@ -21,6 +21,14 @@
         it { should contain_class('sensu::backend::default_resources') }
         it { should_not contain_class('sensu::backend::datastore::postgresql') }
 
+        context 'on systemd host', if: Puppet.version.to_s =~ %r{^5} do
```
ghoneycutt: shouldn't this also match 6.0

treydock: No, because we don't rspec unit test against older version of Puppet 5. We unit test against latest 5.x and latest 6.x.

PR #1202, opened 2020-01-25
Several fixes for sensu_bonsai_asset

treydock:
# Pull Request Checklist

## Description

Fix sensu_bonsai_asset so that when multiple bonsai assets with different latest version are defined, don't use latest version for first resource queried from bonsai. Previous code attempted to cache the wrong value from bonsai API query.

Also fix issue where deleting a bonsai asset would generate an `absent` resource for `sensu_asset` that would cause an attempt to delete non-existent asset which would cause errors.

These issues were not seen before because the assets being tested with multiple `latest` values just so happened to have same version in Bonsai.

### DIFF
--- a/spec/acceptance/sensu_bonsai_asset.rb
+++ b/spec/acceptance/sensu_bonsai_asset.rb
@@ -49,9 +49,16 @@ describe 'sensu_bonsai_asset', if: RSpec.configuration.sensu_full do
 
   context 'install bonsai asset - latest' do
     it 'should work without errors' do
+      # Define two assets as 'latest' using different providers
+      # to ensure the latest of each is used
+      # See https://github.com/sensu/sensu-puppet/pull/1202
       pp = <<-EOS
       include sensu::backend
       sensu_bonsai_asset { 'sensu/sensu-pagerduty-handler':
+        ensure  => 'present',
+        version => 'latest',
+      }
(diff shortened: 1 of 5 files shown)


spec/acceptance/sensu_bonsai_asset.rb
```diff
@@ -60,6 +60,10 @@
         version  => 'latest',
         provider => 'sensu_api',
       }
+      sensu_bonsai_asset { 'sensu/sensu-go-has-contact-filter':
```
ghoneycutt: How does this test show us anything?

treydock: The condition of failure was having multiple bonsai assets defined having different latest versions in Bonsai, using the same provider. The previous tests had one defined via `sensuctl` provider and one via `sensu_api` provider.  Having two resources using `sensuctl` would use the latest version from one and incorrectly assume it was the latest version for other resources of same provider.  This is a more direct test that was detected in contact routing example with same resources.

ghoneycutt: Could you please put a comment in the code to explain this and link to this issue.

treydock: Comment added.

PR #1203, opened 2020-01-25
Support Sensu Go secrets features

treydock:
# Pull Request Checklist

## Description

* Add sensu_secrets_vaults_provider type/providers
* Add sensu_secret type/providers
* Add secrets property to sensu_check, sensu_handler and sensu_mutator

## Motivation and Context

Support for secrets will be added in 5.17. This adds support for those changes.


treydock:
Tests will fail until Sensu Go 5.17 is released, the CI tests should pass though.

lib/puppet/type/sensu_secret.rb
```diff
@@ -0,0 +1,93 @@
+require_relative '../../puppet_x/sensu/type'
+require_relative '../../puppet_x/sensu/array_property'
+require_relative '../../puppet_x/sensu/array_of_hashes_property'
+require_relative '../../puppet_x/sensu/hash_property'
+require_relative '../../puppet_x/sensu/integer_property'
+
+Puppet::Type.newtype(:sensu_secret) do
+  desc <<-DESC
+@summary Manages Sensu Secrets
+@example Add a secret to default namespace
```
ghoneycutt: `Manage a secret in` as opposed to `Add a secret to`, is a way to describe what's happening in an idempotent fashion.

lib/puppet/type/sensu_secrets_provider.rb
```diff
@@ -0,0 +1,170 @@
+require_relative '../../puppet_x/sensu/type'
+require_relative '../../puppet_x/sensu/array_property'
+require_relative '../../puppet_x/sensu/array_of_hashes_property'
+require_relative '../../puppet_x/sensu/hash_property'
+require_relative '../../puppet_x/sensu/integer_property'
+
+Puppet::Type.newtype(:sensu_secrets_provider) do
+  desc <<-DESC
+@summary Manages Sensu Secrets provider
+@example Add a secrets vault provider
```
ghoneycutt: same as above

lib/puppet/type/sensu_secret.rb
treydock: Wording updated

lib/puppet/type/sensu_secrets_provider.rb
treydock: Wording updated

treydock:
Marking WIP. One thing I don't like is `sensu_secrets_provider` is entirely configured via the `client` hash which is complicated to validate and when logged it's nearly impossible to tell what changed.

I think I'm going to treat the `VaultProvider` type of secrets provider like we do AD, LDAP and OIDC auths, and make it a distinct type. This would mean to manage a vault provider you'd have `sensu_secrets_vault_provider` and instead of a big hash to configure it you'd have several properties that build the hash. We'd just have to document that the properties are keys in the `client` hash to make it clear how to translate from Sensu Go docs to Puppet docs.

This change would make it easier to support  future secrets providers if upstream adds more than the Vault and Env providers.

treydock:
Renamed `sensu_secrets_provider` to `sensu_secrets_vault_provider` which is a much cleaner implementation. This also ensures we are able to redact the `token` value as that's a sensitive value used for authentication.

PR #1206, opened 2020-01-29
Add manage_agent_user parameter to sensu::backend

### LINKED ISSUE #1205: [Feature Request] Add ability to remove default agent user and groups
## Description of problem
As we are running the Enterprise variant of Sensu, we are setting up TLS Authentication. In the process of setting this up, I removed the agent user and system:agents group as described in the Sensu Docs. However, the next puppet run put it back in. 

I would like to have the ability to disable the configuring of the default user since it's impossible to remove while using puppet.

treydock:
# Pull Request Checklist

## Description

Allow disabling management of `agent` Sensu user. Also allow agent user to be disabled.

## Related Issue

Fixes #1205 

## Motivation and Context

Sensu Go docs describe agent authentication via TLS rather than a password and this facilitates that. Need to either stop managing the agent user or disable the user. This pull requests adds support for both.

### DIFF
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -40,6 +40,13 @@
 #   This parameter is mutually exclusive with ssl_key_source
 # @param include_default_resources
 #   Sets if default sensu resources should be included
+# @param manage_agent_user
+#   Sets if the Sensu agent user should be managed
+# @param agent_user_disabled
+#   Sets if the Sensu agent user should be disabled
+#   Not applicable if `manage_agent_user` is `false`
+#   This is useful if using agent TLS authentication
+#   See https://docs.sensu.io/sensu-go/latest/guides/securing-sensu/#sensu-agent-tls-authentication
 # @param show_diff
 #   Sets show_diff parameter for backend.yml configuration file
 # @param license_source
@@ -90,6 +97,8 @@ class sensu::backend (
   Optional[String] $ssl_key_source = $facts['puppet_hostprivkey'],
   Optional[String] $ssl_key_content = undef,
   Boolean $include_default_resources = true,
+  Boolean $manage_agent_user = true,
+  Boolean $agent_user_disabled = false,
   Boolean $show_diff = true,
   Optional[String] $license_source = undef,
   Optional[String] $license_content = undef,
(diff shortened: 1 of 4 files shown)


treydock:
@tdmike See if this works for your requirement in #1205. If you want to disable the agent user you can set `agent_user_disabled` to `true` for `sensu::backend`. You can completely unmanage with `manage_agent_user`.

manifests/backend.pp
```diff
@@ -40,6 +40,10 @@
 #   This parameter is mutually exclusive with ssl_key_source
 # @param include_default_resources
 #   Sets if default sensu resources should be included
+# @param manage_agent_user
+#   Sets if the Sensu agent user should be managed
+# @param agent_user_disabled
+#   Sets if the Sensu agent user should be disabled
```
ghoneycutt: Add something about not being applicable if `manage_agent_user` is `false`.

manifests/backend.pp
```diff
@@ -40,6 +40,10 @@
 #   This parameter is mutually exclusive with ssl_key_source
 # @param include_default_resources
 #   Sets if default sensu resources should be included
+# @param manage_agent_user
```
ghoneycutt: Link the docs where Sensu says you should do this. Curious if this is a common use case.

treydock: I think more appropriate to not use `manage_agent_user` and just disable the agent, so will update the docs for that parameter.

manifests/backend.pp
treydock: COmment updated and added link to Sensu Go docs for agent TLS which is when it would be most useful to disable this unnecessary user.

ghoneycutt:
The docs read to remove the user whereas this code will simply stop managing it and the user would still exist. Do we need to support the ability to `ensure => absent` ?

ghoneycutt:
If disabling the user works, let's open a PR against the sensu docs to reflect disabling instead of removing.

treydock:
It's not yet possible to delete users in Sensu Go, you can only disable them. The agent user is added as part of fresh install but can be disabled, which is what this pull request does.

tdmike:
Like @treydock says, you can't remove the user in Sensu Go, only disable it. And that works fine. I am setting up a testing environment as I type this. I will let you know if all works as expected.

PR #1215, opened 2020-02-10
Several fixes for sensu_bonsai_asset

### LINKED ISSUE #1210: Duplicate Bonsai Assets when using multiple namespaces
## Description of problem
When defining the same Bonsai Assets for multiple namespaces you get a duplicate declaration
- What did you do?
Created two namespaces to enable separation between environments, tried to add the same check to both namespaces
- What happened?
Received the following error:
```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Sensu_bonsai_asset[sensu-plugins/sensu-plugins-cpu-checks] is already declared at (file: /etc/puppetlabs/code/environments/development/site-modules/profile/manifests/sensu_manage.pp, line: 30); cannot redeclare (file: /etc/puppetlabs/code/environments/development/site-modules/profile/manifests/sensu_manage.pp, line: 35) (file: /etc/puppetlabs/code/environments/development/site-modules/profile/manifests/sensu_manage.pp, line: 35, column: 3) on node sensu01.dev.mgt.osp.its.uu.nl
```
I expected this to happen, due to the usage of the asset name in de declaration.
Create two namespaces
Add same check to both namespaces

This is a master bases puppet setup

- Your OS: Red Hat Enterprise Linux 7.
- Your Ruby version:
- Your version of Puppet: Puppet Enterprise 5.5.10
- Your version of Sensu: 5.16.1
- Your version of this module: 4.4.0

treydock:
# Pull Request Checklist

## Description

Several fixes for `sensu_bonsai_asset` when using composite names and the same asset in multiple namespaces. Previous behavior was broken for default value of `rename` property. New default matches what `sensuctl` does when you do `sensuctl asset add <bonsai asset>`. This is not breaking because past behavior for a name like `sensu/sensu-pagerduty-handler` will still behave the same while the composite name where change takes place would have never actually worked and would have raised errors if used before this fix.

## Related Issue

Fixes #1210

### DIFF
--- a/spec/acceptance/sensu_bonsai_asset.rb
+++ b/spec/acceptance/sensu_bonsai_asset.rb
@@ -15,6 +15,10 @@ describe 'sensu_bonsai_asset', if: RSpec.configuration.sensu_full do
         version  => '0.1.0',
         provider => 'sensu_api',
       }
+      sensu_bonsai_asset { 'sensu/sensu-ruby-runtime in default':
+        ensure  => 'present',
+        version => 'latest',
+      }
       EOS
 
       if RSpec.configuration.sensu_use_agent
@@ -43,6 +47,14 @@ describe 'sensu_bonsai_asset', if: RSpec.configuration.sensu_full do
         data = JSON.parse(stdout)
         version = data['metadata']['annotations']['io.sensu.bonsai.version']
         expect(version).to eq('0.1.0')
+      end
+    end
+
+    it 'should have bonsai asset' do
+      on node, 'sensuctl asset info sensu/sensu-ruby-runtime --format json' do
+        data = JSON.parse(stdout)
+        name = data['metadata']['name']
+        expect(name).to eq('sensu/sensu-ruby-runtime')
       end
     end
   end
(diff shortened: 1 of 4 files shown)


spec/acceptance/sensu_bonsai_asset.rb
```diff
@@ -15,6 +15,15 @@
         version  => '0.1.0',
         provider => 'sensu_api',
       }
+      sensu_bonsai_asset { 'sensu/sensu-ruby-runtime in default':
+        ensure  => 'present',
+        version => 'latest',
+      }
+      sensu_namespace { 'dev': ensure => 'present' }
+      sensu_bonsai_asset { 'sensu/sensu-ruby-runtime in dev':
```
ghoneycutt: does this resource type know to auto require the dev namespace above?

treydock: Yes.  We have the tests setup for all things with `namespace` property.

Auto requires:

https://github.com/sensu/sensu-puppet/blob/master/lib/puppet/type/sensu_bonsai_asset.rb#L40

Leads to:

https://github.com/sensu/sensu-puppet/blob/master/lib/puppet_x/sensu/type.rb#L29

Spec tests are shared for all things that use a namespace 

https://github.com/sensu/sensu-puppet/blob/master/spec/unit/sensu_bonsai_asset_spec.rb#L232

Leads to:

https://github.com/sensu/sensu-puppet/blob/master/spec/shared_examples/types.rb#L79

PR #1218, opened 2020-02-18
Manage license through sensu_license type

### LINKED ISSUE #1217: Feature Request - Check if license is installed
Hi Guys,

I ran into an issue where my license failed to install when we deployed the sensu cluster. Would it be possible to implement a check when the `license_source` or `license_content` is defined puppet checks if it's installed and if it isn't puppet installs it?

treydock:
# Pull Request Checklist

## Description

Manage licenses through `sensu_license` type.

## Related Issue

Fixes #1217 

## Motivation and Context

Previously a failed license add would not retry if license wasn't added and only way to force Puppet to re-add failed Exec was to modify or delete the license file and let it get re-created and trigger the Exec. This ensures some better logic around license management and ensures changes to the license file still refresh the license inside `sensuctl`.

### DIFF
--- a/spec/acceptance/05_enterprise_spec.rb
+++ b/spec/acceptance/05_enterprise_spec.rb
@@ -32,4 +32,55 @@ describe 'sensu::backend class', unless: RSpec.configuration.sensu_cluster do
       its(:exit_status) { should eq 0 }
     end
   end
+  context 'updates license file' do
+    it 'should work without errors and be idempotent' do
+      pp = <<-EOS
+      class { 'sensu::backend':
+        license_source => '/root/sensu_license.json',
+      }
+      EOS
+
+      # Remove license file to ensure refresh works
+      on node, puppet("resource file /etc/sensu/license.json ensure=absent")
+      if RSpec.configuration.sensu_use_agent
+        site_pp = "node 'sensu_backend' { #{pp} }"
+        puppetserver = hosts_as('puppetserver')[0]
+        create_remote_file(puppetserver, "/etc/puppetlabs/code/environments/production/manifests/site.pp", site_pp)
+        on node, puppet("agent -t --detailed-exitcodes"), acceptable_exit_codes: [2]
+      else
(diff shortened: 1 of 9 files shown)


spec/acceptance/05_enterprise_spec.rb
```diff
@@ -28,6 +28,57 @@ class { 'sensu::backend':
...
+  context 're-adds license file' do
+    it 'should work without errors and be idempotent' do
+      pp = <<-EOS
+      class { 'sensu::backend':
+        license_source => '/root/sensu_license.json',
+      }
+      EOS
+
+      # Remove license to verify it can re-add
+      on node, "sensuctl delete -f /etc/sensu/license.json"
```
ghoneycutt: Use the new type's ensure => absent instead of calling the sensuctl delete directly, so we have a test for that functionality.

spec/unit/sensu_license.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  include_examples 'autorequires', false, false do
+    let(:res) { configure }
+  end
+
+  [
+    :file,
+  ].each do |property|
+    it "should require property when ensure => present" do
```
ghoneycutt: Needs a test when ensure is absent

ghoneycutt:
Great work!

spec/acceptance/05_enterprise_spec.rb
treydock: The type supports ensure=absent but the type is only used through `sensu::backend` and that class doesn't support setting the license to absent. I updated the acceptance test to use `ensure=absent` but it's not intended to be used.  I didn't want to add logic to the backend class that might remove someone's license which could do nasty things if they have more than 100 nodes.

spec/unit/sensu_license.rb
treydock: Test added. Also changed logic so `file` is required regardless of `ensure` because it's used by both present and absent.

PR #1219, opened 2020-02-18
Fixes for pdk validate

treydock:
# Pull Request Checklist

## Description

Issues not detected by `rake lint` which I can't figure out why that's the case, but these issues do get picked up by `pdk validate`.


ghoneycutt:
Closing in favor of #1220

PR #1225, opened 2020-02-26
Fix to tests to work with Bolt 2

treydock:
# Pull Request Checklist

## Description

This fixes tests to work with Bolt v2 that removed `--nodes` flag and replaced with `--targets`. There are no changes necessary to the bolt tasks in this module, just how tests were executed.


treydock:
The appveyor failure is the transient issue we sometimes see where a file we copy into environment is missing for some reason.

PR #1226, opened 2020-02-29
Fix issue where absent HOME would break Dir.home

### LINKED ISSUE #1072: Could not evaluate: couldn't find login name -- expanding `~'

treydock:
# Pull Request Checklist

## Description

Fix regression due to how `Dir.home` does the home directory lookup.  Seems `Dir.home` will try to expand `~` which won't work if `HOME` env var is not set.  This change will catch that exception as well as if `Dir.home` is not supported and fall back to home directory lookup via `Etc` module.

I tried to add acceptance tests for this but all attempts to unset `HOME` but I was able to reproduce the issue interactively inside Docker.

```
[root@sensu_agent /]# unset HOME
[root@sensu_agent /]# /opt/puppetlabs/puppet/bin/ruby -e 'print Dir.home'
-e:1:in `home': couldn't find login name -- expanding `~' (ArgumentError)
        from -e:1:in `<main>'
```

Before this patch:

```
[root@sensu_agent /]# unset HOME ; puppet apply -e "include sensu::cli"
Notice: Compiled catalog for sensu_agent in environment production in 0.14 seconds
Error: /Stage[main]/Sensu::Cli/Sensuctl_configure[puppet]: Could not evaluate: couldn't find login name -- expanding `~'
```

After this patch:

```
[root@sensu_agent /]# puppet apply -e 'include sensu::cli'
Notice: Compiled catalog for sensu_agent in environment production in 0.14 seconds
Notice: Applied catalog in 0.85 seconds
```

Attempts to do the `unset HOME` in acceptance tests failed as no matter how I tried it seemed that the issue would not occur, something with how the acceptance tests are executed would cause `unset HOME` to have no effect.

## Related Issue

Fixes #1072

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 ---
 language: ruby
 
-bundler_args: --without system_tests development
-
 cache: bundler
 
 before_install:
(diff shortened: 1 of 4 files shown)


treydock:
Tests will fail until #1225 is merged and this is rebased.

PR #1227, opened 2020-03-02
Allow agents to have subscriptions defined as a resource

### LINKED ISSUE #1222: Lost the ability to separately define subscriptions
## Description of problem

I'm looking to migrate our Sensu install from the old core/1.x way to the new Go codebase. Aside from all of the backend changes I need to make I've realised there's now no way to separately define agent subscriptions.

Previously in the old way, I could just add the following to any class:
```puppet
sensu::subscription { 'blah':
  ensure => present,
}
```
Whereas now, that defined type has disappeared and it seems I need to know all of the subscriptions that will be required/applied to an agent ahead of time to pass them once to the `sensu::agent` class.

- Your OS: RHEL 7
- Your Ruby version:
- Your version of Puppet: 5.5.18
- Your version of Sensu: 5.17.2
- Your version of this module: master

This is partly the problem of the underlying Sensu Go codebase as it looks like they've removed support for configuration merging, however this could be maybe solved by using a concat type resource to build the agent configuration; a `sensu::subscription` define could then create a concat fragment.

treydock:
# Pull Request Checklist

## Description

Allow subscriptions for agents to be defined in such a way that works like old Sensu module and plays well with defining subscriptions in profile classes for various resources.

The content of `agent.yml` is generated from fragments at runtime after catalog is compiled, unit tests can only test fragment data but acceptance tests are able to verify the final agent.yml content.

## Related Issue

Fixes #1222 and #1183 

## Motivation and Context

This makes it easier to dynamically generate `agent.yml` based on profiles assigned to a given puppet host.  Instead of having to hardcode subscriptions, subscriptions can be built based on profile classes assigned to a given host.


manifests/agent/subscription.pp
```diff
@@ -0,0 +1,22 @@
...
+# @example
+#   sensu::agent::subscription { 'mysql': }
+#
+# @param subscription
+#   Name of the subscription to add to agent.yml, defaults to `$name`.
+# @param order
+#   Order of the datacat fragment
+#
+define sensu::agent::subscription (
+  String $subscription = $name,
```
ghoneycutt: Should this be `String[1]` ?

manifests/agent/subscription.pp
```diff
@@ -0,0 +1,22 @@
...
+  String $order        = '50',
```
ghoneycutt: Should this also be `String[1]` or perhaps `Integer`?

manifests/agent/subscription.pp
treydock: I made that change

manifests/agent/subscription.pp
treydock: I used `String[1]` because doesn't need to be Integer and all docs for datacat use a string and the default in datacat is the string `'50'`.

treydock:
Had to add updates to Ubuntu and Debian docker builds because bootstrapping Puppet was failing due to lack of HTTPS support in apt.  Normally apt module would handle this but needed these packages in order to bootstrap Puppet.

bodgit:
Great, I'll check it out. Good ol' datacat module :wink:

PR #1228, opened 2020-03-11
Ignore builtin sensu.io/managed_by label

treydock:
# Pull Request Checklist

## Description

When checking if `labels` are in sync, ignore `sensu.io/managed_by` label as this is a built-in label provided by sensuctl.  Without this change then Puppet will continuously want to update the value for labels to remove the builtin label.

## Motivation and Context

This is to address a change that will likely be released with Sensu Go 5.19. The behavior of this module won't have any effect on versions prior to 5.19 as long as someone wasn't using `sensu.io/managed_by` labels which if they are they will get overwritten by Sensu Go 5.19.


ghoneycutt:
Hi @treydock seems tests are failing

treydock:
I ran the failing tests on master branch and they are also failing so I'm guessing Puppet broke. I notice the last time these tests passed it was with Puppet 6.13.0 but now there is 6.14.0.  I am investigating what changed, as it wasn't this module that broke.

treydock:
I found the problem. We use `sensu_backend` as the Puppet server name and underscores are not valid for DNS names and a change in Puppet 6.14.0 now gets upset with invalid DNS names for a server.  I will open a new PR for the fix as it's rather involved.

ghoneycutt:
I rebased against master with your fix.

treydock:
@ghoneycutt  This is ready to merge

PR #1232, opened 2020-03-19
Avoid facter 4, breaks unit tests

treydock:
# Pull Request Checklist

## Description

Seems puppetlabs pushed a facter 4.0.11 rubygem that breaks unit tests. No supported version of puppet-agent uses facter 4 and we use facterdb for unit test facts. This avoids facter 4.


treydock:
Added commit to fix some strings formatting issues.  Seems newline is stripped if last item in line is backtick.

PR #1233, opened 2020-03-26
Use the lower bound version (5.1.0) of stdlib for Vagrant

ghoneycutt:
# Pull Request Checklist

Vagrant environment is having issues bringing up default sensu-backend.

## Motivation and Context

Fix vagrant environment


ghoneycutt:
Close in favor of #1234 which contains this commit

PR #1243, opened 2020-04-15
Change how it's determined when to run specific acceptance tests

treydock:
# Pull Request Checklist

## Description

Tests are run based on `BEAKER_sensu_mode` environment variable which defaults to `base`

This doesn't change when or how things are tested, just splits tests into smaller groups so things like examples will run separate than custom types.  The core centos7 tests will run like all other OSes but also include postgresql tests and bolt which were also only done with centos7 before. It's still only centos7 that is used to test clusters and custom types.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,52 +64,77 @@ matrix:
     stage: acceptance
   - rvm: 2.4.9
     services: docker
-    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_sensu_full=yes
+    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet5 BEAKER_sensu_mode=full
     script: bundle exec rake beaker
     stage: acceptance
   - rvm: 2.5.7
     services: docker
-    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_full=yes
+    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_mode=full
     script: bundle exec rake beaker
     stage: acceptance
   - rvm: 2.5.7
     services: docker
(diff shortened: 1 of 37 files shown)


DESIGN.md
```diff
@@ -82,6 +82,15 @@ The class tests are ordered with numeric prefixes to control the order they run.
 
 The type of `sensu_check` will have its tests in `sensu_check_spec.rb`. The tests to run are adding resources, updating resources, and deleting resources. Some extra test cases should be added based on any complexities of a given type.
 
+The resources `sensu_cluster_role`, `sensu_cluster_role_binding`, `sensu_role`, and `sensu_role_binding` are grouped into `sensu_rbac_resources_spec.rb` with the goal of speeding up testing times.
+
+By default only tests for class resources run which is the same as setting the enviornment variable `BEAKER_sensu_mode=base`.  The other possible modes are the following:
```
ghoneycutt, line 87: I dont see this in .travis.yml, should I?

.travis.yml
```diff
@@ -64,52 +64,77 @@ matrix:
...
     stage: acceptance
   - rvm: 2.5.7
     services: docker
-    env: BEAKER_sensu_ci_build=yes BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_full=yes
+    env: BEAKER_sensu_ci_build=yes BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_mode=full
+    script: bundle exec rake beaker
+    stage: acceptance
+  - rvm: 2.5.7
+    services: docker
+    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_mode=examples
```
ghoneycutt, line 82: Think we're missing puppet5 for examples

DESIGN.md
treydock: The default of `base` is set in `spec/spec_helper_acceptance.rb` where the setting is defined. If nothing is defined it defaults to `base`.

.travis.yml
treydock: Updated to have puppet5 tested too.

PR #1248, opened 2020-05-01
Use Ruby 2.4.10 for testing Puppet 5

ghoneycutt:
Puppet has released a new version of Puppet 5 that uses Ruby 2.4.10.


ghoneycutt:
Puppet's release statements do not match their documentation (https://puppet.com/docs/puppet/5.5/about_agent.html). Going to keep in line with their docs.

PR #1251, opened 2020-05-23
READ DESCRIPTION: Improved handling of passwords for sensu_user

treydock:
# Pull Request Checklist

## Description

* Deprecate `sensu::old_password` and `sensu::agent_old_password`, no longer used or necessary
* Validate `sensu_user` passwords are at least 8 characters long
* All resources except now auto-require `Sensu_user[admin]`

## Motivation and Context

Sensu Go now makes it easier to manage passwords in such a way that using the old password is no longer necessary.  This deprecates the unnecessary password so that catalogs don't produce errors if people have provided the `old_password` parameters.

The changes to support this should be released in Sensu Go 5.21 but this should still work before 5.21 by passing the plain text password when managing users if the command to generate the hash fails.

There will be deprecation warnings for public resources with `old_password`. All private resources had `old_password` removed.

I've added hopefully catchy bold to pull request title to grab people's attention in CHANGELOG.

### DIFF
--- a/lib/puppet/provider/sensu_user/sensu_api.rb
+++ b/lib/puppet/provider/sensu_user/sensu_api.rb
@@ -74,6 +74,7 @@ Puppet::Type.type(:sensu_user).provide(:sensu_api, :parent => Puppet::Provider::
         configure_cmd << '--trusted-ca-file'
         configure_cmd << resource[:configure_trusted_ca_file]
       end
+      Puppet.notice('Executing sensuctl configure')
       Puppet::Provider::Sensuctl.sensuctl(configure_cmd)
     end
     @property_hash[:ensure] = :present
@@ -97,6 +98,7 @@ Puppet::Type.type(:sensu_user).provide(:sensu_api, :parent => Puppet::Provider::
           configure_cmd << '--trusted-ca-file'
           configure_cmd << resource[:configure_trusted_ca_file]
         end
+        Puppet.notice('Executing sensuctl configure')
         Puppet::Provider::Sensuctl.sensuctl(configure_cmd)
       end
     end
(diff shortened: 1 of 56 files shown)


lib/puppet/provider/sensu_user/sensu_api.rb
```diff
@@ -74,6 +74,7 @@ def create
         configure_cmd << '--trusted-ca-file'
         configure_cmd << resource[:configure_trusted_ca_file]
       end
+      Puppet.notice('Executing sensuctl configure')
```
ghoneycutt, line 77: Are these present in a puppet run? Could you paste an example?

manifests/backend.pp
```diff
@@ -190,6 +190,8 @@
     configure                 => true,
     configure_url             => $sensu::api_url,
     configure_trusted_ca_file => $sensu::trusted_ca_file,
+    provider                  => 'sensu_api',
```
ghoneycutt, line 193: Is it necessary to specify the provider? Isn't there only one?

lib/puppet/provider/sensu_user/sensu_api.rb
treydock: This is from vagrant where I changed admin password:

```
[root@sensu-backend ~]# puppet apply backend.pp 
Notice: Compiled catalog for sensu-backend in environment production in 1.16 seconds
Notice: /Stage[main]/Sensu::Backend/Sensu_user[admin]/password: changed password
Notice: Executing sensuctl configure
Notice: Applied catalog in 4.97 seconds
```

manifests/backend.pp
treydock: There are two providers for all types except those prefixed with `sensuctl`.  They are `sensuctl` and `sensu_api`.  By using the API to manage the admin user, removes potential circular dependency as sensuctl requires the admin user password and setting the admin user password via sensuctl would create a dependency problem when removing the `old_password` necessity. The API is able to either use existing `password` or reads tokens from the config files written by `sensuctl` which will exist after bootstrap and those are the "backup" form of API authentication on the backend host.

treydock: Also the default provider for all types is `sensuctl` as a form of backwards compatibility before the API provider existed.  In order to use the API you must specify the `sensu_api` provider.

PR #1252, opened 2020-05-24
Fix Windows acceptance tests and update Postgresql dependency range

treydock:
# Pull Request Checklist

## Description

Use latest Sensu Go for testing which is what happens when chocolatey tests run. This should resolve failures with appveyor as I ran the tests locally in Vagrant and had no issues.  Also fix version test because now the version is more than just `5.20.0` it's like `5.20.0+ee` which we fixed for Travis tests but missed Windows.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -15,6 +15,10 @@
     {
       "name": "richardc/datacat",
       "version_requirement": ">= 0.6.0 < 2.0.0"
+    },
+    {
+      "name": "puppetlabs/postgresql",
+      "version_requirement": ">= 6.4.0 < 7.0.0"
     }
   ],
   "operatingsystem_support": [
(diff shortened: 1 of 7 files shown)


metadata.json
```diff
@@ -15,6 +15,10 @@
     {
       "name": "richardc/datacat",
       "version_requirement": ">= 0.6.0 < 2.0.0"
+    },
+    {
+      "name": "puppetlabs/postgresql",
+      "version_requirement": ">= 6.0.0 < 7.0.0"
```
ghoneycutt, line 21: Wouldn't this be 6.4.0 as well to match `.fixtures.yml`

treydock: Yep, this is now fixed.

treydock:
For postgresql change, removed unnecessary soft dependency mention in README and also removed part in acceptance test setup as there is logic to automatically install modules in metadata.json, so those changes reduce duplication.

PR #1255, opened 2020-07-25
Support Sensu Go 6

treydock:
# Pull Request Checklist

## Description

Changes necessary to support Sensu Go 6.

Changes to `agent.yml` have to be made via API or sensuctl from the agent. This requires the agent to be aware of admin password.  Added `sensu_agent_entity_config` type to handle updating the config entries for agent entities. Added `sensu_agent_entity_validator` to ensure the agent entity is bootstrapped before attempting to add the config entries.

For backend added `sensu::backend_upgrade` Bolt task that can be used to perform the `sensu-backend upgrade` command using Bolt.

For now this is WIP because at this time redacted labels and annotations will return back from both API and sensuctl as `REDACTED` which makes puppet checking impossible.  There are also no README updates for the upgrade procedure and changes associated with this PR.


README.md
```diff
@@ -1047,6 +1047,10 @@ sensu_bonsai_asset { 'sensu/sensu-pagerduty-handler':
 
 The following Bolt tasks are provided by this Module:
 
+**sensu::backend\_upgrade**: Perform backend upgrade via `sensu-backend upgrade` command.
+
+Example: `bolt task run sensu::backend_upgrade --targets sensu_backend`
```
ghoneycutt, line 1052: I believe you only run on *one* of the backends, so the target will be different.

treydock: Right, maybe I should change `--targets sensu_backend` to `--targets sensu-backend.example.com` to make it clear it's FQDN but IIRC in Bolt a target name isn't necessarily a FQDN, it could be something else like Sensu entity name if using this modules target resolver.  But yes, I believe you only run the migration on one of the backends if using a cluster setup, as the change will replicate to other backends.

spec/acceptance/windows_spec.rb
```diff
@@ -77,7 +75,7 @@ class { 'sensu::agent':
           'backend-url'     => ['wss://sensu-backend:8081'],
           'password'        => 'P@ssw0rd!',
           'name'            => 'sensu-agent',
-          'subscriptions'   => ['base','windows'],
+          'namespace'       => 'default',
```
ghoneycutt, line 78: Could you please add a comment block at the top of this that explains our testing methodology and how backends are not supported on windows and because of that we can't test certain things.

treydock: Comment added at top of file.

README.md
```diff
@@ -55,19 +56,21 @@ Please note, that this is a **Partner Supported** module, which means that techn
...
 | 5.0 - 5.15      | latest v3                           |
 | 5.16+           | latest v4                           |
+| 6.0+            | latest v5                           |
 
 ### Upgrade note
 
 Sensu Go 5.x is a rewrite of Sensu and no longer depends on redis and rabbitmq.
 Version 3 of this module supports Sensu Go 5.0.0 to 5.15.x.
 Version 4 of this module supports Sensu Go 5.16+.
+Version 5 of this module supports Sensu Go 6.0+.
```
ghoneycutt, line 73: We should change to use notation like this
```
>= 5.0.0 < 5.16.0
>= 5.16.0 < 6.0.0
>= 6.0.0 < 7.0.0
```

README.md
```diff
@@ -103,6 +106,41 @@ sensu_asset { 'test':
...
+  ...
+}
+```
+
+See [API Providers](#api-providers) for example Hiera that can be used in a file like `common.yaml` to easily share the admin password with agents.
+
+This module will still continue to write subscriptions and other agent configurations to `agent.yml` so that if an agent entity is deleted it can be recreated
+by restarting the `sensu-agent` service.
+
+**NOTE**: At this time redaction of labels or annotations is not supported and will cause this module to produce errors. Do not use label or annotation redaction at this time. See [limitations](#limitations) for details.
```
ghoneycutt, line 142: Remove this phrase as it is redundant. 

redaction. <strike>at this time</strike>

ghoneycutt (changes requested):

README.md
treydock: Changed to use that notation

README.md
treydock: Phrase removed

ghoneycutt (approved):

PR #1260, opened 2020-08-11
Installation of sensu-plugins-ruby fails on CentOS8

gtufte:
Installation of sensu-plugins-ruby fails on CentOS8 since it is not present in the installed repository. This 'fix' fetches and installs the el7-package from a given URL-source

This PR provides the option to install sensu-plugins-ruby on RedHat/CentOS8 from a given URL source. 

## Description

This PR only affects systems within the RedHat family on major version 8 where the gem "sensu-plugins-ruby" is fetched and installed "manually". 

## Motivation and Context

I did this fix in a module created in-house, but ran quickly into duplicate declaration issues since we use this module for both CentOS 7 and CentOS 8. Thought it would be best to add the fix here.

## How Has This Been Tested?
Tok a CentOS 8 system without the gem. Added the fix. Ran puppet -  success. Tried several runs, everything works as expected. Also tried removing the package and the run another puppet run with success.

### DIFF
--- a/manifests/plugins.pp
+++ b/manifests/plugins.pp
@@ -39,6 +39,7 @@ class sensu::plugins (
   Optional[Boolean] $manage_repo = undef,
   String $package_ensure = 'installed',
   String $package_name = 'sensu-plugins-ruby',
+  String $package_url = "https://packagecloud.io/sensu/community/packages/el/7/sensu-plugins-ruby-0.2.0-1.el7.x86_64.rpm/download.rpm",
   Array $dependencies = [],
   Array $gem_dependencies = [],
   Variant[Array, Hash] $plugins = [],
@@ -61,10 +62,25 @@ class sensu::plugins (
     $package_require = undef
   }
 
-  package { 'sensu-plugins-ruby':
-    ensure  => $package_ensure,
-    name    => $package_name,
-    require => $package_require,
+  if "${facts['os']['family']}${facts['os']['release']['major']}" == "RedHat8" {
+    exec { 'fetch sensu-plugins-ruby rpm':
+      command => "/usr/bin/curl -L ${package_url} --output /opt/${package_name}.rpm",
+      creates => "/opt/${package_name}.rpm",
+      user    => 'root',
+    }
+    package { 'sensu-plugins-ruby':
+      ensure  => $package_ensure,
+      name    => $package_name,
+      require => Exec['fetch sensu-plugins-ruby rpm'],
+      source => "/opt/${package_name}.rpm",
+    }
+  }
+  else {
+    package { 'sensu-plugins-ruby':
+      ensure  => $package_ensure,
+      name    => $package_name,
+      require => $package_require,
+    }
   }
 
   ensure_packages($dependencies)


treydock:
There are some lint issues as well as unit test failures for the unit tests that run against RHEL8.

Also you will want to remove this from the acceptance tests: https://github.com/sensu/sensu-puppet/blob/master/spec/acceptance/04_plugins_spec.rb#L10-L12

manifests/plugins.pp
```diff
@@ -61,10 +62,25 @@
...
+    exec { 'fetch sensu-plugins-ruby rpm':
+      command => "/usr/bin/curl -L ${package_url} --output /opt/${package_name}.rpm",
+      creates => "/opt/${package_name}.rpm",
+      user    => 'root',
+    }
+    package { 'sensu-plugins-ruby':
+      ensure  => $package_ensure,
+      name    => $package_name,
+      require => Exec['fetch sensu-plugins-ruby rpm'],
+      source => "/opt/${package_name}.rpm",
```
treydock, line 75: The alignment of `=>` is one lint issue

manifests/plugins.pp
```diff
@@ -61,10 +62,25 @@
     $package_require = undef
   }
 
-  package { 'sensu-plugins-ruby':
-    ensure  => $package_ensure,
-    name    => $package_name,
-    require => $package_require,
+  if "${facts['os']['family']}${facts['os']['release']['major']}" == "RedHat8" {
```
treydock, line 65: The double quotes around `RedHat8` should be single quotes

manifests/plugins.pp
```diff
@@ -39,6 +39,7 @@
   Optional[Boolean] $manage_repo = undef,
   String $package_ensure = 'installed',
   String $package_name = 'sensu-plugins-ruby',
+  String $package_url = "https://packagecloud.io/sensu/community/packages/el/7/sensu-plugins-ruby-0.2.0-1.el7.x86_64.rpm/download.rpm",
```
treydock, line 42: This should be single quotes as there are no variables.

treydock: There will also need to be documentation at the top of file like this:

```
# @param package_url
#    The URL of the package. Only used for systems based on RedHat 8
```

ghoneycutt:
I'm not sure I follow the reasoning behind this. Couldn't you just make the correct package available in your yum repositories and then the code would work as it is?

ghoneycutt:
Thank for you opening this pull request, though I think the approach is not the right way to address this problem. I'm going to close this as the fix is to simply take the EL7 package and make it available in a yum repo for the EL8 systems until there is a proper EL8 package. Repository management is out of scope for this module so we should not be implementing a work around for what is essentially making the right package available through package management system.

If you need assistance managing your yum repositories on a host or setting up a yum server, check out https://github.com/ghoneycutt/puppet-module-yum as it is feature complete to the yum documentation.

treydock:
@gtufte Along with @ghoneycutt suggestion you can define the download of the RPM in profile class then override the package in this module like this:

```puppet
Package <| title == 'sensu-plugins-ruby' |> {
  source => '...value goes here',
  ... override other parameters here
}
```

As long as the above comes after `sensu::plugin`it will allow you to change how the package is defined.

PR #1264, opened 2020-08-12
Improve sensu-backend init

### LINKED ISSUE #1261: Improve sensu-backend init
Currenlty `sensu-backend init` is only executed on Package refresh for `sensu-go-backend` which means it's possible for a failure to result in backend not getting re-run if needed.

The exec either needs to work by looking for exit code 3 to know if init has run or move to native type that looks at exit code.

treydock:
# Pull Request Checklist

## Description

Rather than only running `sensu-backend init` on refresh of Package, run it if the exit code is not 3. 

## Related Issue

Fixes #1261

### DIFF
--- a/manifests/backend.pp
+++ b/manifests/backend.pp
@@ -343,13 +343,15 @@ class sensu::backend (
 
   exec { 'sensu-backend init':
     path        => '/usr/bin:/bin:/usr/sbin:/sbin',
-    refreshonly => true,
     environment => [
       'SENSU_BACKEND_CLUSTER_ADMIN_USERNAME=admin',
       "SENSU_BACKEND_CLUSTER_ADMIN_PASSWORD=${sensu::password}",
     ],
     returns     => [0, 3],
-    subscribe   => Package['sensu-go-backend'],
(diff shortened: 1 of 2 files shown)


manifests/backend.pp
```diff
@@ -343,13 +343,12 @@
...
   exec { 'sensu-backend init':
     path        => '/usr/bin:/bin:/usr/sbin:/sbin',
-    refreshonly => true,
     environment => [
       'SENSU_BACKEND_CLUSTER_ADMIN_USERNAME=admin',
       "SENSU_BACKEND_CLUSTER_ADMIN_PASSWORD=${sensu::password}",
     ],
     returns     => [0, 3],
-    subscribe   => Package['sensu-go-backend'],
+    unless      => 'sensu-backend init ; [ $? -eq 3 ] && exit 0 || exit 1',
```
ghoneycutt, line 351: Given the complexity of inline shell code and an odd exit code, please add a comment above this resource to explain what is happening and why.

treydock: Added comments.

PR #1265, opened 2020-08-13
Add sensu::backend_upgrade task

treydock:
# Pull Request Checklist

## Description

Adds the `sensu::backend_upgrade` bolt task that can run `sensu-backend upgrade` during Sensu Go 6 upgrade.

This was pulled out of #1255


treydock:
The one real failure with CI was likely race condition with events getting added to database when using postgresql.  The other failures are with CI repos and those are allowed failures.  I restarted the one real failure to see if it happens again.

PR #1266, opened 2020-08-13
Make sensu_ad_auth group_search optional

treydock:
# Pull Request Checklist

## Description

The `group_search` when configuring `sensu_ad_auth` is now optional.

## Motivation and Context

Sensu Go 6 added a feature to support AD memberOf attribute for group lookups on users rather than searching groups.  To do this you must not define `group_search` when configuring AD auth.

### DIFF
--- a/lib/puppet/type/sensu_ad_auth.rb
+++ b/lib/puppet/type/sensu_ad_auth.rb
@@ -28,6 +28,25 @@ Puppet::Type.newtype(:sensu_ad_auth) do
     ],
   }
 
+@example Add an AD auth that uses memberOf attribute by omitting group_search
+  sensu_ldap_auth { 'ad':
+    ensure              => 'present',
+    servers             => [
+      {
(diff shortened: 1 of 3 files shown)


treydock:
Docs for this from Sensu: https://docs.sensu.io/sensu-go/latest/operations/control-access/auth/#active-directory-ad-authentication

lib/puppet/type/sensu_ad_auth.rb
```diff
@@ -57,7 +76,7 @@
     Keys:
     * host: required
     * port: required
-    * group_search: required
+    * group_search: optional (omit to use memberOf)
```
ghoneycutt, line 79: A note here that says it is required with Sensu 5 and as of Sensu 6 it is optional.

treydock: Note updated.

PR #1267, opened 2020-08-13
Add support for Sensu Go Web

### LINKED ISSUE #1263: sensu Web UI community version installation support
## information request

will you support the installation of the new Sensu Web UI for the community version: https://github.com/sensu/web ?
This is being used in the community version only and I wonder whether I should wait for you to add this component to the puppet module, or f I need to do it myself.

treydock:
# Pull Request Checklist

## Description

Adds support for Sensu Go Web.

There is code in this project that requires being run from a git repo, as I tried downloading the tag tarball and running and that did not work.  Because there is no good way to download latest stable release, we must define a version, so updated docs to reflect this being a key parameter.  This is only supported for systemd hosts even though this could run on non-systemd.  The reason was simply the simplicity of deploying a systemd service and not having to maintain a complicated sysVinit script.

## Related Issue

Fixes #1263

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -75,6 +75,8 @@ add_custom_fact :service_provider, ->(os, facts) {
     else
       'systemd'
     end
+  when 'windows'
+    'windows'
   else
     'systemd'
   end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -67,6 +67,8 @@ add_custom_fact :puppet_localcacert, ->(os, facts) {
   end
 }
 
+# Must add this fact from stdlib because
+# rspec-puppet-facts and facterdb only mock core Facter facts
 add_custom_fact :service_provider, ->(os, facts) {
   case facts[:osfamily]
   when 'RedHat'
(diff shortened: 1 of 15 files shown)


spec/spec_helper.rb
```diff
@@ -67,6 +67,21 @@
   end
 }
 
+add_custom_fact :service_provider, ->(os, facts) {
```
ghoneycutt, line 70: What's this for? Can't it just use the actual service_provider fact?

treydock: The library for rspec-puppet-facts pulls facts from facterdb gem and that gem only includes core Facter facts, not facts from stdlib.  This fact is from stdlib module so the only way to incorporate it into rspec-puppet-facts is to add it using `add_custom_fact`.  This is not using real facter facts during unit tests, that's not possible so all facts are mocked using rspec-puppet-facts to produce real-looking facts during rspec tests.

ghoneycutt: Could you please add a comment to reflect that.

treydock: comment added

portertech:
Sensu Go Web ([sensu/web](https://github.com/sensu/web)) is out of scope for this Puppet module. The commercially supported Sensu Go builds and packages, already supported by this Puppet module, include the web UI (it runs as part of the Sensu Go Backend process). The Puppet code developed on this branch/PR can be reused by Puppet users outside of this module.

PR #1274, opened 2020-10-05
Support Sensu Go 6.1

treydock:
# Pull Request Checklist

## Description

This adds support for many changes associated with Sensu Go 6.1.

* Add back support for redacted labels/annotations on agents
* Use PATCH API query when updating agent entity configs (support for fallback to PUT for < 6.1.0)
  * Deprecate `sensuctl` provider for agent entities configs since PATCH support only works with API and not sensuctl.  The default was and still is to use API
* Add `disable_offline_access` property to `sensu_oidc_auth` type
* Add `strict`, `batch_buffer`, `batch_size`, `batch_workers`, properties to `sensu_postgres_config` type
  * Add corresponding `postgresql_*` parameters to `sensu::backend` class for these new properties
* Allow `prometheus_text` value for `output_metric_format` property for `sensu_check`


lib/puppet/provider/sensu_oidc_auth/sensuctl.rb
```diff
@@ -68,6 +74,13 @@ def create
       value = resource[property]
       next if value.nil?
       next if value == :absent || value == [:absent]
+      if min_version.key?(property)
+        v = min_version[property]
+        if ! version_cmp(v)
+          Puppet.warning("Sensu_oidc_auth[#{resource[:name]}] Property #{property} skipped, does not mean minimum Sensu Go version of #{v}")
```
ghoneycutt, line 80: does not mean? perhaps does not meet

manifests/agent.pp
```diff
@@ -88,22 +88,17 @@
   Optional[Hash[String[1],String]] $annotations = undef,
   Optional[Hash[String[1],String]] $labels = undef,
   String[1] $namespace = 'default',
-  Optional[Array[String[1]]] $redact = undef,
+  Array[String[1]] $redact = ['password','passwd','pass','api_key','api_token','access_key','secret_key','private_key','secret'],
```
ghoneycutt, line 91: are these the default values that sensu ships with?

lib/puppet/provider/sensu_postgres_config/sensuctl.rb
```diff
@@ -5,6 +5,15 @@
 
   mk_resource_methods
 
+  def min_version
```
ghoneycutt, line 8: Are there spec tests that show the warning when the version is below the minimum?

manifests/agent.pp
treydock: Those are the defaults from Sensu.

lib/puppet/provider/sensu_oidc_auth/sensuctl.rb
treydock: I fixed typo.

lib/puppet/provider/sensu_postgres_config/sensuctl.rb
treydock: Not possible it seems.

This is not catching the warning despite it clearly getting called:
```ruby
expect(Puppet).to receive(:warning)
```

Ran the code in Vagrant with Sensu Go 6.0.0 installed and get this:

```
Warning: Sensu_postgres_config[postgresql] Property strict skipped, does not meet minimum Sensu Go version of 6.1.0
Warning: Sensu_postgres_config[postgresql] Property batch_buffer skipped, does not meet minimum Sensu Go version of 6.1.0
Warning: Sensu_postgres_config[postgresql] Property batch_size skipped, does not meet minimum Sensu Go version of 6.1.0
Warning: Sensu_postgres_config[postgresql] Property batch_workers skipped, does not meet minimum Sensu Go version of 6.1.0
```

This must be either a limitation with rspec or some quirk with how Puppet.warning is defined in Ruby to make the call not something we can test.  The best we can do is test that the `next` line is called which is tested by showing the new properties are omitted from being set when not supported.

ghoneycutt (approved):

PR #1275, opened 2020-10-12
Add output_metric_tags property to sensu_check

treydock:
# Pull Request Checklist

## Description

Adds the `output_metric_tags` property to `sensu_check`.  This was something added in Sensu Go 6.1. Because there is no default value applied, there is not version check since this new property is opt-in.

https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-schedule/checks/#output-metric-tags


ghoneycutt (approved):

ghoneycutt:
Released in v5.2.0

PR #1280, opened 2020-10-23
Improved handling of changes to sensu::etc_dir

treydock:
# Pull Request Checklist

## Description

Add support for changing `etc_dir` in the `sensu` class.  Previously changing that value would move configs but not tell the Sensu daemons where to find the configs.

This handling only works with systemd based sytems, adds dependency on campotcamp/systemd

Added a few improvements to documentation around `sensu` class and how to properly ensure agents can communicate with backend API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,10 +100,11 @@ Beginning with Sensu Go 6, some changes to `agent.yml` will only bootstrap an ag
 If you wish to make changes to values such as `subscriptions`, `labels` or `annotations` after a host is added to Sensu this must be done
 via the Sensu Go API. To support this it's now required that agents have the ability to make API calls.
 
-In order to ensure agents can make API calls either via API or sensuctl the agent must be told about the admin password:
+In order to ensure agents can make API calls either via API or sensuctl the agent must be told about the admin password and API host:
 
 ```
 class { 'sensu':
+  api_host                     => 'sensu-backend.example.com',
   agent_entity_config_password => 'supersecret',
 }
 class { 'sensu::agent':
@@ -324,6 +325,10 @@ sensu::agent_entity_config_password: supersecret
 This module supports Windows Sensu Go agent via chocolatey beginning with version 5.12.0.
 
 ```puppet
+class { 'sensu':
+  api_host                     => 'sensu-backend.example.com',
+  agent_entity_config_password => 'supersecret',
+}
 class { 'sensu::agent':
   backends      => ['sensu-backend.example.com:8081'],
   subscriptions => ['windows'],
@@ -1219,6 +1224,8 @@ Examples can be found in the [examples](https://github.com/sensu/sensu-puppet/tr
 * [Slack Alerts](https://github.com/sensu/sensu-puppet/blob/master/examples/slack_alerts.pp) - Example of setting up Slack alerts
 
 ## Limitations
+
+Changing `sensu::etc_dir` is only supported on systems using systemd.
 
 The type `sensu_user` does not at this time support `ensure => absent` due to a limitation with sensuctl, see [sensu-go#2540](https://github.com/sensu/sensu-go/issues/2540).
 
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -35,6 +35,8 @@
 #   Sensu agent service ensure value.
 # @param service_enable
 #   Sensu agent service enable value.
+# @param service_path
+#   The path to sensu-agent service executable
 # @param config_hash
 #   Sensu agent configuration hash used to define agent.yml.
 # @param backends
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -1220,6 +1225,8 @@ Examples can be found in the [examples](https://github.com/sensu/sensu-puppet/tr
 
 ## Limitations
 
+Changing `sensu::etc_dir` is only supported on systems using systemd.
```
ghoneycutt, line 1228: This specifically excludes X, Y and Z

spec/acceptance/00_backend_spec.rb
```diff
@@ -92,6 +92,42 @@ class { 'sensu::backend':
     end
   end
 
+  context 'etc_dir change' do
+    it 'should work without errors' do
```
ghoneycutt, line 96: To ensure that the old, default is not being used, perhaps add something like

```puppet
file { '/etc/sensu':
  ensure  => absent,
  recurse => true,
}
```

spec/acceptance/01_agent_spec.rb
```diff
@@ -92,6 +92,56 @@ class { 'sensu::agent':
     end
   end
 
+  context 'etc_dir changed' do
+    it 'should work without errors' do
+      pp = <<-EOS
+      class { '::sensu':
```
ghoneycutt, line 98: same here with ensuring that `/etc/sensu` no longer exists

ghoneycutt (changes requested):

README.md
treydock: Do we really want to define which OSes do not support Systemd? I feel like an administrator should know that and puts us in the business of telling admins how their systems work.  I also do not know which OSes we support work with systemd so that will take some time to dig up if we really want that information to be documented.

spec/acceptance/00_backend_spec.rb
treydock: That won't tell us anything that isn't already being tested. The test for port change configs are how we test this works.  We modify the `api-port` via the YAML config.  If that YAML config was not read from new location the port would not get changed.  Testing the new config gets used is a more direct test anyways.

spec/acceptance/01_agent_spec.rb
treydock: per other comment we are already testing that.

```
'api-port' => 4041,
```

We test that the port we define in configs is actually being used and listening, if the config wasn't being used the default port would get used which is different port number.

README.md
treydock: Looks like Ubuntu you can opt-in to systemd on systems that don't use systemd by default, so this gets into some territory where we can't say which OS has systemd because it's really a configuration choice. The only one I can say for sure is EL6 so seems like just saying "systems using systemd" is about as good a description as we can give without potentially falsely labeling some OSes as not supported when they could be supported if configured differently.

ghoneycutt: OK, makes sense.

spec/acceptance/00_backend_spec.rb
ghoneycutt: add a comment that explains this

treydock: Comments added, also to agent acceptance tests that test the same thing.

treydock:
The tests in this PR will continue to fail until #1281 is merged and this is rebased with those changes.

ghoneycutt (approved):

manifests/agent.pp
```diff
@@ -217,8 +220,8 @@
   # Only necessary for Puppet < 6.1.0,
   # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
   if versioncmp($facts['puppetversion'],'6.1.0') < 0 and $facts['service_provider'] == 'systemd' {
-    Package['sensu-go-agent'] ~> Exec['sensu systemctl daemon-reload']
-    Exec['sensu systemctl daemon-reload'] -> Service['sensu-agent']
+    Package['sensu-go-agent'] ~> Class['systemd::systemctl::daemon_reload']
+    Class['systemd::systemctl::daemon_reload']-> Service['sensu-agent']
```
ghoneycutt, line 224: spaces around arrow

manifests/backend.pp
```diff
@@ -234,8 +237,8 @@
   # Only necessary for Puppet < 6.1.0,
   # See https://github.com/puppetlabs/puppet/commit/f8d5c60ddb130c6429ff12736bfdb4ae669a9fd4
   if versioncmp($facts['puppetversion'],'6.1.0') < 0 and $facts['service_provider'] == 'systemd' {
-    Package['sensu-go-backend'] ~> Exec['sensu systemctl daemon-reload']
-    Exec['sensu systemctl daemon-reload'] -> Service['sensu-backend']
+    Package['sensu-go-backend'] ~> Class['systemd::systemctl::daemon_reload']
+    Class['systemd::systemctl::daemon_reload']-> Service['sensu-backend']
```
ghoneycutt, line 241: spaces around arrow

ghoneycutt:
Added a lint knit since you have to rebase anyhow.

manifests/agent.pp
treydock: This change was made

manifests/backend.pp
treydock: This change was made

treydock:
This PR was rebased with fixes from #1281

PR #1281, opened 2020-10-23
Fix sensu_agent_entity_config purging to not purge entity subscription

treydock:
# Pull Request Checklist

## Description

Avoid purging the entity subscription that cannot be deleted.  This was actually not an issue with 6.1.0 because of a bug with Sensu Go but the fix with 6.1.1 now correctly makes this entity subscription not be able to be deleted so we must ignore it when purging.

### DIFF
--- a/lib/puppet/type/sensu_resources.rb
+++ b/lib/puppet/type/sensu_resources.rb
@@ -131,6 +131,11 @@ Puppet::Type.newtype(:sensu_resources) do
     if resource[:key] == 'sensu.io/managed_by'
       return false
     end
+    # Do not purge entity subscription as it can not be deleted
+    # https://github.com/sensu/sensu-puppet/pull/1280
+    if resource[:config] == 'subscriptions' && resource[:value] == "entity:#{resource[:entity]}"
+      return false
+    end
     if ! self[:agent_entity_configs].nil? && ! self[:agent_entity_configs].include?(resource[:config])
       return false
     end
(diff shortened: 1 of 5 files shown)


lib/puppet/type/sensu_resources.rb
```diff
@@ -131,6 +131,9 @@ def check_agent_entity_config(resource)
     if resource[:key] == 'sensu.io/managed_by'
       return false
     end
+    if resource[:config] == 'subscriptions' && resource[:value] == "entity:#{resource[:entity]}"
```
ghoneycutt, line 134: Please add a comment here explaining the why and link to issue/pr.

spec/unit/sensu_resources_spec.rb
```diff
@@ -121,6 +121,17 @@
     expect(catalog.resources.size).to eq(1)
     expect(ret).to include(instance_config)
   end
+  it 'should not purge sensu_agent_entity_config for subscription entity' do
```
ghoneycutt, line 124: copy the comment here

ghoneycutt (changes requested):

spec/unit/sensu_resources_spec.rb
treydock: comment added

lib/puppet/type/sensu_resources.rb
treydock: comment added

ghoneycutt (approved):

PR #1284, opened 2020-11-05
Fix agent so it uses the correct api-url

ghoneycutt:
Without this patch, the agent generates the api-url based on the first backend in the array as opposed to using the actual api_url specified through `sensu::api_host` and `sensu::api_port`.


ghoneycutt:
Closing in favor of #1286

PR #1288, opened 2020-11-22
Support Puppet 7

treydock:
# Pull Request Checklist

## Description

Add support for Puppet 7 (and Facter 4)

I did not use Ruby 2.7.0 for acceptance tests because the RVM version of Ruby used for beaker tests has zero impact on the tests, the version of Ruby and Puppet used for the test is set by `BEAKER_PUPPET_COLLECTION` so you can use whatever version of Ruby is supported by the beaker code and it won't impact the version of Ruby used for actual acceptance tests.  The beaker library is not yet validated on Ruby 2.7 so I opted for safest approach while still testing proper version of Puppet and Ruby inside Docker.  It's sort of like installing RVM Ruby 2.0.0 on a host and it having no impact on Puppet Ruby because Puppet Ruby is embedded in the puppet-agent package, and the beaker tests are using puppet-agent packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,10 @@ matrix:
     env: CHECK="validate lint strings:generate reference" PUPPET_GEM_VERSION="~> 6"
     stage: unit
     bundler_args: --without system_tests development
+  - rvm: 2.5.7
+    env: CHECK="validate lint strings:generate reference" PUPPET_GEM_VERSION="~> 7"
+    stage: unit
+    bundler_args: --without system_tests development
   - rvm: 2.4.9
     env: CHECK="parallel_spec" PUPPET_GEM_VERSION="~> 5"
     stage: unit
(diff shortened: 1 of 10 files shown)


treydock:
Modified pull request to only add Puppet 7 support.

.travis.yml
```diff
@@ -69,7 +81,7 @@ matrix:
     stage: acceptance
   - rvm: 2.5.7
     services: docker
-    env: BEAKER_set="centos-7" BEAKER_PUPPET_COLLECTION=puppet6 BEAKER_sensu_mode=full
```
ghoneycutt, line 72: why do full runs with puppet versions 5 and 7 and not 6?

treydock: I think that was copy + paste mistake in vim.  I put the puppet6 tests back.

spec/acceptance/99_facts_spec.rb
```diff
@@ -3,6 +3,15 @@
 describe 'sensu::backend class', if: ['base','full'].include?(RSpec.configuration.sensu_mode) do
   backend = hosts_as('sensu-backend')[0]
   agent = hosts_as('sensu-agent')[0]
+  let(:facter_command) do
```
ghoneycutt, line 6: This code is getting duplicated. To DRY things up, could it be moved to the spec_helper?

treydock: Not really. It's best to define variables in rspec using `let` to ensure proper test isolation and not a great way to do that in central place that I know of. It's also slightly differen logic for Windows tests than other tests so not much to DRY up.

PR #1291, opened 2020-12-03
Add token_file parameter to sensu_secrets_vault_provider

treydock:
# Pull Request Checklist

## Description

Adds the `token_file` parameter to `sensu_secrets_vault_provider` that allows the token to be read from a file rather than from the `token` property.


ghoneycutt (approved):
Excellent work!

PR #1298, opened 2021-01-22
Support Sensu Go 6.2

treydock:
# Pull Request Checklist

## Description

Main changes:

* Allow `agent.yml` to become authoritative by setting `sensu::agent::agent_managed_entity: true`
* Support PostgreSQL enabling round robin scheduling
* Allow use of `memberOf` with `sensu_ldap_auth`
* Add support for Ubuntu 20.04

### DIFF
--- a/spec/classes/api_spec.rb
+++ b/spec/classes/api_spec.rb
@@ -1,7 +1,10 @@
 require 'spec_helper'
 
 describe 'sensu::api', :type => :class do
-  on_supported_os.each do |os, facts|
+  # Only test 1 OS to speed up tests when behavior does not vary based on OS facts
+  on_supported_os({
+    supported_os: [{ 'operatingsystem' => 'RedHat', 'operatingsystemrelease' => ['7'] }]
+  }).each do |os, facts|
     context "on #{os}" do
       let(:facts) { facts }
       let(:node) { 'test.example.com' }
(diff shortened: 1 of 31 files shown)


spec/classes/api_spec.rb
```diff
@@ -1,7 +1,9 @@
 require 'spec_helper'
 
 describe 'sensu::api', :type => :class do
-  on_supported_os.each do |os, facts|
+  on_supported_os({
+    supported_os: [{ 'operatingsystem' => 'RedHat', 'operatingsystemrelease' => ['7'] }]
```
ghoneycutt, line 5: Why did this change?

treydock: To speed up unit tests.  This and several other unit test files don't change based on OS tested, and we already did this pattern on a few other files before this pull request.  The classes that actually change based on OS facts still test all OSes defined in metadata.json.

ghoneycutt: Please add a comment with each that explains why this is happening.

ghoneycutt (approved):
Please merge after adding comments. All tests already passing.

spec/classes/api_spec.rb
treydock: Comment added

PR #1301, opened 2021-01-23
Fix descriptions

### LINKED ISSUE #1300: Some descriptions are missing
Running `pdk bundle exec rake reference` shows the following descriptions are missing.

```
[warn]: Missing a description for Puppet resource property 'ensure' at lib/puppet/type/sensu_agent_entity_config.rb:57.
[warn]: Missing description for param 'namespace' in Puppet Task resolve_reference.
[warn]: Missing description for param 'subscription' in Puppet Task resolve_reference.
[warn]: Missing description for param 'interface_list' in Puppet Task resolve_reference.
[warn]: Missing description for param 'uri_ipaddress' in Puppet Task resolve_reference.
```

treydock:
# Pull Request Checklist

## Description

Use `ensurable` without a block, which has the same result.  I think the block was used during early development before the final solution for provider was created.  The Puppet code responsible for `ensuable` maintains what was deleted here.

The `uri_address` for the resolve_reference task never did anything and was never documented in README where we document what that task does.  I am not sure why it was ever present but it doesn't do anything.

## Related Issue

Fixes #1300

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -4846,25 +4846,19 @@ Generate targets from Sensu Go
 
 Data type: `Optional[String[1]]`
 
-
+Namespace to pull entities from
 
 ##### `subscription`
 
 Data type: `Optional[String[1]]`
 
-
+Find entities with this subscription
 
 ##### `interface_list`
 
 Data type: `Optional[Array]`
 
-
-
-##### `uri_ipaddress`
-
-Data type: `Optional[Boolean]`
-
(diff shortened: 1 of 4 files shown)


ghoneycutt (approved):

treydock:
@ghoneycutt The only test failures are CI enabled tests which should be allowed to fail but for some reason appear to still make the status of over all pull request show as failed

PR #1305, opened 2021-02-09
Support refresh of sensu_secrets_vault_provider token_file

treydock:
# Pull Request Checklist

## Description

Allow the file for sensu_secrets_vault_provider's token_file parameter to trigger a refresh that will force an update of the token value.

The acceptance tests were changed so that the resources testing token_file had only change taking place being the token_file contents to verify that a refresh event triggers the change without other properties triggering a change.  The changes of other properties are tested with other resources in the same tests.

### DIFF
--- a/spec/acceptance/sensu_secrets_spec.rb
+++ b/spec/acceptance/sensu_secrets_spec.rb
@@ -21,10 +21,15 @@ describe 'sensu_secrets_vault_provider', if: RSpec.configuration.sensu_mode == '
           "burst" => 100
         },
       }
+      file { '/tmp/secret':
+        ensure  => 'file',
+        content => "supersecret\n",
+        notify  => Sensu_secrets_vault_provider['my_vault-token_file'],
+      }
       sensu_secrets_vault_provider { 'my_vault-token_file':
         ensure       => 'present',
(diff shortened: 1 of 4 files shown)


spec/acceptance/sensu_secrets_spec.rb
```diff
@@ -171,16 +175,24 @@
...
+        ensure  => 'file',
+        content => "supersecret2\n",
+        notify  => Sensu_secrets_vault_provider['my_vault-token_file'],
+      }
       sensu_secrets_vault_provider { 'my_vault-token_file':
         ensure       => 'present',
-        address      => "https://vaultserver.example.com:8201",
-        token_file   => '/tmp/secret',
+        address      => "https://vaultserver.example.com:8200",
+        token_file   => "/tmp/secret",
```
ghoneycutt, line 186: We should be consistent in using single quotes for non-interpolated strings.

treydock: I only changed this particular line back to single quotes. The majority of the strings in this file and other files used for acceptance testing are using double quotes and will generate a very large diff if I start changing them all.

PR #1315, opened 2021-11-10
Set SensuAPIValidator timeout comments to 30 seconds

robmcelhinney:
Match comments and REFERENCE file.

# Pull Request Checklist

Comments in `sensu_api_validator.rb` and REFERENCE.md files say that the timeout is set to 15 but it's actually set to 30.
Set it to 15 to keep it consistent.

### DIFF
--- a/lib/puppet/type/sensu_api_validator.rb
+++ b/lib/puppet/type/sensu_api_validator.rb
@@ -42,7 +42,7 @@ DESC
   end
 
   newparam(:timeout) do
-    desc 'The max number of seconds that the validator should wait before giving up and deciding that sensu_api is not running; defaults to 15 seconds.'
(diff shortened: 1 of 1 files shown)


ghoneycutt:
Would prefer to keep the default as is and instead update any documentation that is not aligned.

robmcelhinney:
> 
> 
> Would prefer to keep the default as is and instead update any documentation that is not aligned.

Ok, I changed the comments to `30` seconds then to keep it consistent.
I'll just use 15 on my config, I prefer keeping it shorter.

ghoneycutt (changes requested):

ghoneycutt:
Thanks @robmcelhinney for finding and fixing this!

PR #1317, opened 2021-12-02
Fix sensu_api_port spelling in sensu_api_validator example

robmcelhinney:
Fix spelling mistake in example


ghoneycutt:
Thank you!

PR #1327, opened 2022-10-17
Remove namespace validation for agents

treydock:
Remove namespace validation for agents
Ensure entity config validator actually gets included

This will avoid validations on agents that agents can't fix.


treydock:
The failures with fixtures-latest are allowed but the Github visualization on PR doesn't convey that information.  Have to go here to see that: https://github.com/sensu/sensu-puppet/actions/runs/3267931461 - the entire run is green with allowed failures in the run.

Issue #1322, opened 2022-05-09
Deploy sensu agent without validation (only the config file)


davidphay:
## Description of problem

- What did you do? 
I try to install sensu agent without the backend

- What happened?
I have an error because we expose only the socket (API is unavailable)

- What did you expect to happen?
I only want configure the sensu agent with the file(agent.yml) without any validation (sensu API or sensuctl)

- How can someone reproduce the problem?
use a fake URL and `agent_managed_entity` param to true.

## Command used and debugging output
```
class {'sensu':
  use_ssl => true,
  api_host => 'my-fake-backend.lab',
  api_port => 443,
}

class { 'sensu::agent':
  backends      => ["my-fake-backend.lab:443"],
  subscriptions => ['linux'],
  labels      => { 'location' => 'uswest', 'contacts' => 'ops@example.com' },
  annotations => { 'cpu.warning' => '90', 'cpu.critical' => '100' },
  entity_name => "test2",
  namespace => "LAB",
  agent_managed_entity  => true,
}
```
- What does your puppet config look like (including any hiera config)

- Is this a masterless or master based puppet setup? 
master based setup

## Platform and version information

- Your OS: Linux CentOS 7
- Your Ruby version: 2.4
- Your version of Puppet: Puppet 7 CE
- Your version of Sensu: 6.6
- Your version of this module:  5.8

## Anything else to add that you think will be helpful?
```
Notice: /Stage[main]/Sensu::Agent/Service[sensu-agent]: Triggered 'refresh' from 1 event
Error: Unable to connect to https://my-fake-backend.lab:443/api/core/v2/namespaces/LAB/entities/test2: Failed to open TCP connection to my-fake-backend.lab:443 (getaddrinfo: Name or service not known)
Notice: Failed to connect to validate entity test2; sleeping 2 seconds before retry
Error: Unable to connect to https://my-fake-backend.lab:443/api/core/v2/namespaces/LAB/entities/test2: Failed to open TCP connection to my-fake-backend.lab:443 (getaddrinfo: Name or service not known)
Notice: Failed to connect to validate entity test2; sleeping 2 seconds before retry
Error: Unable to connect to https://my-fake-backend.lab:443/api/core/v2/namespaces/LAB/entities/test2: Failed to open TCP connection to my-fake-backend.lab:443 (getaddrinfo: Name or service not known)
Notice: Failed to connect to validate entity test2; sleeping 2 seconds before retry
[...]

PR #1328, opened 2022-11-07
Adding the option to opt out on entity validation

gtufte:
# Pull Request Checklist

Adding the option to opt out on entity validation

## Description

When the agent is configured to be self managed, both connecting to the Sensu API for agent configuration and validation gets in the way.

## Related Issue

https://github.com/sensu/sensu-puppet/issues/1322

Fixes # .

## Motivation and Context

I use Sensu myself and I've always deployed agents to be "standalone" and would like to be able to continue that way.

## How Has This Been Tested?

Deployed it within my org, added:
```
sensu::agent::agent_managed_entity: true
sensu::agent::validate_agent_entity: false
```
to my Hiera-config, and have thing working as before

## General

I might need some help in this PR with tests etc.

### DIFF
--- a/manifests/agent.pp
+++ b/manifests/agent.pp
@@ -74,7 +74,8 @@
 #   Defaults to `C:\ProgramData\sensu\log\sensu-agent.log`
 # @param agent_entity_config_provider
 #   The provider to use when managing sensu_agent_entity_config resources
-#
+# @param validate_agent_entity
+#   Determines if sensuctl and sensu_api types will validate if the entity exists within its namespace
 class sensu::agent (
   Optional[String] $version = undef,
   Optional[String[1]] $package_source = undef,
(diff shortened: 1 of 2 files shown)


manifests/agent.pp
```diff
@@ -295,9 +297,11 @@
     subscribe => $service_subscribe,
   }
 
-  sensu_agent_entity_validator { $config['name']:
-    ensure    => 'present',
-    namespace => $config['namespace'],
-    provider  => 'sensu_api',
+  if $validate_agent_entity {
```
treydock, line 300: I think you could get rid of this parameter and just use `agent_managed_entity` and if the agent manages the entity, don't use the validator.  This is already how some behavior works like here: https://github.com/sensu/sensu-puppet/blob/5a067b9713434e7e791b6ae887e1580f326efeaa/manifests/agent/annotation.pp#L52

PR #1337, opened 2023-06-01
Use maintained systemd dependencies

jfroche:
## Description

camptocamp module for systemd is not maintained since 16/08/21, use voxpupuli module instead

## Motivation and Context
camptocamp systemd is no longer maintained and have old dependencies (such as stdlib) that prevents Puppet modules upgrade

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -21,7 +21,7 @@
       "version_requirement": ">= 6.4.0 < 8.0.0"
     },
     {
-      "name": "camptocamp/systemd",
+      "name": "puppet/systemd",
       "version_requirement": ">= 2.0.0 < 4.0.0"
     }
   ],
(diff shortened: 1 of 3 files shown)


metadata.json
```diff
@@ -21,7 +21,7 @@
       "version_requirement": ">= 6.4.0 < 8.0.0"
     },
     {
-      "name": "camptocamp/systemd",
+      "name": "puppet/systemd",
       "version_requirement": ">= 2.0.0 < 4.0.0"
```
treydock, line 25: The fixtures has 4.2.0 so this version range needs to be updated to include the latest version released.

yorickps: @treydock the version limits have been updated

yorickps:
Can this PR be reviewd? Thanks.

Issue #1347, opened 2025-06-05
Deprecated functions


mburlic:
error in puppetserver.log

`"Puppet This function is deprecated, please use stdlib::to_yaml instead. at [\"/etc/puppetlabs/code/environments/production/modules/sensu/manifests/backend.pp\", 329]:[\"/etc/puppetlabs/code/environments/production/modules/argo/manifests/mon/sensu.pp\", 48]\n   (location: /etc/puppetlabs/code/environments/production/modules/stdlib/lib/puppet/functions/deprecation.rb:35:in `deprecation'; /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/dispatch.rb:60:in `invoke'; /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/dispatcher.rb:47:in `block in dispatch'; /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/dispatcher.rb:46:in `catch'; /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/dispatcher.rb:46:in `dispatch'; /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/pops/functions/function.rb:43:in `block in call'; /opt/puppetlabs/puppet/lib/ruby/vend`

Please check code for deprecated functions.
Thank you

Issue #1346, opened 2025-03-24
sensu_check Puppet resource does not support pipelines parameter.


saulfigueiredo-jam247:
I'm using:
- puppet --version : 8.7.0
- modules: sensu/sensu 5.11.1 (latest)

I am trying to apply a pipeline to my Sensu check; I've created the pipeline manually and now I'm trying to assign it via my puppet configuration instead of doing it manually

```
  lookup('sensu_checks', {merge => hash}).each |$check_name, $check| {
    sensu_check { $check_name:
      * => $check,
    }
  }
```

```yaml
sensu_checks:
  check_diskspace:
    command: "/opt/sensu-plugins-ruby/embedded/bin/check-disk-usage.rb"
    interval: 60
    subscriptions:
      - common
    handlers:
      - alert_slack
    pipelines:
      - type: Pipeline
        api_version: core/v2
        name: reduce_alerts
    timeout: 10
```

But them I get the error:

```
{"level":"err","message":"no parameter named 'pipelines' (file: /etc/puppetlabs/code/localmodules/roles/manifests/sensu_go/server.pp, line: 39) on Sensu_check[check_diskspace] (file: /etc/puppetlabs/code/localmodules/roles/manifests/sensu_go/server.pp, line: 39) on node sensu-go-qmrf.c.dcb-reference-1285.internal","source":"Puppet","tags":["err"],"file":null,"line":null,"version":1,"@timestamp":"2025-03-24T13:09:31.544847524+00:00"}
```

Am I doing anything wrong ?

Issue #1345, opened 2024-10-07
Annotations data type is too restrictive


elfranne:
Annotations data type is too restrictive, that makes it not compatible with remediation for example:
https://github.com/sensu/sensu-remediation-handler?tab=readme-ov-file#example-check-definition-and-remediation-request-configuration

```puppet
  sensu_check { 'check-puppet-errors-with-remidiation':
    ensure         => present,
    command        => 'check-puppet-errors',
    interval       => 300,
    subscriptions  => ['system/linux'],
    runtime_assets => ['nmollerup/sensu-check-puppet'],
    annotations    => {
      'comment'                              => 'The last Puppet run had failures. Investigation needed',
      'io.sensu.remediation.config.actions'  => [
        {
          'description' => 'Retry another Puppet run.',
          'request'     => 'remediation-puppet-linux',
          'occurrences' => [1],
          'severities'  => [1,2]
        },
      ],
    },
    handlers       => ['remediation'],
  }
```

Module version: 5.11.1 (latest)
Sensu version: 6.11 (OSS)
OS: Ubuntu 22.04 Jammy

Issue #1325, opened 2022-09-09
Update dependency limits


mjksohonet:
Hi,
the current version of the puppetlabs-stdlib module is 8.4.0.
Please update the dependency limits so that we can avoid problems like:
```
[Librarian]     Checking sensu-sensu/5.8.0 <https://forgeapi.puppetlabs.com/>
[Librarian]       Resolved sensu-sensu (= 5.8.0) <https://forgeapi.puppetlabs.com/> (from Puppetfile) at sensu-sensu/5.8.0 <https://forgeapi.puppetlabs.com/>
[Librarian]   Resolved sensu-sensu (= 5.8.0) <https://forgeapi.puppetlabs.com/> (from Puppetfile)
[Librarian] Conflict between puppetlabs-stdlib (< 9.0.0, >= 7.0.0) <https://forgeapi.puppetlabs.com/> (from puppet-puppetboard) and puppetlabs-stdlib (< 7.0.0, >= 5.1.0) <https://forgeapi.puppetlabs.com/> (from sensu-sensu)
Could not resolve the dependencies.
```

thepro101:
Any update on this? This sensu module has been holding up our entire environment

PR #1358, opened 2026-09-01
Puppet 8 modernization: add sensu_pipeline type, fix annotations, drop EOL platforms

sourabhpatel-sumo:
## Description

Modernizes the module for Puppet 8 and current OS support. Adds `sensu_pipeline` resource type
and pipeline support in `sensu_check`. Fixes annotations to accept mixed value types. Drops EOL
platforms.

The `sensu-plugin-ruby` removal originally scoped here has been split into a follow-up,
stacked PR targeting `6.0.0` (`plugin-removal`, based on this branch) — it's a breaking change
with no prior deprecation cycle and shouldn't ride along with this otherwise non-breaking
`5.12.0` release. `sensu_bonsai_asset` is the documented replacement (`README.md`).

## Related Issues

- #1325 — stdlib dependency too restrictive, conflicted with other modules requiring >= 7.0.0
- #1347 — deprecated `to_yaml` function replaced with `stdlib::to_yaml` in `backend.pp`
- #1346 — `sensu_check` lacked `pipelines` parameter; resolved by adding `pipelines`/`fallback_pipeline` properties and new `sensu_pipeline` type
- #1345 — annotations data type was too restrictive (strings only); resolved by allowing Array/Hash values

## Motivation and Context

Puppet 6/7 and the previously supported OS platforms are EOL.

## Breaking / behavior changes

- `sensu_user`'s `configure_url` default changed from `http://` to `https://`. A `sensu_user`
  resource that doesn't set `configure_url` explicitly (outside `sensu::backend`, which already
  derives it correctly from `use_ssl`) will now attempt HTTPS. On `use_ssl => false` deployments
  this will fail unless `configure_url` is set explicitly.
- `tasks/install_agent_windows.ps1`'s `$PackageSource` no longer has a hardcoded fallback MSI
  URL (the old default pointed at a stale Sensu Go 5.13.1 build) — it must now be supplied, and
  fails with a clear error if it's missing rather than a stale silent default.
- `appveyor.yml` is removed (CI has moved to GitHub Actions); `.pdkignore`/`.sync.yml` updated
  to match.

## How Has This Been Tested?

- `bundle exec rake sensu_validate lint strings:generate reference` — passes clean
- `bundle exec rake spec` — **3719 examples, 0 failures** (`spec/unit` + `spec/classes`)
- Acceptance tests:
  - `rocky-9` base mode — clean except `sensu::plugins`, which is `skip`-marked (see below)
  - `rocky-8-cluster` — 24 examples, 0 failures (exercises the EL8 `repo.pp` path specifically)
  - `debian-12` base mode — 77 examples, 0 failures, 14 pending
- Ruby 3.4 / Puppet 8 on macOS (arm64)

`sensu::plugins` acceptance coverage (`spec/acceptance/04_plugins_spec.rb`) is intentionally
`skip`-marked: the legacy `packagecloud.io/sensu/community` repo has no `sensu-plugins-ruby`
package for any OS this module currently supports — confirmed failing identically across
rocky-9/debian-12/almalinux-9/ubuntu-22.04. This is exactly why the subsystem is being removed
in the follow-up PR rather than fixed here.

## General

- [x] Tests pass - `bundle exec rake sensu_validate lint spec`


sarvodaya-kumar-26 (commented):
## Code Review

Solid modernization PR — tests are comprehensive and the scope is well-documented. Found a few issues worth addressing before merge, ranging from a potential data-loss risk to minor code quality items.

manifests/backend.pp
```diff
@@ -220,8 +220,15 @@
...
+    'state-dir'                        => $state_dir,
+    'api-listen-address'               => '[::]:8080',
+    'agent-port'                       => 8081,
+    'etcd-listen-client-urls'          => 'http://127.0.0.1:2379',
+    'etcd-advertise-client-urls'       => 'http://127.0.0.1:2379',
+    'etcd-listen-peer-urls'            => 'http://127.0.0.1:2380',
+    'etcd-initial-advertise-peer-urls' => 'http://127.0.0.1:2380',
+    'etcd-initial-cluster'             => 'default=http://127.0.0.1:2380',
+    'etcd-initial-cluster-state'       => 'new',
   }
```
sarvodaya-kumar-26, line 232: **Bug (data-loss risk):** `etcd-initial-cluster-state: new` is unconditionally written to `backend.yml`. On an **existing** cluster that upgrades to this module version, the next Puppet run rewrites the config with `new`. If the backend restarts (package upgrade, reboot), etcd may attempt to bootstrap a fresh cluster — potentially causing data loss or split-brain.

This should either be omitted from `$default_config` (let etcd use its compiled-in default of `new` for genuinely new clusters) or be gated behind a condition (e.g. a parameter, or an `unless File.exist?(state_dir/member)` style check).

manifests/repo.pp
```diff
@@ -5,36 +5,52 @@
...
-      # https://packagecloud.io/sensu/stable/install#bash-rpm
-      exec { 'dnf makecache sensu':
+      package { 'curl':
+        ensure => 'installed',
+      }
+
+      exec { 'install sensu repository':
+        path    => '/usr/bin:/bin:/usr/sbin:/sbin',
+        command => 'curl -s https://packagecloud.io/install/repositories/sensu/stable/script.rpm.sh | bash',
+        unless  => 'dnf repolist | grep -q sensu',
```
sarvodaya-kumar-26, line 25: **Security:** Installing the repo via `curl | bash` executes arbitrary remote code without integrity verification. A MITM or compromised CDN response can substitute the script. The `unless` guard only skips re-runs — it doesn't verify the script contents.

Consider downloading the script to a temp file first and validating a checksum, or using the native `yumrepo` resource with a GPG key (as the `else` branch already does for non-EL8).

manifests/repo.pp
```diff
@@ -5,36 +5,52 @@
...
+        require => Package['curl'],
+      }
+
+      # Use exec to replace all gpgcheck=1 with gpgcheck=0
+      exec { 'disable sensu gpg check':
+        path    => '/usr/bin:/bin:/usr/sbin:/sbin',
+        command => 'sed -i "s/gpgcheck=1/gpgcheck=0/g" /etc/yum.repos.d/sensu_stable.repo',
+        unless  => 'grep -q "gpgcheck=0" /etc/yum.repos.d/sensu_stable.repo',
```
sarvodaya-kumar-26, line 33: **Security:** This disables GPG signature checking on the Sensu repo for EL8. Combined with the `curl | bash` install above, there is **zero** package-integrity verification on EL8 nodes — any tampered RPM from the mirror will be installed silently.

If the GPG key is genuinely broken on EL8, a better approach would be to import the key separately and keep `gpgcheck=1`, or at minimum document why this is necessary and file an upstream issue.

manifests/backend.pp
```diff
@@ -369,14 +376,17 @@
...
+    $init_require = Service['sensu-backend']
   } else {
     $init_require = undef
   }
+
   exec { 'sensu-backend init':
     path        => '/usr/bin:/bin:/usr/sbin:/sbin',
-    command     => "sensu-backend init --config-file ${sensu::etc_dir}/backend.yml",
+    command     => 'sensu-backend init',
     environment => [
```
sarvodaya-kumar-26, line 390: **Bug:** The previous code used `--config-file ${sensu::etc_dir}/backend.yml` for both `command` and `unless`. Removing it means `sensu-backend init` falls back to searching `/etc/sensu/backend.yml`. If `etc_dir` is customised (e.g. `/opt/sensu`), init will connect to the wrong etcd endpoint or fail to find the config entirely.

Should this be:
```puppet
command => "sensu-backend init --config-file ${sensu::etc_dir}/backend.yml",
```

lib/puppet/type/sensu_pipeline.rb
```diff
@@ -0,0 +1,185 @@
+require_relative '../../puppet_x/sensu/type'
+require_relative '../../puppet_x/sensu/array_of_hashes_property'
+require_relative '../../puppet_x/sensu/hash_of_strings_property'
+require_relative '../../puppet_x/sensu/hash_property'
+
+SENSU_PIPELINE_VALIDATE_REF = lambda do |ref, field_name|
+  unless ref.is_a?(Hash)
```
sarvodaya-kumar-26, line 7: **Code quality:** `SENSU_PIPELINE_VALIDATE_REF` is a top-level Ruby constant. Puppet auto-loads type files multiple times per catalog compilation, which will emit `warning: already initialized constant SENSU_PIPELINE_VALIDATE_REF` on each reload. On busy Puppet masters this can generate thousands of log warnings.

Wrap it in a guard or move it inside a module:
```ruby
SENSU_PIPELINE_VALIDATE_REF ||= lambda do |ref, field_name|
```
or
```ruby
module PuppetX::Sensu
  PIPELINE_VALIDATE_REF = lambda { ... }
end
```

lib/puppet/type/sensu_user.rb
```diff
@@ -99,7 +99,7 @@ def should_to_s(newvalue)
 
   newparam(:configure_url) do
     desc "URL to use with 'sensuctl configure'"
-    defaultto 'http://127.0.0.1:8080'
+    defaultto 'https://127.0.0.1:8080'
   end
```
sarvodaya-kumar-26, line 103: **Breaking change:** The default flipped from `http://` to `https://`. Any `sensu_user` resource that doesn't explicitly set `configure_url` (e.g. user-managed resources outside `sensu::backend`) will now attempt an HTTPS connection. On deployments with `use_ssl = false`, the backend listens on plain HTTP — `sensuctl configure` will get an SSL handshake error and fail the Puppet run.

Was this intentional? If so, the upgrade notes / changelog should call it out explicitly.

lib/puppet/provider/sensuctl.rb
```diff
@@ -64,19 +65,21 @@ def convert_boolean_property_value(value)
...
-    opts[:combine] = true unless opts.key?(:combine)
+    opts = {}
+    opts[:failonfail] = failonfail.nil? ? true : failonfail
+    opts[:combine] = combine.nil? ? true : combine
+    opts[:custom_environment] = custom_environment if custom_environment
     execute(cmd, opts)
   end
-  def sensuctl(*args)
-    self.class.sensuctl(*args)
+  def sensuctl(cmd_args, **opts)
```
sarvodaya-kumar-26, line 81: **Bug (Puppet 8 / Ruby 3.4):** The fallback uses bare `YAML.load(d)` without `permitted_classes:`. On Ruby 3.1+ / Psych 4, this raises `Psych::DisallowedClass` for any non-primitive type (Symbol, Time, etc.) in sensuctl YAML output. Since this PR targets Puppet 8 / Ruby 3.4, the fallback path could silently drop resources.

Should be:
```ruby
docs = output.split('---').map { |d| YAML.safe_load(d, permitted_classes: [Symbol, Time, Date]) }
```
(Same concern applies to `YAML.load_stream` on the primary path — worth checking if it needs `permitted_classes` too.)

manifests/agent.pp
```diff
@@ -77,6 +77,22 @@
 # @param validate_entity
 #   Sets whether to validate the agent's entity before attempting
 #   to configure the entity
+# @param keepalive_handlers
+#   List of handlers for keepalive events. Maps to `keepalive-handlers` in agent.yml.
+# @param keepalive_interval
+#   Number of seconds between keepalive events. Maps to `keepalive-interval` in agent.yml.
+# @param keepalive_warning_timeout
```
sarvodaya-kumar-26, line 84: **Nit (design consideration):** Writing a full unit file at `/etc/systemd/system/sensu-agent.service` permanently overrides the package-shipped unit at `/usr/lib/systemd/system/sensu-agent.service`. Any future package upgrade that ships a new unit (hardening directives, socket activation, etc.) will be silently ignored.

Since there's already a dropin file (`sensu-agent-start.conf`) right below this block that overrides `ExecStart`, consider whether the dropin alone is sufficient instead of replacing the entire unit.

amdprophet:
## Review

Thanks for taking this on — the module is unusable on Puppet 8 today and this fixes that. The `sensu_pipeline` type, the annotations fix (#1345), and the `stdlib::to_yaml` fix (#1347) all look right, and I verified the new `sensu_check` properties against upstream (details below). Most of my comments are about a few changes that aren't mentioned in the description and are riskier than the rest of the PR.

Reviewed at 211 files, +4,122 / −4,980. I read the non-spec diff in full and checked the new Sensu fields against the `sensu/core` protos and `sensu-go` v6.14.2.

---

## Blocking

### 1. `manifests/repo.pp` — the EL8 branch disables all package signature verification

```puppet
exec { 'disable sensu gpg check':
  command => 'sed -i "s/gpgcheck=1/gpgcheck=0/g" /etc/yum.repos.d/sensu_stable.repo',
  unless  => 'grep -q "gpgcheck=0" /etc/yum.repos.d/sensu_stable.repo',
}
```

I fetched the file the packagecloud script actually installs (`config_file.repo?os=el&dist=8`). It contains both `repo_gpgcheck=1` and `gpgcheck=1`. Running that `sed` against it gives:

```
 4:repo_gpgcheck=0     <- collateral damage from the substring match
 5:gpgcheck=0
16:repo_gpgcheck=0
17:gpgcheck=0
```

The pattern `gpgcheck=1` matches inside `repo_gpgcheck=1`, so it turns that off too. And the `unless` guard doesn't match on a fresh install, so this does run.

That matters because packagecloud signs **repo metadata**, not the RPMs — `repo_gpgcheck=1` is the only signature check available here. It's why the `else` branch on line 47 deliberately keeps `repo_gpgcheck => 1, gpgcheck => 0`. The EL8 path discards the one check the non-EL8 path preserves, as an unintended side effect.

Combined with `curl -s ... | bash` (line 24) running unpinned remote code as root, EL8 nodes end up installing Sensu with no cryptographic verification at all.

The `yumrepo` in the `else` branch already handles `el/8` — the baseurl scheme is identical. Unless there's a specific EL8 failure driving this, I'd delete the whole `if String($repo_release) == '8'` branch. If there is one, could you describe it in the PR? From the diff the special case doesn't have an obvious motivation.

Worth noting your reported acceptance runs were Rocky 9 and Ubuntu 24.04, so this branch — the riskiest change in the PR — wasn't covered by them.

Minor, same block: `package { 'curl': }` on line 18 is unnamespaced and will duplicate-declare on any node that manages curl elsewhere.

### 2. `manifests/backend.pp` — dropping `api-url` breaks TLS for the backend's internal client

The new `$default_config` removes the `api-url` key that master emitted as `api-url: https://<certname>:8080`.

In sensu-go v6.14.2 `backend/cmd/start.go`, `flagAPIURL` defaults to `"http://localhost:8080"` — plaintext. Since the module defaults to `use_ssl => true`, the API listener serves TLS while the backend's own client now falls back to plain HTTP against it.

The spec change makes this read as intentional rather than flagging it:

```diff
-          |api-url: https://test.example.com:8080
```

Suggest restoring it, derived from `use_ssl`.

On the rest of the new `$default_config`: I diffed all eight hardcoded keys against `start.go` and seven are byte-identical to the upstream defaults (`agent-port: 8081`, `api-listen-address: "[::]:8080"`, the etcd URLs, `etcd-initial-cluster-state: new`). Only `etcd-advertise-client-urls` differs (`127.0.0.1` vs upstream `localhost`), which is benign. So the values are fine — it's specifically the `api-url` removal that regresses. See also the partial-override note below.

### 3. `manifests/agent.pp:318-368` — permanently shadowing the packaged systemd unit

This writes a hand-rolled unit to `/etc/systemd/system/sensu-agent.service`, which takes precedence over `/usr/lib/systemd/system/sensu-agent.service` indefinitely. Every future upstream change to that unit — hardening flags, dependency ordering, `LimitNOFILE` — is silently ignored on managed nodes.

Four issues in that block:

- **`User=sensu` / `Group=sensu` are hardcoded** (lines 334-335), but `sensu::user` and `sensu::group` are documented public parameters (`init.pp:70,72`). Setting them now produces a broken unit.
- **The `ExecStart` on line 337 is dead.** The `systemd::dropin_file` immediately below clears it (`ExecStart=`) and re-sets it from `$service_path` / `$sensu::agent_config_path`. Two sources of truth, one of which never takes effect.
- **`Exec['systemd-reload']` (line 351) is an unnamespaced global title**, likely to collide in a real catalog. It's also redundant — `systemd::dropin_file` already triggers a daemon-reload.
- **`$_env_file_line` is emitted twice** — line 336 in the unit, line 364 in the dropin.

`spec/classes/agent_spec.rb` isn't in the diff, and the existing coverage only asserts `contain_systemd__dropin_file('sensu-agent-start.conf')` — nothing exercises the new unit file.

If the dropin already worked on master, I'd drop this block entirely. If it didn't, `systemd::unit_file` with parameterized user/group is the right tool, plus spec coverage.

---

## Should fix

**Version should be `6.0.0`, not `5.12.0`**

To be clear about what I'm *not* claiming: dropping Puppet 6/7 as a minor is consistent with this module's own precedent — v5.8.0 dropped Puppet 5 (`2947ca8d`) and v5.9.0 removed EOL OSes (`8f224ab9`), both minors. No objection there.

The issue is the plugin subsystem, which is separate:

| Removed | Status on master |
|---|---|
| `sensu::plugins` | Public, `REFERENCE.md:1118`, ~90 lines of README with six worked examples |
| `sensu_plugin` type | Public, `REFERENCE.md:4070` + six documented parameters |
| `sensu_gem` package provider | Public (`package { provider => sensu_gem }`) |

None of these carry a deprecation warning on master. The description says the mechanism is "deprecated," but that's not signalled in code — a user on 5.11.1 gets no warning, then a compilation failure on 5.12.0. This project's pattern has been deprecate-then-remove-at-a-major (v5.0.0 carried #1273 "Remove various deprecations").

Since `sensu_bonsai_asset` already exists and you've documented the migration at `README.md:892`, I don't think a deprecation cycle is needed — just bump to 6.0.0 and list the removals. Alternatively, split the PR: Puppet 8 migration as 5.12.0 (urgent, uncontroversial, ships immediately) and the plugin removal as 6.0.0. That would also let the systemd and repo changes bake separately.

**`tasks/install_agent_windows.ps1` — undisclosed breaking change**

`$PackageSource` becomes `Mandatory = $True` and the hardcoded MSI URL is removed. Existing Windows task invocations that omit it now fail at parameter binding. The two schemas also disagree: `install_agent_windows.json` declares `String[1]` (required) while `install_agent.json` declares `Optional[String[1]]`.

**Missing test coverage for new code**

- `spec/unit/provider/sensu_pipeline/` doesn't exist. Two new providers (~135 lines each) ship with zero unit specs, while every other type in the module has them.
- No specs anywhere for `puppet_x/sensu/*`, despite this PR changing `insync?` and adding `validate`/`munge` to `HashProperty`. Those are exactly the changes that want direct tests.

**`ArrayOfHashesProperty#insync?` is order-sensitive**

```ruby
is.map(&strip_nils) == should.map(&strip_nils)
```

`pipelines`, `subdues`, and `output_metric_thresholds` are all semantically sets. If sensuctl returns them in a different order than declared, every run reports drift and re-applies. Suggest sorting by a stable key before comparing.

**Dead `labels` filter in `HashProperty#insync?`**

The `sensu.io/managed_by` filter added there can't run: all `labels` properties use `HashOfStringsProperty`, which already has an identical filter. `HashProperty` only parents `annotations`, `fallback_pipeline`, and `subdue`. Worth removing so the next reader doesn't assume it's load-bearing.

**`sensu-backend init` lost both its config file and its readiness guard**

```diff
-  'command' => 'sensu-backend init --config-file /etc/sensu/backend.yml',
-  'require' => 'Sensu_api_validator[sensu]',
+  'command' => 'sensu-backend init',
+  'require' => 'Service[sensu-backend]',
```

Without `--config-file`, `init` uses its own defaults rather than the managed `backend.yml`, which diverges as soon as anyone sets `config_hash`. And the service being *started* isn't the same as the API being *ready* — `Sensu_api_validator` was bridging exactly that gap. This looks like a source of flaky first-run failures.

**etcd partial-override footgun**

Because all eight etcd keys are now always present in `backend.yml`, an operator who overrides only `etcd-initial-cluster` via `config_hash` silently keeps advertising `http://127.0.0.1:2380` as their peer URL. `tests/sensu-backend-cluster.pp:16-23` overrides all eight, so acceptance won't catch it. Either omit these keys when unset, or validate that they're set as a group.

**`sensu_api.rb` — nil-URL fallback**

`uri = url ? URI.join(url, path) : URI("http://localhost#{path}")`, three times. A nil URL is a config bug; silently retargeting to plaintext localhost turns a loud failure into a confusing one. Suggest raising.

**`sensu_user.rb` — `configure_url` default flips `http://` → `https://`**

Right direction, but it's an undocumented behavior change. Worth a line in the description.

**CI coverage**

- The validate step drops `strings:generate reference`, so REFERENCE.md drift is no longer enforced in CI. I can see you run `strings:validate:reference` locally and REFERENCE.md is in fact fully in sync in this PR — it's the next PR I'd worry about.
- Both `.fixtures-latest.yml` allow-failure jobs are deleted. Collapsing the puppet dimension 4 → 1 is correct now, but those canaries were testing against *latest* stdlib/postgresql/systemd, not older Puppet. Worth keeping one.
- `parallel_spec` → `spec` slows CI and leaves `parallel_tests` unused in the Gemfile.

**Dependency bounds contradict each other**

`metadata.json` pins stdlib `>= 9.0.0 < 10.0.0`, but `tests/provision_basic_el.sh` and `provision_basic_debian.sh` install `>= 8.0.0 < 11.0.0`. Acceptance can pass against a stdlib version the module declares unsupported.

**`tests/generate-etcd-ssl.sh` — unverified download**

`curl -sL` of cfssl 1.6.5 into `/usr/local/bin` with no checksum or signature. Test-only, but it's a root-owned binary on a fresh box — a `sha256sum -c` would be cheap.

**Smaller items**

- `parse_yaml_dump`'s rescue falls back to `output.split('---').map { YAML.load(d) }` — the exact broken behavior being fixed. It'll mangle any document containing `---` in a value.
- Five providers (`sensu_cluster_federation`, `..._member`, `sensu_etcd_replicator`, `sensu_postgres_config`, `sensu_secrets_vault_provider`) inline the dump+parse logic instead of calling `dump()`, and inconsistently (`self.parse_yaml_dump` vs `Puppet::Provider::Sensuctl.parse_yaml_dump`). `dump()` still exists, now used once.
- `SENSU_PIPELINE_VALIDATE_REF` is a top-level global constant in `lib/puppet/type/sensu_pipeline.rb` — worth moving inside the type or namespacing.
- `def self.sensuctl(args, failonfail: nil, ..., **_kwargs)` — all call sites are migrated correctly, but the catch-all will silently swallow typo'd option names in future ones.
- `URI.escape(name, '/')` → `name.gsub('/', '%2F')` escapes only `/`. Fine for Bonsai's `namespace/name`; `URI.encode_www_form_component` is the drop-in if names ever widen.
- `.rubocop.yml` sets `TargetRubyVersion: '3.4'`, but puppet-agent 8 ships Ruby 3.2 — should probably be `3.2`, or you'll accept syntax the agent can't parse.
- `metadata.json` removes `pdk-version` / `template-url` / `template-ref` while `.sync.yml` stays. Worth finishing the move away from modulesync or keeping the keys.
- The Gemfile drops eight puppet-lint plugins including `legacy_facts` and `top_scope_facts` — both directly useful for a Puppet 8 migration.
- Amazon 2023 is added to the supported-OS list and mapped to `repo_release = '9'`, but `spec/fixtures/facts/amazon-2-x86_64.facts` and `amazon-2018-x86_64.facts` are deleted with no 2023 fixture added, so the new mapping is untested.

---

## Nits

- `RUN_ACCEPTANCE: 1` is set in the acceptance workflows and read nowhere in the repo.
- `acceptance-full`'s `exclude: {puppet: puppet8, ci_build: yes}` is a no-op after the matrix rewrite.
- `fallback_pipeline` requires `name`/`type`/`api_version` but doesn't reject unknown keys, while `pipelines` and `subdues` do. Same shape, different strictness.
- `BEAKER_debug: true` → `false` will make CI failures harder to diagnose.
- The cluster acceptance set drops `use_agent: yes` without explanation.

---

## What's good here

- **The new `sensu_check` properties are all real.** I checked each against `sensu/core` `v2/*.proto`: `subdue = 16`, `pipelines = 32`, `output_metric_thresholds = 33`, `subdues = 34`, `fallback_pipeline = 35`, `ttl_status = 36`, `asset_status = 37`, plus `Pipeline.ContinueOnError = 3`. The Puppet property types map correctly to the wire types (`ttl_status`→Integer for int32, `asset_status`→Array for repeated string, etc.). Validation is more thorough than it had to be — the nested `days`/day-name checks on `subdue` and the `thresholds`/`tags` structure on `output_metric_thresholds` are good. Only leniencies: `min`/`max` aren't validated as Strings and `repeat` elements aren't type-checked.
- **Two spec files renamed into existence:** `spec/acceptance/sensu_bonsai_asset.rb` → `_spec.rb` and `spec/unit/sensu_license.rb` → `_spec.rb`. Those assertions had never run. Nice catch.
- **REFERENCE.md is genuinely complete and in sync** — `subdue` (2615), `subdues` (2629), `sensu_pipeline` (4091+), `continue_on_error` (4140), all seven `keepalive_*`, and zero stale references to the removed plugin API.
- **The README migration note at line 892** pointing `sensu::plugins` users at `sensu_bonsai_asset` is the right way to do a removal.
- Test certs are valid through Aug 2036 with current CRLs — no hidden expiry cliff.
- Both Ruby shims (`spec/vendor/archive/tar/minitar.rb`, `spec/openssl_unfreeze.rb`) are small, commented, explain *why* they exist, and are excluded from the published module by `.pdkignore`. That's the right way to carry a workaround.
- Docker image caching + `timeout-minutes` in acceptance, and the local-run docs in `tests/README.md` (Apple Silicon `user_agent` workaround, `BEAKER_DESTROY=no`, corporate CA note) will save the next contributor real time.

---

## Suggested landing path

1. Delete the `repo.pp` EL8 branch and let `yumrepo` handle `el/8`.
2. Restore `api-url`, derived from `use_ssl`.
3. Drop the hand-rolled systemd unit — or parameterize user/group, namespace the exec, remove the dead `ExecStart`, and add `agent_spec.rb` coverage.
4. Add `spec/unit/provider/sensu_pipeline/{sensuctl,sensu_api}_spec.rb`.
5. Bump to `6.0.0` and enumerate the removals, or split the plugin removal into its own PR.
6. Restore `strings:generate reference` to the validate step and keep one `.fixtures-latest.yml` canary.

The pipeline support and the Ruby 3 migration are solid and I'd take them as-is. The systemd and repo changes read more like debugging artifacts that survived into the diff — I'd want those split out and justified separately.

---

<sub>Caveat: I couldn't run the suite locally (no Ruby ≥ 3.2 available on this machine), so all of the above is from static reading plus verification against the `sensu/core` protos, sensu-go v6.14.2 `backend/cmd/start.go`, the live packagecloud config file, and Puppet's `exec/posix` provider. I read the non-spec diff in full; spec and nodeset diffs by diffstat and targeted greps. Review assisted by Claude Code.</sub>

lib/puppet/type/sensu_user.rb
sourabhpatel-sumo: Added to the description, CHANGELOG is autogenerated.

sourabhpatel-sumo:
@amdprophet @sarvodaya-kumar-26 — pushed updates addressing both reviews:

- Split the plugin removal into #1359 (stacked on this branch, targets 6.0.0), per @amdprophet's suggestion — this PR stays non-breaking at 5.12.0.
- Fixed all blocking items: etcd-initial-cluster-state data-loss risk, curl | bash + GPG-disable on EL8, sensu-backend init --config-file regression, SENSU_PIPELINE_VALIDATE_REF namespaced under PuppetX::Sensu, YAML.safe_load with permitted_classes on both sensuctl.rb call sites, hand-rolled systemd unit removed (dropin only remains).
- sensu_user's configure_url http→https change is now disclosed in the description's "Breaking / behavior changes" section.
- Description and test numbers updated to current state (3719 examples, 0 failures; three acceptance runs listed).

Let me know if anything needs another look.

PR #1359, opened 2026-09-08
Remove deprecated sensu-plugins-ruby subsystem (v6.0.0)

sourabhpatel-sumo:
## Summary

Stacked on #1358 (`puppet8-modernization`) — this is the breaking-change follow-up split out of
that PR so v5.12.0 can ship independently. This PR bumps to **v6.0.0** and removes the deprecated
`sensu-plugins-ruby` subsystem entirely:

- `sensu::plugins` class and `sensu::repo::community` class
- `sensu_plugin` type/provider and `sensu_gem` package provider
- Related hiera keys (`sensu::plugins::dependencies` in `data/os/{RedHat,Debian}.yaml`)
- All associated spec files (unit, class, and acceptance)

## Scope — what this does *not* touch

`sensu-ruby-runtime` is a **Bonsai asset**, registered through the `sensu_bonsai_asset` type — a
backend-registered tarball the agent downloads at runtime. It is **unaffected by this PR** and still
exercised in `spec/acceptance/sensu_bonsai_asset_spec.rb`. Ruby plugins delivered as Bonsai assets
keep working exactly as they do today.

What this PR removes is `sensu-plugins-ruby`: a Sensu Core 1.x-lineage *host-level* installer that
adds a third OS repo (`packagecloud.io/sensu/community`, separate from `sensu/stable`), installs an
omnibus package with its own embedded Ruby at `/opt/sensu-plugins-ruby/embedded/`, installs a
compiler toolchain, and shells out to `sensu-install` to compile gems at Puppet-run time. The two
share the word "ruby" and nothing else.

**Why:** the modern replacement is `sensu_bonsai_asset` (see the migration note already added to
`README.md`): where you previously used `sensu::plugins { plugins => ['disk-checks'] }`, use
`sensu_bonsai_asset` with the corresponding Bonsai namespace/name instead
(e.g. `sensu/sensu-plugins-disk-checks`).

Package availability on `packagecloud.io/sensu/community`, checked against every OS this module
currently supports (`metadata.json`):

| Platform | `sensu-plugins-ruby` published? |
|---|---|
| el/8 (RedHat/Rocky/AlmaLinux 8) | **Yes** — `0.2.0-1.el8`, `0.3.0-2.el8`, **x86_64 only**, built 2020-08-11, embedded Ruby 2.4 (EOL Mar 2020) |
| el/9 (RedHat/Rocky/AlmaLinux 9) | No — never published |
| Debian 12 (bookworm) | No — never published |
| Ubuntu 22.04 (jammy), 24.04 (noble) | No — never published |
| Amazon 2023 | Mis-resolves to el/7 via a stale `/^201\d$/` regex in `repo/community.pp` — a bug, not support |
| Windows | N/A — `sensu::plugins` fails on Windows |

**Correction to an earlier claim in this PR:** I originally said no package exists for any currently
supported OS — that's wrong for EL8. The acceptance skip was verified on rocky-9, debian-12,
almalinux-9 and ubuntu-22.04; `04_plugins_spec.rb` is gated on `sensu_mode == 'base'` and the only
EL8 nodeset is `rocky-8-cluster.yml` (cluster mode), so no EL8 node ever ran it.
`sensu::plugins` still functions on EL8 x86_64 today. See discussion below for usage data on that
platform.

## Test plan

Full local CI-equivalent run on this branch (Apple Silicon; two acceptance-full jobs that require
an x86_64-only Puppet package are noted below and were verified via CI on the parent branch
instead):

- [x] `rake sensu_validate lint strings:generate reference` — clean, `REFERENCE.md` matches
- [x] `rake parallel_spec` (CI's exact `BUNDLE_WITHOUT=system_tests:development`) — **3701 examples, 0 failures**
- [x] Acceptance base mode: rocky-9, debian-12, ubuntu-24.04, ubuntu-22.04, almalinux-9 — all 69/0 (6 pending enterprise-only, expected)
- [x] Acceptance full/types/examples: rocky-9 and ubuntu-24.04 — full 7/0, types 254/0, examples 10/0 (both sets)
- [x] Acceptance cluster: rocky-8-cluster, ubuntu-24.04-cluster — 24/0 (both)
- [ ] Acceptance `bolt` mode and `types` mode with `use_agent=yes` (both rocky-9-only jobs) — not
      runnable on Apple Silicon: confirmed by direct repro that Puppet Inc. publishes no aarch64
      EL9 RPM for `puppet-bolt` or `puppetserver`. Neither job is affected by this diff (no bolt-
      or puppetserver-setup code touched here), and both already pass on x86_64 GitHub Actions
      CI for the parent `puppet8-modernization` branch.
- [x] Confirmed no dangling references to the removed classes/types anywhere in manifests,
      `REFERENCE.md`, or remaining specs

🤖 Generated with [Claude Code](https://claude.com/claude-code)


amdprophet (changes requested):
This is a major change as https://github.com/sensu/sensu-ruby-runtime has historically been used by many plugins from the community (not just our own). If we haven't already, please determine how many users are still using ruby plugins. I can see this being a blocker for upgrading to v6.0.0 of the puppet module for some people.

sourabhpatel-sumo:
@amdprophet I've updated the PR description with a correction and the scope of what this actually touches — short version: `sensu-ruby-runtime` (Bonsai asset) is untouched, and my "no package
  for any supported OS" claim was wrong for EL8. See the description for the detail and the availability table.

  **On the usage numbers you asked for.** Pulled from packagecloud — 30-day downloads, API `downloads_count`, summed across all 41 artifacts in the repo (which contains only this package). **Total 297**:

  | Distro | 30d | Vendor EOL | Still supported? |
  |---|---|---|---|
  | ubuntu/quantal | 44 | Apr 2014 (12.10) | No |
  | el/7 | 42 | Jun 2024 | No |
  | ubuntu/bionic | 42 | May 2023 | No |
  | ubuntu/trusty | 39 | Apr 2019 | No |
  | el/6 | 27 | Nov 2020 | No |
  | ubuntu/xenial | 26 | Apr 2021 | No |
  | ubuntu/focal | 22 | Apr 2025 | No — dropped by #1358 |
  | **el/8** | **21** | **May 2029** | **Yes** |
  | debian/stretch | 16 | Jun 2022 | No |
  | debian/jessie | 12 | Jun 2020 | No |
  | debian/buster | 6 | Jun 2024 | No |

  Two things stand out:

  1. **92.9% is on platforms #1358 already dropped as EOL.** Those users can't reach v5.12.0 regardless of this PR — only the el/8 slice could be blocked by the removal itself.
  2. **el/8 is 21 — and Ubuntu 12.10, EOL for twelve years, is 44.** RHEL 6 is 27. The spread is nearly flat with no concentration on newer platforms, which looks more like mirror/scanner/CI traffic than real convergence — organic usage would skew toward the newest supported platform.

  My read: el/8's 21 sits inside the noise floor set by platforms that can't have live users. I'd rather say that carefully — "indistinguishable from automated traffic," not "nobody uses this" — since I can't actually tell the two apart from download counts alone. Worth noting `package_ensure` defaults to `installed` (idempotent), so these numbers undercount the installed base while overcounting distinct users either way.

  One caveat on the data itself: the repo page's "installs in the last 30 days" widget shows 80 for the same window, not 297 — apparently a different metric than the API's `downloads_count`. Everything above uses the API figure consistently, but I wanted to flag the discrepancy rather than have it surface later.

  **If you have visibility I don't** — support tickets, customer reports, known EL8 shops — that would override my read here, and I'd trust it over the download data.

  **One regression worth naming plainly:** `sensu_plugin`'s `source` parameter (a custom gem mirror URL passed to `sensu-install`) has no `sensu_bonsai_asset` equivalent — Bonsai installs assets, not arbitrary gems from a private server. I found one public control-repo relying on exactly this. The workaround (publish an asset tarball, use `sensu_asset` with `builds`) is real extra work for anyone in that position.

  **And you're right that there's no deprecation cycle here** — I checked, and there never has been, on any part of this subsystem, in its whole history. That's a gap on my end, not a defense of the PR as it stands.

  Given that, I think the honest options are:

  1. Ship the removal in 6.0.0 as-is, now with the corrected rationale in the description.
  2. Add deprecation warnings to #1358/v5.12.0 first (still unmerged, subsystem still intact there) and hold this PR until v5.12.0 is actually on the Forge — since a deprecation cycle only counts once people can install a version that warns them.

  I don't have strong footing to pick between these myself — happy to do whichever you think is right, and (2) is not much extra work if you'd rather be safe.

amdprophet (approved):

